PR #95, opened 2020-05-13
The `Headers` object is now flattened before fetch

### LINKED ISSUE #88: Request headers are overriden when using `LDflex-Comunica`
# Description of the bug
 
When called with the following values:
```
"init": {
  "headers": {
    "_headers": {
      "accept": [
        "application/n-quads,application/trig;q=0.95,application/ld+json;q=0.9,application/n-triples;q=0.8,text/turtle;q=0.5,application/rdf+xml;q=0.5,application/json;q=0.45,text/n3;q=0.35,application/xml;q=0.3,text/xml;q=0.3,image/svg+xml;q=0.3,text/html;q=0.2,application/xhtml+xml;q=0.18"
      ]
    }
  }
},
"input": "https://www.w3.org/People/Berners-Lee/card"
```
in a `NodeJS` environment (using `cross-fetch)`, `solid-auth-fetcher` performs a query with default `Accept` headers, instead of the ones specified in the `init` data structure. 

This bug was observed while upgrading [LDflex-Comunica](https://github.com/NSeydoux/LDflex-Comunica). I'm not entirely sure the `init` structure above is compatible with the expected `RequestInit` [type](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.requestinit.html), namely the additional wrapping `_headers` seems extraneous. However, this exact same structure was processed successfully by `solid-auth-client`. A special case might be added to patch the issue on a short term, but if the problem comes from the data structure passed by `LDflex-Comunica`, then actual patch should be implemented there.

# Steps to reproduce

Clone `https://github.com/NSeydoux/LDflex-Comunica`, run `npm ci` and `npm run test`. The tests timeout for a different reason, but the core issue is that the request isn't stubbed by PollyJS because it doesn't match the expected signature because the headers have been mangled.

NSeydoux:
Resolves #88

It seems that an object implementing correctly the `Headers` interface is not processed correctly by `node-fetch`, pulled in by `cross-fetch`. I'll develop a minimal example to verify that this diagnosis is correct, and if so create an appropriate issue in the relevant repo.

### DIFF
--- a/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
+++ b/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
@@ -86,14 +86,14 @@ export default class DpopAuthenticatedFetcher implements IAuthenticatedFetcher {
       // perform unauthenticated fetch
       return this.fetcher.fetch(url, {
         ...requestInit,
-        headers: flattenHeaders(requestInitiWithDefaults?.headers)
+        headers: flattenHeaders(requestInitiWithDefaults.headers)
       });
     }
 
     return this.fetcher.fetch(url, {
       ...requestInit,
       headers: {
-        ...flattenHeaders(requestInitiWithDefaults?.headers),
+        ...flattenHeaders(requestInitiWithDefaults.headers),
         authorization: `DPOP ${authToken}`,
         dpop: await this.dpopHeaderCreator.createHeaderToken(
           this.urlRepresentationConverter.requestInfoToUrl(url),
--- a/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
+++ b/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
@@ -32,7 +32,7 @@ import { IDpopHeaderCreator } from "../../dpop/DpopHeaderCreator";
 import { IUrlRepresentationConverter } from "../../util/UrlRepresenationConverter";
 import { IStorageUtility } from "../../localStorage/StorageUtility";
 
-function flattenHeaders(
+export function flattenHeaders(
   headersToFlatten: Headers | string[][] | Record<string, string> | undefined
 ): Record<string, string> {
   const flatHeaders: Record<string, string> = {};
--- a/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
+++ b/src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
@@ -32,6 +32,17 @@ import { IDpopHeaderCreator } from "../../dpop/DpopHeaderCreator";
 import { IUrlRepresentationConverter } from "../../util/UrlRepresenationConverter";
 import { IStorageUtility } from "../../localStorage/StorageUtility";
 
+function flattenHeaders(
+  headersToFlatten: Headers | string[][] | Record<string, string> | undefined
+): Record<string, string> {
+  const flatHeaders: Record<string, string> = {};
+  const iterableHeaders = new Headers(headersToFlatten);
+  iterableHeaders.forEach((value, key) => {
+    flatHeaders[key] = value;
+  });
+  return flatHeaders;
+}
+
 @injectable()
 export default class DpopAuthenticatedFetcher implements IAuthenticatedFetcher {
   constructor(
@@ -75,16 +86,14 @@ export default class DpopAuthenticatedFetcher implements IAuthenticatedFetcher {
       // perform unauthenticated fetch
       return this.fetcher.fetch(url, {
         ...requestInit,
-        headers: {
-          ...requestInitiWithDefaults.headers
-        }
+        headers: flattenHeaders(requestInitiWithDefaults?.headers)
       });
     }
 
     return this.fetcher.fetch(url, {
       ...requestInit,
       headers: {
-        ...requestInitiWithDefaults.headers,
+        ...flattenHeaders(requestInitiWithDefaults?.headers),
         authorization: `DPOP ${authToken}`,
         dpop: await this.dpopHeaderCreator.createHeaderToken(
           this.urlRepresentationConverter.requestInfoToUrl(url),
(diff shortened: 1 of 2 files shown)


src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
```diff
@@ -75,16 +86,14 @@ export default class DpopAuthenticatedFetcher implements IAuthenticatedFetcher {
       // perform unauthenticated fetch
       return this.fetcher.fetch(url, {
         ...requestInit,
-        headers: {
-          ...requestInitiWithDefaults.headers
-        }
+        headers: flattenHeaders(requestInitiWithDefaults?.headers)
```
RubenVerborgh, line 89: Can be replaced by `Object.fromEntries(new Headers(requestInitiWithDefaults?.headers).toEntries())`

RubenVerborgh: Or alternatively, change the `flattenHeaders` method to that implementation for easier reuse.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
```diff
@@ -32,6 +32,17 @@ import { IDpopHeaderCreator } from "../../dpop/DpopHeaderCreator";
 import { IUrlRepresentationConverter } from "../../util/UrlRepresenationConverter";
 import { IStorageUtility } from "../../localStorage/StorageUtility";
 
+function flattenHeaders(
+  headersToFlatten: Headers | string[][] | Record<string, string> | undefined
```
RubenVerborgh, line 36: I don't think `string[][]` can occur.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
NSeydoux: TypeScript complains that there is no `toEntries` or `entries` methods in the `Headers` interface, even if it is mandated by the [MDN spec](https://developer.mozilla.org/en-US/docs/Web/API/Headers). Here is the `Headers` interdaces as per `lib.dom.d.ts`:
```
/** This Fetch API interface allows you to perform various actions on HTTP request and response headers. These actions include retrieving, setting, adding to, and removing. A Headers object has an associated header list, which is initially empty and consists of zero or more name and value pairs.  You can add to this using methods like append() (see Examples.) In all methods of this interface, header names are matched by case-insensitive byte sequence. */
interface Headers {
    append(name: string, value: string): void;
    delete(name: string): void;
    get(name: string): string | null;
    has(name: string): boolean;
    set(name: string, value: string): void;
    forEach(callbackfn: (value: string, key: string, parent: Headers) => void, thisArg?: any): void;
}
```

Similar mismatches might very well be causing the observed issue in the first place.

RubenVerborgh: Mwm, we should just cast here then; TypeScript is wrong.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
NSeydoux: This is part of the type signature of `RequestInit.headers`.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
NSeydoux: The types definition in `node-fetch` seem to be aligned with the spec, I'll use them for casting.

RubenVerborgh: Weird that they don't accept `Headers` then though.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
RubenVerborgh: Weird, not on MDN. The `Headers` constructor takes it, but not the `headers` field as far as I can see. No biggy though, it will work.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
Vinnl: Might be because MDN still lists `entries` as:

>     Experimental. Expect behavior to change in the future.

Otherwise, consider [contributing it upstream](https://github.com/Microsoft/TypeScript/blob/master/CONTRIBUTING.md#contributing-libdts-fixes).

RubenVerborgh: To be fair, all properties of the object are experimental 😁

NSeydoux: Actually casting this would have an impact on several interfaces, such as `IFetcher` and `IAuthenticatedFetcher`, and I think it will have rippling effects all over the place. Since the proposed fix patches the issue at hand right now, I suggest that a deeper typing refactoring is looked at in a different PR. If you think it's an appropriate approach, I'll create the associated issue.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
NSeydoux: This issue is similar to the one discussed above: there's a divergence in the `Headers` constructors. This will be updated when we straighten out the typing.

src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
RubenVerborgh: Was thinking about a quick local `any` cast for `entries`, but no strong opinion.

Vinnl: That's... A good point :P Then maybe consider a PR to TS :)

PR #100, opened 2020-05-13
Release/0.0.3

NSeydoux:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "solid-auth-fetcher",
```
Vinnl, line 2: For some reason, npm doesn't complain about these names misaligning, but then if you do an `npm install` later, even if it doesn't need to install anything, it updates this name and you'll suddenly have unstaged changes in your repo. So best to change it now, I suppose :)

```suggestion
  "name": "@inrupt/solid-auth-fetcher",
```

(That said, I think Emmet said we didn't want `@inrupt/`-scoped packages? I did it for lit-solid-core because that's the only way to publish it to GitHub Packages, but on npm we might want to settle on a scope before spreading the word...)

NSeydoux: That's a good point. The discussion is still ongoing I think, so what I suggest we do is to stick with the `@inrupt` scope right now for consistency, but restrain from making any PR to external projects until this question is settled. We can still work on forked versions of these projects (e.g. `LDflex-comunica`), and when we have an answer for the scope we'll only have to update these forks, before actually contributing to the community with the stable scope. What do you think ?

Vinnl: Yep, but be sure to communicate with Emmet/Oz, since I think I heard words going around about an announcement being planned in the short term.

PR #101, opened 2020-05-14
Dynamic registration

jaxoncreed:


__tests__/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import "reflect-metadata";
+import { FetcherMock } from "../../../src/util/__mocks__/Fetcher";
+import ClientRegistrar from "../../../src/login/oidc/ClientRegistrar";
+import { IssuerConfigFetcherFetchConfigResponse } from "../../../src/login/oidc/__mocks__/IssuerConfigFetcher";
+import { Response as NodeResponse } from "node-fetch";
```
NSeydoux, line 26: Why use the `node-fetch` `Response` type here, instead of `lib.dom.d.ts` `Response` ? A mix of these two types appear in other places in the code, and I'm curious about the motivation for it.

__tests__/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.spec.ts
```diff
@@ -89,5 +89,30 @@ describe("AuthorizationCodeWithPkceOidcHandler", () => {
...
       expect(session.neededAction).toMatchObject(RedirectorResponse);
     });
+
+    it("Handles login when a client secret is present", async () => {
+      const authorizationCodeWithPkceOidcHandler = getAuthorizationCodeWithPkceOidcHandler();
+      const oidcOptions: IOidcOptions = {
+        ...standardOidcOptions,
+        client: {
+          ...standardOidcOptions.client,
+          clientSecret: "I'm can't cook because I only drink Soylent"
```
NSeydoux, line 99:
```suggestion
          clientSecret: "I can't cook because I only drink Soylent"
```
Client secrets: a developer's diary.

__tests__/login/oidc/ClientRegistrar.spec.ts
Vinnl: If I understand it correctly, in this case it's because it's not a type but a class, i.e. `new NodeResponse` is specifically executed below, and unfortunately, it does not appear to be compatible with the browser's Response API?

__tests__/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.spec.ts
```diff
@@ -89,5 +89,30 @@ describe("AuthorizationCodeWithPkceOidcHandler", () => {
       );
       expect(session.neededAction).toMatchObject(RedirectorResponse);
     });
+
+    it("Handles login when a client secret is present", async () => {
```
Vinnl, line 93: nit:

```suggestion
    it("handles login when a client secret is present", async () => {
```

src/login/oidc/ClientRegistrar.ts
```diff
@@ -0,0 +1,95 @@
...
+    issuerConfig: IIssuerConfig
+  ): Promise<IClient>;
+}
+
+@injectable()
+export default class ClientRegistrar implements IClientRegistrar {
+  constructor(@inject("fetcher") private fetcher: IFetcher) {}
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private clientResponseToClientObject(response: Record<string, any>): IClient {
```
Vinnl, line 40: Is there a reason this needs to be so loose?

```suggestion
  private clientResponseToClientObject(response: { clientId: string; clientSecret: string }): IClient {
```

__tests__/login/oidc/ClientRegistrar.spec.ts
jaxoncreed: Exactly. You can't instantiate Response in lib.dom.d.ts.

NSeydoux: Thanks !

PR #109, opened 2020-05-18
Replaced Headers type by NodeHeaders class for instantiation

NSeydoux:
Following a conversation we had around this exact topic in an NSS PR, I realized that the code I just wrote to fix the Headers issue used `lib.dom.d.ts:Headers` as a class, which is incorrect. The weird thing is that this line is covered by unit tests that did not pick up the issue, neither did the downstream libraries using the code (e.g. `LDflex-Comunica`). It's only when writing a super-simple code snippet (see below) for test purpose that this blew up:
```javascript
const auth = require('@inrupt/solid-auth-fetcher')

auth.fetch("http://some.iri/")
.then(response => response.text())
.then(data => console.log(data));
```


NSeydoux:
Makes sense, and since I checked that things were running okay with the `LDflex-Comunica` test suite, the same may have happened.

PR #115, opened 2020-05-20
Refresh token

jaxoncreed:
Refresh token support. This does not include 100% branch coverage as I plan to refactor some things in the hardening sprint.


src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
```diff
@@ -52,6 +52,7 @@ export default class AuthorizationCodeWithPkceOidcHandler
   }
 
   async handle(oidcLoginOptions: IOidcOptions): Promise<ISolidSession> {
+    oidcLoginOptions = oidcLoginOptions as IAccessTokenOidcOptions;
```
Vinnl, line 55: If we're expecting the additional properties from `IAccessTokenOidcOptions` here, shouldn't the parameter have that type? The cast seems dangerous...

```suggestion
  async handle(oidcLoginOptions: IAccessTokenOidcOptions): Promise<ISolidSession> {
```

src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -0,0 +1,120 @@
...
+    // Perform refresh token flow
+    const tokenRequestInit: RequestInit = {
+      method: "POST",
+      headers: {
+        DPoP: await this.dpopHeaderCreator.createHeaderToken(
+          issuerConfig.tokenEndpoint,
+          "POST"
+        ),
+        "content-type": "application/x-www-form-urlencoded"
+      },
```
Vinnl, line 79: To avoid the cast on line 90, you might want to consider not to define this type so much narrower than it is. If you want to ensure it conforms to `RequestInit` still, you could do something like this:

```suggestion
    const tokenRequestInit: RequestInit & { headers: Record<string, string> } = {
      method: "POST",
      headers: {
        DPoP: await this.dpopHeaderCreator.createHeaderToken(
          issuerConfig.tokenEndpoint,
          "POST"
        ),
        "content-type": "application/x-www-form-urlencoded"
      },
```

src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    if (clientSecret) {
+      (tokenRequestInit.headers as Record<
+        string,
+        string
+      >).Authorization = `Basic ${this.btoa(`${clientId}:${clientSecret}`)}`;
+    }
+
+    const tokenResponse = await (
```
Vinnl, line 96: What if the response is invalid? You shouldn't need the casts below (`as string`); ideally, a validator acts as a type guard.

__tests__/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    mocks: Partial<typeof defaultMocks> = defaultMocks
+  ): TokenRefresher {
+    return new TokenRefresher(
+      mocks.fetcher ?? defaultMocks.fetcher,
+      mocks.issueConfigFetcher ?? defaultMocks.issueConfigFetcher,
+      mocks.storageUtility ?? defaultMocks.storageUtility,
+      mocks.dpopHeaderCreator ?? defaultMocks.dpopHeaderCreator
+    );
+  }
+
```
Vinnl, line 52: As mentioned in the class itself: it would be nice to also have a test in which the fetcher returns invalid JSON, or the request fails, and verifies that the resulting error is meaningful to the developer and can be gracefully dealt with.

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
```diff
@@ -50,6 +50,7 @@ export default class LegacyImplicitFlowOidcHandler implements IOidcHandler {
   }
 
   async handle(oidcLoginOptions: IOidcOptions): Promise<ISolidSession> {
+    oidcLoginOptions = oidcLoginOptions as IAccessTokenOidcOptions;
```
Vinnl, line 53: As with my other comment: it'd be great if you could find a way to avoid this cast.

```suggestion
  async handle(oidcLoginOptions: IAccessTokenOidcOptions): Promise<ISolidSession> {
```

src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
jaxoncreed: I guess the cast is dangerous if you don't perform a test before-hand. I added a test.

src/login/oidc/refresh/TokenRefresher.ts
jaxoncreed: Cool. Done

PR #121, opened 2020-05-20
Fixes regression on the mangled headers issue

NSeydoux:
The original issue here is that some headers are not properly passed by `solid-auth-fetcher` to the underlying library that performs the fetch. This bug was encountered when upgrading LDflex-Comunica, which it turns out passes a slightly non-standard `Headers` object. An initial fix to this issue failed to consider the execution in a Node environment, and was assuming some features only implemented by the browser. Fixing this new issue introduced a regression regarding the original problem, and this (dirty) fix aims at satisfying both use cases.

Most of the code in this PR is a temporary fix required to upgrade the SDK, but on the long run I'd like to sit down and see where we should contribute in order to clean this up.


src/authenticatedFetch/dpop/DpopAuthenticatedFetcher.ts
```diff
@@ -31,16 +31,84 @@ import { IFetcher } from "../../util/Fetcher";
...
+ * This might require a fix upstream to be cleaned up.
+ *
+ * @param headersToFlatten A structure containing headers potentially in several formats
+ */
 export function flattenHeaders(
   headersToFlatten: Headers | string[][] | Record<string, string> | undefined
 ): Record<string, string> {
+  if (headersToFlatten === undefined) {
+    return {};
+  }
```
jaxoncreed, line 87: The functions in this file seem like they could be used in other classes as well (The the bearer authenticated fetcher). A new class should be created called `HeaderFlattener` or something like that and it should be injected into this class and any other class that makes a fetch.

jaxoncreed: I also now notice that the `if (!authToken) {` on line 85 should not actually be in this file. Instead there should be another handler called `UnauthenticatedFetcher` that handles the case where the auth token isn't present.

NSeydoux:
I have a design question. The `DpopAuthenticatedFetcher::canHandle` method checks that the request type is `dpop`, and the `DpopAuthenticatedFetcher::handle` methods checks for the presence of the auth token. Combined to the fact that `AuthFetcher::fetch` enforces the request type to be `dpop`, the unauthenticated fetcher is never used, and on unauthenticated fetch the dpop fetcher fails because of the lack of an auth token.

A fix currently implemented is that `DpopAuthenticatedFetcher::canHandle` also checks for the presence of the auth token, so that if the auth token is missing, the dpop handler isn't called, and the aggregated handler defaults to the unauthenticated handler. However, this seems like an improper use of the handler pattern. I think that a better approach would be to change https://github.com/inrupt/solid-auth-fetcher/blob/d274ddd806f8d1a28f0f8f8f0e5bbd37f15a885f/src/AuthFetcher.ts#L77 in order to remove the `dpop` type when not applicable. However, I'm not really sure how to do this without the user having to specify if the request should be authenticated or not, which we absolutely don't want.Attempting to fetch with the `dpop` type, and defaulting to unauthenticated on fail, is also possible, but seems inelegant. What do you think ?

Issue #90, opened 2020-05-12
Does not fail elegantly when token exchange doesn't work


jaxoncreed:
The error given is "decoded is not defined"

nicolasmondada:
This has been addressed on #411 and the use of [oidc-client](https://github.com/IdentityModel/oidc-client-js) library

PR #132, opened 2020-05-26
Fix/90 Token Error handling

jaxoncreed:
Fix for #90


src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -124,12 +124,26 @@ export default class AuthCodeRedirectHandler implements IRedirectHandler {
...
+        tokenResponse &&
+        tokenResponse.access_token &&
+        tokenResponse.id_token &&
+        typeof tokenResponse.access_token === "string" &&
+        typeof tokenResponse.id_token === "string" &&
+        (!tokenResponse.refresh_token ||
+          typeof tokenResponse.refresh_token === "string")
+      )
+    ) {
+      throw new Error("IDP /token route returned an invalid response.");
```
NSeydoux, line 139: Is the token endpoint always going to be `/token`, or should the error message use `issuerConfig.tokenEndpoint` instead ?

src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Check the response
+    if (
+      !(
+        tokenResponse &&
+        tokenResponse.access_token &&
+        tokenResponse.id_token &&
+        typeof tokenResponse.access_token === "string" &&
+        typeof tokenResponse.id_token === "string" &&
+        (!tokenResponse.refresh_token ||
+          typeof tokenResponse.refresh_token === "string")
```
NSeydoux, line 109: This looks similar to the `AuthCodeRedirectHandler` response validation. Should/could it be factored out as a function that gets called whenever appropriate ?

src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
jaxoncreed: Good catch

__tests__/login/oidc/TokenRequester.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+          Authorization: "Basic Y29vbEFwcDp0aGVTZWNyZXRTYXVjZUlzSnVzdE1heW8=",
+          "content-type": "application/x-www-form-urlencoded"
+        },
+        body:
+          "grant_type=refresh_token&refresh_token=thisIsARefreshToken&client_id=coolApp"
+      }
+    );
+  });
+
+  it("Fails elegantly if the idp returns a bad value", async () => {
```
pmcb55, line 165: Multiple "it("Fails elegantly if..." references in these tests, which doesn't really make sense (i.e. all failures should be 'elegant'!) - so should just be replaced with 'standard' wording like "it("should fail if...".

__tests__/login/oidc/TokenRequester.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    );
+    defaultMocks.joseUtility.decodeJWT.mockReset();
+    defaultMocks.joseUtility.decodeJWT.mockResolvedValueOnce(
+      values.jwt ?? defaultReturnValues.jwt
+    );
+  }
+
+  it("Properly follows the refresh flow", async () => {
+    setUpMockedReturnValues({});
+    const TokenRefresher = getTokenRequester();
```
pmcb55, line 118: I assume this local var should be named 'tokenRefresher' (lower case 'T') - but also, is this var a 'Refresher' or a 'Requester'? From the code it seems it's a token requester that just happens to expect a refresh token, which still makes it a `tokenRequester` (pedantic I know!).

__tests__/login/oidc/TokenRequester.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    await expect(
+      /* eslint-disable @typescript-eslint/camelcase */
+      TokenRefresher.request("global", {
+        grant_type: "refresh_token",
+        refresh_token: "thisIsARefreshToken"
+      })
+      /* eslint-enable @typescript-eslint/camelcase */
+    ).rejects.toThrowError(
+      "The issuer https://idp.com does not support the refresh_token grant"
```
pmcb55, line 200: Two points:
1) I always delimit placeholders in message strings with '[' and ']' (i.e. in this case the error should be `"The issuer [https://idp.com] does not ...`
2) I wouldn't hardcode the expected param value, I'd replace with the test value, which I think means: `The issuer [${defaultReturnValues.storageIdp}] does not...`

__tests__/login/oidc/TokenRequester.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    });
+    const TokenRefresher = getTokenRequester();
+    await expect(
+      /* eslint-disable @typescript-eslint/camelcase */
+      TokenRefresher.request("global", {
+        grant_type: "refresh_token",
+        refresh_token: "thisIsARefreshToken"
+      })
+      /* eslint-enable @typescript-eslint/camelcase */
+    ).rejects.toThrowError("The idp returned a bad token without a sub.");
```
pmcb55, line 242: Error message here - why is it 'idp' and not 'IDP' (and in fact I prefer 'IdP', but more important to just be consistent everywhere!). I'd reword slightly too to: "The IdP returned a bad token: missing 'sub' claim."

__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -95,78 +82,41 @@ describe("AuthCodeRedirectHandler", () => {
...
+          redirectByReplacingState: true
         }
       );
-      expect(
-        defaultMocks.tokenSaver.saveTokenAndGetSession
-      ).toHaveBeenCalledWith("userId", "idToken", "accessToken", undefined);
     });
 
-    it("makes a code request with a Basic authorization header if a secret is present", async () => {
+    it("Fails if a session was not retrieved", async () => {
```
pmcb55, line 109: Should be lower-case message string, ie. "fails if a...".

src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -56,93 +49,38 @@ export default class AuthCodeRedirectHandler implements IRedirectHandler {
...
+    url.set("query", {});
 
-    if (clientSecret) {
-      (tokenRequestInit.headers as Record<
-        string,
-        string
-      >).Authorization = `Basic ${this.btoa(`${clientId}:${clientSecret}`)}`;
+    const session = await this.sessionCreator.getSession(localUserId);
+    if (!session) {
+      throw new Error("There was a problem creating a session.");
```
pmcb55, line 78: Yikes! This seems no better than the dreaded "Server error" message! Is there any means to determine what the problem may have been (with the acknowledgement that less info is better from a security perspective, but I'd hope *some* indication of the problem might be helpful here)...?

src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
pmcb55: And the value itself should be delimited in the message string (I always use '[' and ']'), as this helps see if the value has leading or trailing spaces.

Issue #122, opened 2020-05-20
Does not bundle well with Webpack


NSeydoux:
**Describe the bug**
If bundled with Webpack, throws an exception when any function is called. 

**To Reproduce**
1. Create a simple React app with `create-react-app`
2. `npm install @inrupt/solid-auth-fetcher`
3. Add `{fetch("http://www.perdu.com/").then(response => response.text).then(data => console.log(data))}` to the newly created app
4. `npm run start`
5. the browser should display the error `TypeError: Object(...) is not a function`.

**Expected behavior**
The functions should be available after bundling.

jaxoncreed:
Wow it takes a super long time to start up the create-react-app project :( https://github.com/facebook/create-react-app/issues/2536

This issue is a bit strange. Working on it.

jaxoncreed:
It seems as if the compiled index.js file does not list "module" at all.

Vinnl:
This might be why both @ajacksified and I didn't manage to get it to work with lit-solid too.

PR #134, opened 2020-05-27
Fix webpack export

ajacksified:
Updates the libraryTarget to export a commonjs module, which plays nicely with nodejs. However, this will _not_ export a global variable like it currently does - which means you can't just drop the compiled javascript on a page (like the first example on the README.) Two questions:

* Is exporting a global variable useful, or can we assume any user of SAF is using a bundler like webpack?
* If so, we can create a "dist/browser.js" file that imports "dist/index.js" and sets `global.solidAuthFetcher`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@inrupt/solid-auth-fetcher",
-  "version": "0.0.4",
+  "version": "0.0.5",
   "license": "MIT",
   "main": "dist/index.js",
   "browser": "browserDist/solid-auth-fetcher.bundle.js",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@inrupt/solid-auth-fetcher",
   "version": "0.0.4",
   "license": "MIT",
-  "main": "dist/index",
+  "main": "dist/index.js",
   "browser": "browserDist/solid-auth-fetcher.bundle.js",
   "types": "dist/index",
   "files": [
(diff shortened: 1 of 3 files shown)


NSeydoux:
I tried this in my local environment, and following the steps in #122, the problem persists. Can you reproduce locally to verify if it's specific to my setup ?

ajacksified:
> I tried this in my local environment, and following the steps in #122, the problem persists. Can you reproduce locally to verify if it's specific to my setup ? I built and published the code in this branch to a local verdaccio, and installed the modified version from this local registry.

Tested with CRA with the latest change - should be good now.

ajacksified:
Latest exports `solidAuthFetcher` in browserDist, and exports a commonjs module in `dist`.

NSeydoux:
Out of curiosity, why was the presence of a `library` field an issue ?

ajacksified:
> Out of curiosity, why was the presence of a `library` field an issue ?

Before that change, you had to `import { solidAuthFetcher } from '@inrupt/solid-auth-fetcher';`. The `library` was being used as part of the exports, essentially doing this: `module.exports = { [library]: solidAuthFetcher }` instead of `module.exports = solidAuthFetcher`.

NSeydoux:
Thanks !

ajacksified:
Should I bump the version in package.json and release to npm when I merge?

NSeydoux:
> Should I bump the version in package.json and release to npm when I merge?

I'd say so

ajacksified:
Published [@0.0.5](https://www.npmjs.com/package/@inrupt/solid-auth-fetcher/v/0.0.5)

PR #135, opened 2020-05-28
Disabled Whitesource CI status check

NSeydoux:
There is a bug in Whitesource that makes it impractical for CI status checks. However, this change still runs a Whitesource scan on push to master, so that our dashboard stays in sync with the state of the repository. It won't affect the CI pipeline, and might introduce false positives that we can manually dismiss, but it still provides some feedback on security issues.

### DIFF
--- a/.github/workflows/whitesource.yml
+++ b/.github/workflows/whitesource.yml
@@ -3,7 +3,8 @@ name: Whitesource Security Scan
 on:
   # The config is set up to verify compliance on pull requests, and to update the
   # whitesource inventory on push to the master branch.
-  pull_request:
+  # NOTE: Due to a bug in Whitesource, it is currently disabled for PRs.
+  # pull_request:
   push:
     branches:
       - master


NSeydoux:
Unfortunately there is not. The ticket itself is private, and so far the observed issues haven't been linked to a documented Whitesource bug.

acoburn:
Not sure this is relevant for this repo, but I have noticed some flakiness in WS scans in the ESS repo (licenses flagged that are incorrect). I don't know if the flakiness you're seeing is the same that I've seen, but rather than disabling the scan entirely, I just (a) don't make the passing scan required (this is a setting in GH) and (b) for failures, it just requires some additional checks on our side.

NSeydoux:
Initially we disabled the scan as a required status check, but even if it is not required, it prevents from merging dependabot PRs, which adds non-neglectible manual upkeep, and the failures were quite systematic on some libraries, so each and every PR would have failing CI checks. Even if not required, I find it problematic enough that I'd rather disable it for the time being and at least figure out if we can prevent the systematic failures before having it as part of the CI.

acoburn:
@NSeydoux that's definitely a reasonable argument :-)
Eventually, we will have to make a decision on whether WhiteSource is the right tool for the job, and so it would be good to flag these sorts of issues.

ajacksified:
In the meantime, we could add `npm audit` in the CI pipeline so we have security auditing. I can make a draft PR to show what that looks like.

emmettownsend:
Should I call the Whitesource account manager back and tell him we are considering moving to another tool because the defects are problematic and slowing us down. We may consider staying if there is transparency around when they plan to fix the defects

Is that worth doing or are they responsive and fixing when we encounter defects?

NSeydoux:
Their support is quite responsive, but I've been unable to understand fully so far if what we were experiencing were bugs from Whitesource, or misconfigurations due to the lack of clarity of their documentation. In the past, when a bug was clearly identified, they notified us very transparently that it was being worked on, so I don't think pressuring them is required yet. I'm still in (loose) contact with the support to understand exactly where the issue is.

Issue #173, opened 2020-06-10
`cross-fetch` overrides `window.fetch` with non-standard behaviour


NSeydoux:
**Bug description**

When using `solid-auth-fetcher` in a modern browser, the used fetcher should be provided by the browser itself. However, it seems that the polyfill returned by `cross-fetch` does not default to `window.fetch`, and has non-standard behaviour. 

What prompted reporting this bug specifically is the fact that the obtained `Response` doesn't have a `.body` method to access its content as a `ReadableStream`.

**To Reproduce**
1. If you don't have a React sandbox, create one (`npx create-react-app my-app`)
2. Install `cross-fetch`
3. Make sure `useEffect` is imported in `App.js`
4. import the `cross-fetch` fetcher: `import {fetch as crossFetch} from "cross-fetch"`
4. Add this code snippet to the `App` function:
```javascript
useEffect(() => {
     const buildFetcher = async () => {
      let fetcher;
      if(typeof window !== undefined && typeof window.fetch !== undefined) {
        console.log("Using the window fetcher")
        fetcher = window.fetch;
      } else {
        console.log("Using cross-fetch")
        fetcher = crossFetch;
      }
      fetcher("https://ruben.verborgh.org/profile/")
      .then(response => response.body)
      .then(stream => stream.getReader())
      .then(reader => {
        return reader.read().then(function processText({ done, value }) {
          if (done) {
            console.log("Stream complete");
            return;
          }
          console.log(value);
          return reader.read().then(processText);
        });
      });
    }
    buildFetcher();
  });
```
5. Run the app: `npm run start`

**Expected behavior**

- The code prints "Using the window fetcher", and then the stream content (byte arrays). This is what happens with the provided code snippet.
- When replacing `fetcher("https://ruben.verborgh.org/profile/")` with `crossFetch("https://ruben.verborgh.org/profile/")`, to force using the polyfill, then an error occurs, "TypeError: stream is undefined", because the response returned by `cross-fetch` does not implement the `body` method.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser: Firefox 77.0.1, Chrome 83

**Additional context**

LDflex expects the response to be readable as a stream. Also, this might have been the root cause of #88, and all the discussion around non-standard headers.

RubenVerborgh:
Can/should we just webpack `cross-fetch` out?

scenaristeur:
[...]

PR #174, opened 2020-06-10
Defaults to a browser fetch instead of `cross-fetch`

NSeydoux:
If a browser fetch implementation is available, then it is used instead of `cross-fetch`. This is because cross-fetch uses a polyfill for fetch, which is incomplete and lacks some standard behaviour that is required in some of our use cases, such as reading responses as streams for LDflex.

This PR contributes to fixing bug #173. However, the problem persists in a `Node` environment.

### DIFF
--- a/__tests__/util/Fetcher.spec.ts
+++ b/__tests__/util/Fetcher.spec.ts
@@ -43,3 +43,19 @@ describe("UuidGenerator", () => {
     );
   });
 });
+
+describe("Browser behaviour", () => {
+  it("should default to the environment's fetch if available", async () => {
+    const mockCrossFetch = jest.requireMock("cross-fetch");
+    mockCrossFetch.mockReturnValueOnce(Promise.resolve("cross-fetch response"));
+    window.fetch = jest
+      .fn()
+      .mockReturnValueOnce(Promise.resolve("browser fetch response"));
+
+    const fetcher = new Fetcher();
+    expect(await fetcher.fetch("https://someurl.com")).toBe(
+      "browser fetch response"
+    );
+    expect(mockCrossFetch.mock.calls.length).toEqual(0);
+  });
+});
--- a/src/util/Fetcher.ts
+++ b/src/util/Fetcher.ts
@@ -32,6 +32,9 @@ export interface IFetcher {
 export default class Fetcher implements IFetcher {
   async fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response> {
     const fetchUrl = url instanceof URL ? url.toString() : url;
+    if (typeof window !== undefined && typeof window.fetch !== undefined) {
+      return window.fetch(fetchUrl, init);
+    }
     return _fetch(fetchUrl, init);
   }
 }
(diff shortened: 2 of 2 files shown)


src/util/Fetcher.ts
```diff
@@ -32,6 +32,9 @@ export interface IFetcher {
 export default class Fetcher implements IFetcher {
   async fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response> {
     const fetchUrl = url instanceof URL ? url.toString() : url;
+    if (typeof window !== "undefined" && typeof window.fetch !== "undefined") {
+      return window.fetch(fetchUrl, init);
+    }
```
RubenVerborgh, line 37: I would rather make this decision at the point where `_fetch` is defined. So redefine an internal `fetch` at the top level, and do this check only once. Also, `typeof window.fetch === "function"`

src/util/Fetcher.ts
```diff
@@ -32,6 +32,9 @@ export interface IFetcher {
 export default class Fetcher implements IFetcher {
   async fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response> {
     const fetchUrl = url instanceof URL ? url.toString() : url;
```
RubenVerborgh, line 34: Also, who are we doing this `toString` for? Is that also cross-fetch?

__tests__/util/Fetcher.spec.ts
```diff
@@ -43,3 +43,19 @@ describe("UuidGenerator", () => {
...
+    mockCrossFetch.mockReturnValueOnce(Promise.resolve("cross-fetch response"));
+    window.fetch = jest
+      .fn()
+      .mockReturnValueOnce(Promise.resolve("browser fetch response"));
+
+    const fetcher = new Fetcher();
+    expect(await fetcher.fetch("https://someurl.com")).toBe(
+      "browser fetch response"
+    );
+    expect(mockCrossFetch.mock.calls.length).toEqual(0);
```
Vinnl, line 59: In anticipation of the ESLint config Jack created:

```suggestion
    expect(mockCrossFetch.mock.calls).toHaveLength(0);
```

NSeydoux:
That's a very good point. If `LDflex` requires `Response` to be readable as streams, and that only works with the browser native fetch (at least not with `cross-fetch`), then is `LDflex` currently supported in IE ? Is there a polyfill used there that fixes the problem ?

RubenVerborgh:
On the topic of IE support, the question is simple: are we testing on IE?
If not, we can offer no guarantees whatsoever.

NSeydoux:
We are not at the moment (that I'm aware of), but Jackson is building an end-to-end test suite next week. I'll ask if IE is included.

Vinnl:
Hmm, but if this is only guaranteed to break in IE _when using LDflex_ (or another library that depends on Header Streams), then it might not be so bad. It's fine as LDflex does not work on IE (as in, that's its prerogative), but _we_ have not committed to that yet.

(There's a difference between not guaranteeing to work, versus guaranteeing not to work. The latter closes the door to IE support _now_, whereas the former keeps the door ajar to at least "best effort" support.)

src/util/Fetcher.ts
NSeydoux: I merged this PR a bit prematurely, and forgot to add this. https://github.com/inrupt/solid-auth-fetcher/pull/183 addresses this comment.

src/util/Fetcher.ts
NSeydoux: The standard `fetch` signature does not include the `URL` type, which is accepted here, so I assume that's the motivation.

Issue #176, opened 2020-06-10
Remove LDFlex/Comunica-specific workarounds from the source code


RubenVerborgh:
**Describe the bug**
https://github.com/inrupt/solid-auth-fetcher/pull/121#issuecomment-642101872

**Steps to solve**
- Does or doesn't LDflex-Comunica pass a correct `Headers` object?

NSeydoux:
After some investigation, I'd say it does not. The `Headers` spec is the following: https://fetch.spec.whatwg.org/#headers-class, and in particular you'll notice that the interface is marked as `iterable`, which means (according to https://heycam.github.io/webidl/#dfn-value-pairs-to-iterate-over) that, among other methods, it should expose a `Symbol.iterator` in order to support structures such as `for...of`. 

It is not the case of the `Headers` object passed by `LDflex-Comunica`, as far as I can tell. It happens that it is the way `node-fetch` uses to look up the headers, which prompted the issue in the first place.

It really doesn't help that `lib.dom.d.ts` types also do their own thing that don't align with the spec either. i implemented an initial fix in #177, but even less workaround would be required if the `LDflex-Comunica` `Headers` object was an iterable.

PR #177, opened 2020-06-10
Clean the LDflex workaround up

NSeydoux:
The previous workaround was unnecessarily convoluted, and a simpler, more standard approach is supported. However, there is still a conflict between the `lib.dom.d.ts` Headers type and the reference https://fetch.spec.whatwg.org/#headers-class. An issue will be opened to track this.

This PR contributes to fixing bug #176.

- [x] I've added a unit test to test for potential regressions of this bug. (actually I removed the LDflex-Comunica-specific test)

### DIFF
--- a/__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
+++ b/__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
@@ -39,24 +39,4 @@ describe("Headers interoperability function", () => {
       ["content-type", "text/turtle"]
     ]);
   });
-
-  it("supports Comunica-style headers structures", () => {
-    // This enables testing that an object similar to the prototypes passed by
-    // Comunica is supported.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const myHeaders: any = {};
-    myHeaders["accept"] = "application/json";
-    myHeaders["content-type"] = "text/turtle";
-    myHeaders["raw"] = (): Record<string, string> => {
-      return {
-        accept: "application/json",
-        "content-type": "text/turtle"
-      };
-    };
-    const flatHeaders = flattenHeaders(myHeaders);
-    expect(Object.entries(flatHeaders)).toEqual([
-      ["accept", "application/json"],
-      ["content-type", "text/turtle"]
-    ]);
-  });
 });
--- a/__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
+++ b/__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
@@ -39,4 +39,20 @@ describe("Headers interoperability function", () => {
       ["content-type", "text/turtle"]
     ]);
   });
+
+  it("supports non-iterable headers if they provide a reasonably standard way of browsing them", () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const myHeaders: any = {};
+    myHeaders["forEach"] = (
+      callback: (value: string, key: string) => void
+    ): void => {
+      callback("application/json", "accept");
+      callback("text/turtle", "content-type");
+    };
+    const flatHeaders = flattenHeaders(myHeaders);
+    expect(Object.entries(flatHeaders)).toEqual([
+      ["accept", "application/json"],
+      ["content-type", "text/turtle"]
+    ]);
+  });
 });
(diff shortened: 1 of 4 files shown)


src/authenticatedFetch/unauthenticated/UnauthenticatedFetcher.ts
```diff
@@ -67,6 +68,9 @@ export default class DpopAuthenticatedFetcher implements IAuthenticatedFetcher {
     url: RequestInfo,
     requestInit?: RequestInit
   ): Promise<Response> {
+    if (requestInit?.headers) {
+      console.log("Received headers");
```
RubenVerborgh, line 72: To be removed

src/authenticatedFetch/unauthenticated/UnauthenticatedFetcher.ts
```diff
@@ -30,6 +30,7 @@ import { IDpopHeaderCreator } from "../../dpop/DpopHeaderCreator";
 import { IUrlRepresentationConverter } from "../../util/UrlRepresenationConverter";
 import { IStorageUtility } from "../../localStorage/StorageUtility";
 import { flattenHeaders } from "../headers/HeadersUtils";
+import { Headers as NodeHeaders } from "node-fetch";
```
RubenVerborgh, line 33: Not used?

src/authenticatedFetch/headers/HeadersUtils.ts
```diff
@@ -67,34 +28,23 @@ function hasRawGetter(
...
-      return headersToFlatten as Record<string, string>;
-    }
-  } else {
-    // headersToFlatten.keys() SHOULD be valid as per the Headers spec. This seems to be an issue
-    // in lib.dom.d.ts that will need some investigation and potentially a contrib upstream.
-    // eslint-disable-next-line
+
+  // If the headers are already a Record<string, string>,
+  // they can directly be returned.
+  if (typeof headersToFlatten.forEach === "undefined") {
```
RubenVerborgh, line 40: In general (I've seen this a couple of times in this library), you might want to explicitly check for the desired type. So in this case `typeof headersToFlatten.forEach !== "function"`.

src/authenticatedFetch/unauthenticated/UnauthenticatedFetcher.ts
RubenVerborgh: Also, a linter rule likely should have caught this.

RubenVerborgh:
Looking into this at https://github.com/comunica/comunica/issues/672; however, it seems that we are just using existing libraries, so more consumers with the same bug might arrive. So might be good to keep this functionality around, even if Comunica gets a workaround.

__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
```diff
@@ -39,24 +39,4 @@ describe("Headers interoperability function", () => {
...
-        accept: "application/json",
-        "content-type": "text/turtle"
-      };
-    };
-    const flatHeaders = flattenHeaders(myHeaders);
-    expect(Object.entries(flatHeaders)).toEqual([
-      ["accept", "application/json"],
-      ["content-type", "text/turtle"]
-    ]);
-  });
```
RubenVerborgh, line 61: Shall we have a test though for `non-iterable Headers-like structure` or something similar? As it might not be uncommon (https://github.com/inrupt/solid-auth-fetcher/pull/177#issuecomment-642501972).

src/authenticatedFetch/unauthenticated/UnauthenticatedFetcher.ts
NSeydoux: Resolved in d084878001cd35bdca8cc97b48d46423fad32c1a.

src/authenticatedFetch/headers/HeadersUtils.ts
NSeydoux: Resolved in 66114c1a61050c4e88d093cd04c73eb24b7f434e

src/authenticatedFetch/unauthenticated/UnauthenticatedFetcher.ts
NSeydoux: Resolved in d084878001cd35bdca8cc97b48d46423fad32c1a. And agreed, enforcing the same linter rules on all our projects is a task that's being completed right now.

__tests__/authenticatedFetch/headers/HeadersUtils.spec.ts
NSeydoux: That's a good point. Added in 29d8f44bc63aeaf91ebb1bd0b8ab48604311eb07

PR #183, opened 2020-06-15
Select the fetcher based on the environment only once

NSeydoux:
The appropriate fetch method is selected when constructing the Fetcher object, instead of running the check at each query.

### DIFF
--- a/src/util/Fetcher.ts
+++ b/src/util/Fetcher.ts
@@ -30,11 +30,18 @@ export interface IFetcher {
 }
 
 export default class Fetcher implements IFetcher {
+  private fetcher: typeof _fetch;
+
+  constructor() {
+    if (typeof window !== "undefined" && typeof window.fetch === "function") {
+      this.fetcher = window.fetch;
+    } else {
+      this.fetcher = _fetch;
+    }
+  }
+
   async fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response> {
     const fetchUrl = url instanceof URL ? url.toString() : url;
-    if (typeof window !== "undefined" && typeof window.fetch !== "undefined") {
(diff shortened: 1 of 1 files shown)


src/util/Fetcher.ts
```diff
@@ -30,11 +30,18 @@ export interface IFetcher {
 }
 
 export default class Fetcher implements IFetcher {
+  private fetcher: typeof _fetch;
+
+  constructor() {
+    if (typeof window !== "undefined" && typeof window.fetch === "function") {
```
RubenVerborgh, line 36: Or even above the level of `Fetcher` itself, but I take it there will not be a lot of instantiations?

src/util/Fetcher.ts
```diff
@@ -30,11 +30,18 @@ export interface IFetcher {
...
+      this.fetcher = window.fetch;
+    } else {
+      this.fetcher = _fetch;
+    }
+  }
+
   async fetch(url: RequestInfo | URL, init?: RequestInit): Promise<Response> {
```
RubenVerborgh, line 43: I still find it slightly weird that we decide here to offer other parameters than the regular `fetch` method; I don't think it's our task to also accept `URL`. From a drop-in replacement, you'd expect that it works no differently from the original.

pmcb55: Yeah - I'm no expert in this area, but I agree the method signature should match standard fetch() **_exactly_** (if possible): https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch. Is there a specific reason to diverge?

NSeydoux:
Due to the architectural changes leading the lib not to be isomorphic anymore, this is no longer relevant.

PR #233, opened 2020-07-31
Restructures the documentation to have smaller pages

NSeydoux:
All of the documentation was in a single README page, which made it harder to navigate. This splits the documentation into smaller pieces, and removes the elements not meant for beta.

# Checklist


docs/api.md
```diff
@@ -0,0 +1,263 @@
+## Single Session API
+
+### login(options) => Promise:void
+Kick off the login process for the user:
+
+```typescript
+import { login } from '@inrupt/solid-client-authn-browser';
```
ajacksified, line 7: FYI, the current package does not export a `login` function: https://github.com/inrupt/solid-client-authn/blob/master/src/index.ts

NSeydoux: Fixed in 569e012104773ba0e4075b9e821c24ce327e95f7, along with other issues with the former API reference.

examples/single/bundle/src/App.js
```diff
@@ -24,15 +24,15 @@ import "regenerator-runtime/runtime";
 
 import {
   Session,
-  getAuthFetcherWithDependencies
+  getClientAuthnWithDependencies
```
pmcb55, line 27: I don't think we should be using `Authn` in our function names - i.e. this should be `getClientAuthenticationWithDependencies()`. (But we can change in another PR - ticket: https://inrupt.atlassian.net/browse/SDK-1181).

PR #234, opened 2020-07-31
Update README.md

pmcb55:
Just suggested wording tweaks...


NSeydoux:
Thanks for the input @jeff-zucker ! What do you think of the phrasing in this commit: 9ea8d30c590afd345b796311bf8eedf631bbfb3e

NSeydoux:
Closing this, as I applied the suggestions in 0604705c1e1e0a35f4cc5969934fcc40f73c1add to simplify merging.

jeff-zucker:
That works for me.  It leaves out actual apps like Data-Kitchen and Solid-shell, but gives enough of a hint that it doesn't matter.   Thanks!

NSeydoux:
No you're right, I didn't think of these use cases at first. I updated to the phrasing you suggested in 30b7ca74ed73d6f293cee3ca53eb54edbe1329ce.

PR #244, opened 2020-08-05
Fix/nss interoperability

NSeydoux:
This PR fixes an interoperability bug with NSS, but it re-introduces a security issue. It should be fixed urgently.

### DIFF
--- a/examples/single/bundle/src/App.js
+++ b/examples/single/bundle/src/App.js
@@ -38,7 +38,7 @@ class App extends Component {
     );
     this.state = {
       status: "loading",
-      loginIssuer: "https://broker.demo-ess.inrupt.com",
+      loginIssuer: "https://nseydoux.inrupt.net",
       fetchRoute: "",
       fetchBody: "",
       session: session,
@@ -59,7 +59,9 @@ class App extends Component {
     } else if (this.state.session.isLoggedIn) {
       this.setState({ status: "dashboard", session });
     } else {
-      const authCode = new URL(window.location.href).searchParams.get("code");
+      const authCode =
+        new URL(window.location.href).searchParams.get("code") ||
+        new URL(window.location.href).searchParams.get("access_token");
       if (!authCode) {
         this.setState({ status: "login" });
       } else {
--- a/examples/single/bundle/src/App.js
+++ b/examples/single/bundle/src/App.js
@@ -38,7 +38,7 @@ class App extends Component {
     );
     this.state = {
       status: "loading",
-      loginIssuer: "https://nseydoux.inrupt.net",
+      loginIssuer: "https://inrupt.net",
       fetchRoute: "",
       fetchBody: "",
       session: session,
@@ -59,6 +59,7 @@ class App extends Component {
     } else if (this.state.session.isLoggedIn) {
       this.setState({ status: "dashboard", session });
     } else {
+      // Depending on which flow login uses, the response will either be "code" or "access_token".
       const authCode =
         new URL(window.location.href).searchParams.get("code") ||
         new URL(window.location.href).searchParams.get("access_token");
--- a/examples/single/bundle/src/App.js
+++ b/examples/single/bundle/src/App.js
@@ -62,6 +62,8 @@ class App extends Component {
       // Depending on which flow login uses, the response will either be "code" or "access_token".
       const authCode =
         new URL(window.location.href).searchParams.get("code") ||
+        // FIXME: Temporarily handle both autch code and implicit flow.
+        // Should be either removved or refactored.
         new URL(window.location.href).searchParams.get("access_token");
       if (!authCode) {
         this.setState({ status: "login" });
--- a/src/dpop/DpopClientKeyManager.ts
+++ b/src/dpop/DpopClientKeyManager.ts
@@ -64,6 +64,8 @@ export default class DpopClientKeyManager implements IDpopClientKeyManager {
         alg: "EC",
         use: "sig"
       });
+
+      // FIXME: Temporarily use insecure storage while the implicit auth flow is requried.
       await this.storageUtility.set(
         this.getLocalStorageKey(),
         JSON.stringify(jwk),
@@ -74,6 +76,8 @@ export default class DpopClientKeyManager implements IDpopClientKeyManager {
 
   async getClientKey(): Promise<JSONWebKey | undefined> {
     let keyString;
+
+    // FIXME: Temporarily use insecure storage while the implicit auth flow is requried.
     try {
       keyString = await this.storageUtility.get(this.getLocalStorageKey(), {
         secure: false
--- a/src/dpop/DpopClientKeyManager.ts
+++ b/src/dpop/DpopClientKeyManager.ts
@@ -69,6 +69,7 @@ export default class DpopClientKeyManager implements IDpopClientKeyManager {
       await this.storageUtility.set(
         this.getLocalStorageKey(),
         JSON.stringify(jwk),
+        // FIXME: implicit flow fails if the DPoP key is stored in the "secure" storage.
         { secure: false }
       );
     }
--- a/src/dpop/DpopClientKeyManager.ts
+++ b/src/dpop/DpopClientKeyManager.ts
@@ -67,7 +67,7 @@ export default class DpopClientKeyManager implements IDpopClientKeyManager {
       await this.storageUtility.set(
         this.getLocalStorageKey(),
         JSON.stringify(jwk),
-        { secure: true }
+        { secure: false }
       );
     }
   }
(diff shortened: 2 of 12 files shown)


examples/single/bundle/src/App.js
```diff
@@ -38,7 +38,7 @@ class App extends Component {
     );
     this.state = {
       status: "loading",
-      loginIssuer: "https://broker.demo-ess.inrupt.com",
+      loginIssuer: "https://nseydoux.inrupt.net",
```
ajacksified, line 41: Should this be the demo?

src/dpop/DpopClientKeyManager.ts
```diff
@@ -67,7 +67,7 @@ export default class DpopClientKeyManager implements IDpopClientKeyManager {
       await this.storageUtility.set(
         this.getLocalStorageKey(),
         JSON.stringify(jwk),
-        { secure: true }
+        { secure: false }
```
ajacksified, line 70: I'd add a `TODO` comment here to explain why.

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
```diff
@@ -38,15 +40,22 @@ export default class LegacyImplicitFlowOidcHandler implements IOidcHandler {
...
   ) {}
 
   async canHandle(oidcLoginOptions: IOidcOptions): Promise<boolean> {
     return !!(
       oidcLoginOptions.issuerConfiguration.grantTypesSupported &&
       oidcLoginOptions.issuerConfiguration.grantTypesSupported.indexOf(
         "implicit"
-      ) > -1
+      ) > -1 &&
+      // Escape hatch to detect that we are talking to NSS and not the id broker
```
pmcb55, line 55: Yike! Yeah, big TODO: or FIXME: needed here (as with Jack's comment above on `secure: false` (which occurs twice)).

src/login/oidc/redirectHandler/TokenSaver.ts
```diff
@@ -43,43 +42,33 @@ export default class TokenSaver implements ITokenSaver {
...
-  ): Promise<ISessionInfo> {
-    throw new Error("Not Implemented");
-    // const decoded = await this.joseUtility.decodeJWT(
-    //   // TODO this should actually be the id_vc of the token
-    //   accessToken as string
-    // );
-    // // TODO validate decoded token
-    // // TODO extract the localUserId from state and put it in the session
+  ): Promise<void> {
+    // throw new Error("Not Implemented");
```
pmcb55, line 50: No need to keep this anymore.

examples/single/bundle/src/App.js
```diff
@@ -59,7 +59,9 @@ class App extends Component {
     } else if (this.state.session.isLoggedIn) {
       this.setState({ status: "dashboard", session });
     } else {
-      const authCode = new URL(window.location.href).searchParams.get("code");
+      const authCode =
+        new URL(window.location.href).searchParams.get("code") ||
```
pmcb55, line 63: A comment here would be good (is it just that NSS and ESS use different param names for the same auth token value...?).

examples/single/bundle/src/App.js
NSeydoux: No right, it should either be `inrupt.net` or `https://broker.demo-ess.inrupt.com`

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
```diff
@@ -38,15 +40,22 @@ export default class LegacyImplicitFlowOidcHandler implements IOidcHandler {
...
+      oidcLoginOptions.issuerConfiguration.grantTypesSupported.indexOf(
+        "urn:ietf:params:oauth:grant-type:jwt-bearer"
```
acoburn, line 57: This is a very brittle comparison.

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
```diff
@@ -60,11 +69,21 @@ export default class LegacyImplicitFlowOidcHandler implements IOidcHandler {
...
-      redirect_url: oidcLoginOptions.redirectUrl.toString(),
+      redirect_uri: oidcLoginOptions.redirectUrl.toString(),
       // The webid scope does not appear in the spec
       scope: "openid webid offline_access",
       state: oidcLoginOptions.sessionId
     };
+
+    await this.dpopClientKeyManager.generateClientKeyIfNotAlready();
+    await this.storageUtility.setForUser(
+      oidcLoginOptions.sessionId,
```
acoburn, line 80: Run these in parallel?

src/login/oidc/redirectHandler/TokenSaver.ts
```diff
@@ -43,43 +42,33 @@ export default class TokenSaver implements ITokenSaver {
...
-    // );
-    // return session;
+    await this.storageUtility.setForUser(
+      sessionId,
+      {
+        accessToken: accessToken as string,
+        webId: decoded.sub as string,
+        idToken: idToken as string,
+        refreshToken: refreshToken as string,
+        isLoggedIn: "true"
```
acoburn, line 69: Are you storing access tokens in session storage? That would be bad

examples/single/bundle/src/App.js
acoburn: I don't really like this. You have code handling authorization_code AND implicit flow as part of the same logical construct. It would be much better, design-wise to separate the different flows into discrete functions. Otherwise you end up conflating too many things right here

NSeydoux: This is more a temporary fix to be able to run tests than anything else. If things work the way I hope, we should be able to use the auth code flow on NSS as well, and just remove this because the demo would always use the auth code flow. Otherwise, i'll clean up the demo to be more structured.

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
NSeydoux: Ideally I'd like to figure out what goes wrong with the auth code flow on NSS, and then this wouldn't be necessary. If the need to use implicit flow on NSS and auth code flow on the broker lasts, a better design (read: any design really) will be needed to replace this.

src/login/oidc/redirectHandler/TokenSaver.ts
NSeydoux: Indeed, this is the first issue I want to look at. For some reason, storing them in memory doesn't work with the implicit flow, and anyway a refactor to keep them in a closure is what we really want.

examples/single/bundle/src/App.js
NSeydoux: Fixed in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/dpop/DpopClientKeyManager.ts
NSeydoux: Addressed in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
NSeydoux: Addressed in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/login/oidc/redirectHandler/TokenSaver.ts
NSeydoux: Fixed in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

examples/single/bundle/src/App.js
NSeydoux: Acknowledged in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
NSeydoux: Acknowledged in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/login/oidc/oidcHandlers/LegacyImplicitFlowOidcHandler.ts
NSeydoux: Fixed in e65ae11bb5e81297c1fbc3e19e0219f4d3b5282d

src/login/oidc/redirectHandler/TokenSaver.ts
NSeydoux: Actually no, the `secure: true` part means that the access token is stored in an in-memory store, not the session storage. The bad news is the DPoP token is stored in the session storage currently, so it's not like it's any better.

PR #258, opened 2020-08-10
Fix client registration issue in the auth code flow on NSS

NSeydoux:
For some reason, dynamic client registration information isn't persisted when interacting with NSS if stored in secure storage. If stored in insecure storage though, it works and prevents us from having to deal with the legacy token.

- [ ] I've added a unit test to test for potential regressions of this bug. (will do when investigating to secure the storage)

### DIFF
--- a/examples/single/bundle/src/App.js
+++ b/examples/single/bundle/src/App.js
@@ -83,14 +83,10 @@ class App extends Component {
   async handleLogin(e, isPopup = false) {
     e.preventDefault();
     this.setState({ status: "loading" });
-    this.state.session
-      .login({
-        redirectUrl: new URL("http://localhost:3001/"),
-        oidcIssuer: new URL(this.state.loginIssuer)
-      })
-      .then(() => {
-        this.setState({ status: "dashboard" });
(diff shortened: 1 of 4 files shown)


ajacksified:
Requests for private resources on both NSS and ESS seem to be failing in the example app.

ajacksified:
If DpopClientKeyManager was refactored to accept a `secure = true` parameter in its constructor, we could at least isolate the security issues to the `LegacyImplicitFlowOidcHandler`.

NSeydoux:
Unfortunately, not only the implicit flow that is impacted: this PR makes it so that we don't use the `LegacyImplicitFlowHandler` anymore, and always use the `AuthorizationCodeWithPkceOidcHandler`. Moreover, the `ClientRegistrar` is orthogonal to the flow that gets used, and it's this piece of code that needs to store the client secret in the non-secure storage when used against NSS.

Basically right now there are multiple items that are stored in the non-secure storage and that should not be:
- The DPoP key used to sign the requests
- The client secret that is returned after the client registration process

PR #266, opened 2020-08-13
Redirect user to IdP logout page

NSeydoux:
This adds a new step to redirect the user when he/she is logging out, so that they can complete a proper OIDC session ending flow.

# Checklist

- [ ] ~Relevant documentation, if any, has been written/updated.~ Generating the API reference will be done later this sprint 

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### New features
 
+- Perform a [spec-compliant](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) log out from the identity provider.
+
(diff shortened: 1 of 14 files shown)


NSeydoux:
I had a different understanding of that conversation: I was talking about logging out of the broker not logging you out of Google. However, I see your point: logging out of the broker is similar to looging out of the NSS IdP, which would log your browser out of all your Solid apps, which is not desirable. 

What do you think if I add a `soft` option, true by default, which logs you out of the app but not of the IdP ? This way it would be easy to provide both a "Log out of the App" and "Log out of Solid" buttons.

Vinnl:
@NSeydoux That sounds reasonable to me, but maybe verify with @nicolasmondada that that's worth spending time on. (Unless it's very little work.)

NSeydoux:
The RP-initiated logout is a feature that won't be included in the library for the time being.

PR #272, opened 2020-08-18
Set up infrastructure for doc and API doc

NSeydoux:
This adds the sphinx infrastructure to generate both the doc and the API doc. No doc or API doc content should be changed in this PR. This is based on the existing setup after API doc split (https://github.com/inrupt/solid-client-js/pull/367)

# Checklist


.github/publish-website.yml
```diff
@@ -0,0 +1,99 @@
+name: Publish website
```
Vinnl, line 1: This doesn't seem to have run (I don't know why), but it wouldn't work anyway: there are two jobs here that both push to the gh-pages branch, one overwriting the other. I did some additional work on solid-client to package them both together into a single deployment: https://github.com/inrupt/solid-client-js/commit/2f27f14fb172628960e1b6f976fd23e0284d33be

PR #274, opened 2020-08-19
Mark all non-public classes as @internal

NSeydoux:
Most of the code in this module should not be part of the public documentation, as the users are only expected to deal with the public API. This marks all the files that should not be exposed as internal.

# Checklist

- [ ] All acceptance criteria are met. (Another PR will complete the content for the classes and interfaces that are actually publicly documented)


NSeydoux:
The API doc looks like this after the PR: api_doc

Vinnl:
Before I go through this in detail: the `stripInternal` TypeScript options allows you to remove functions from the public type definitions when they're marked with `@internal`. However, this is a somewhat risky endeavour: if, for example, you mark a type as `@internal` that played a role in another function's return type, then that return type will look like `any` for the consumer, i.e. disabling all type checking.

Thus, if the goal is just to have them not listed in the API docs, `@hidden` (or its alias `@ignore`) is probably safer.

Vinnl (approved):
OK, nothing more to add to that other comment :)

NSeydoux:
Right, that would be an issue then, because when the library is split, some of the "internal" classes and types will be reused by environment-specific modules. In f10d32bd759c072ded6df48587855a8dd0ff7394, the `@internal` annotations are changed into `@hidden`, which still yields the desired documentation, but should preserve the type annotations previously marked as internal for consuming libraries.

PR #275, opened 2020-08-19
Documentation of the public API

NSeydoux:
This adds content to the documentation of the public API of `solid-client-authn-browser`.

# Checklist

- [ ] All acceptance criteria are met. (It seems that some additional work is required on the infrastructure of the doc generation pipeline, that will be done in a separate PR)


src/Session.ts
```diff
@@ -78,15 +112,29 @@ export class Session extends EventEmitter {
     this.emit("login");
   };
 
+  /**
+   * Fetches data using available login information. If the user is not logged in, this will behave as a regular `fetch`. The signature of this method is identical to the [canonical `fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).
+   *
+   * @param url The URL from which data should be fetched.
+   * @param init Optional parameters customizing the request, by specifying an HTTP method, headers, a body, etc. Follows the [WHATWG Fetch Standard](https://github.github.io/fetch/).
```
Vinnl, line 119: The other link is to the polyfill :)

```suggestion
   * @param init Optional parameters customizing the request, by specifying an HTTP method, headers, a body, etc. Follows the [WHATWG Fetch Standard](https://fetch.spec.whatwg.org/).
```

src/Session.ts
```diff
@@ -78,15 +112,29 @@ export class Session extends EventEmitter {
...
+  /**
+   * Logs the user out of the application. This does not log the user out of the identity provider, and should not redirect the user away.
+   */
   logout = async (): Promise<void> => {
     await this.clientAuthentication.logout(this.info.sessionId);
     this.emit("logout");
   };
 
+  /**
+   * Completes the login process by receiving the information provided by the identity provider through redirect.
```
Vinnl, line 134:
```suggestion
   * Completes the login process by processing the information provided by the identity provider through redirect.
```

src/Session.ts
```diff
@@ -78,15 +112,29 @@ export class Session extends EventEmitter {
...
+   *
+   * @param url The URL where the user has been redirected, as information is given by the identity provider to the app through query strings.
```
Vinnl, line 136:
```suggestion
   * @param url The URL of the page handling the redirect, including the query parameters — these contain the information to process the login.
```

src/Session.ts
```diff
@@ -100,10 +148,20 @@ export class Session extends EventEmitter {
     return sessionInfo;
   };
 
+  /**
+   * Register a callback function to be called when a user completes login.
+   *
```
Vinnl, line 153:
```suggestion
   * Register a callback function to be called when a user completes login.
   *
   * The callback is called when {@link handleIncomingRedirect} completes successfully.
   *
```

Assuming that this is true? If not, a comment on when it _is_ called would be useful :)

src/SessionManager.ts
```diff
@@ -41,6 +44,19 @@ export class SessionManager extends EventEmitter {
...
+   * Constructor for the SessionManager object. It is typically used as follows:
+   *
+   * ```typescript
+   * import { SessionManager } from "@inrupt/solid-client-authn-browser";
+   * import customStorage from "./myCustomStorage";
+   *
+   * const sessionManager = new SessionManager({
+   *   secureStorage: customStorage
+   * });
+   * ```
```
Vinnl, line 57:
```suggestion
   * ```
   * See {@link IStorage} for more information on how to define your own storage mechanism.
```

Assuming that initialising some `customStorage` is required? If not, a note about when to use it could be useful.

src/SessionManager.ts
```diff
@@ -32,6 +32,9 @@ export interface ISessionManagerOptions {
   insecureStorage?: IStorage;
 }
 
+/**
+ * A SessionManager instance aims at managing all the sessions in an application, each session being associated to an individual user.
```
Vinnl, line 36:
```suggestion
 * A SessionManager instance can be used to manage all the sessions in an application, each session being associated to an individual user.
```

src/SessionManager.ts
```diff
@@ -117,19 +145,39 @@ export class SessionManager extends EventEmitter {
     return session;
   }
 
+  /**
+   * @param sessionId A unique session identifier.
+   * @returns true if a session associated to the given ID exists.
```
Vinnl, line 150:
```suggestion
   * @returns A Promise resolving to true if a session associated to the given ID exists, and false if not.
```

To emphasise that a non-existent Session won't result in the Promise rejecting.

src/SessionManager.ts
```diff
@@ -117,19 +145,39 @@ export class SessionManager extends EventEmitter {
...
+   * @param callback a function executed when a session logs out, with the session as a parameter.
+   */
   onSessionLogout(callback: (session: Session) => unknown): void {
-    this.on("sessionLogin", callback);
+    this.on("sessionLogout", callback);
   }
 
+  /**
+   * Removes a session from the pool managed by the manager.
+   * This does not log out the session, but it does remove references to it, so after this call the session will become unreachable.
```
Vinnl, line 177: Why would you want to do this? (The answer might be a relevant addition to this docblock.)

src/Session.ts
```diff
@@ -34,10 +37,35 @@ export interface ISessionOptions {
...
+ * A {@link Session} object represents a user's session on an application. The session holds state, as it stores information enabling acces to private resources after login for instance.
+ */
 export class Session extends EventEmitter {
+  /**
+   * Information regarding the current session.
+   */
   public readonly info: ISessionInfo;
+
+  /**
+   * @hidden
```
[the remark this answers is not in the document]
Vinnl: Ah haha, you already did that :)

Vinnl (approved):
Just some suggestions, take them or leave them as you see fit :)

src/Session.ts
NSeydoux: It is currently not, but it should be: that's a bug I'm willing to fix ^^

src/SessionManager.ts
NSeydoux: It currently is required, but that constraint may change once the storage happens securely. Unfortunately, until I tackle this issue, it is hard to tell how the API will be impacted.

src/SessionManager.ts
NSeydoux: This is useful after a user has logged out, otherwise your server will have an ever-growing list of active sessions. But indeed, that sounds like useful context ^^

src/ILoginInputOptions.ts
```diff
@@ -22,11 +22,32 @@
...
+  /**
+   * The user's identity provider.
+   */
   oidcIssuer?: URL;
+  /**
+   * The URI within this application that the user should be redirected to after successful login. This can be either a web URL or a mobile URL scheme.
+   */
   redirectUrl?: URL;
+  /**
+   * A ID for your application, **previously registered to the identity provider**. Only if you don't want to do [dynamic registration](https://tools.ietf.org/html/rfc7591).
```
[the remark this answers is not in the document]
NSeydoux: Changed to "Only required if users of your app log in an identity provider known a priori, which only supports a predefined set of apps and prevents [dynamic registration](https://tools.ietf.org/html/rfc7591)."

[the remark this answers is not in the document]
NSeydoux: Fixed in 1152d1ca245cbd3191341d02b777027a5dd94c1b

src/Session.ts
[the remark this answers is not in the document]
NSeydoux: Cleaned up in aa4e74daec85be60b2792ec3518b16848b5e64b9

src/Session.ts
```diff
@@ -34,10 +37,35 @@ export interface ISessionOptions {
...
+   *
+   * ```typescript
+   * const session = new Session(
+   *   {
+   *     clientAuthentication: getClientAuthenticationWithDependencies({})
+   *   },
+   *   "mySession"
+   * );
+   * ```
+   * @param sessionOptions The options enabling the correct instanciation of the session. Either both storages or clientAuthentication are required.
```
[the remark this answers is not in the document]
NSeydoux: Documented in aa4e74daec85be60b2792ec3518b16848b5e64b9

src/SessionManager.ts
NSeydoux: Documented in 9ab620d1620dd73612b5dfa336fdcb71ac2e8925

PR #276, opened 2020-08-19
Docs: Update to build

kay-kim:
Minor fixes to configuration and makefile to build the docs:

- Update API docs conf.py - API docs don't need tabs/copy button extensions (and isn't part of the requirements.txt)
- Update API docs Makefile - when copying to  build directory, update the api docs structure
- Update API docs Makefile - clean .md files in all api directories, not just modules
- Update reference docs Makefile - missing the sphinx-builder line for the html target


NSeydoux (approved):
Thanks for fixing this !

PR #278, opened 2020-08-24
Refactored the docs directories

NSeydoux:
Both api and prose docs are now under the same `docs` directory, as introduced in https://github.com/inrupt/solid-client-js/pull/367/commits/2f27f14fb172628960e1b6f976fd23e0284d33be.


Vinnl (approved):
Cool. Time to add preview deployments via Vercel as well?

PR #277, opened 2020-08-21
Fix typo in SessionManager's detachSession method

travis:
fix typo

This PR fixes bug #.

- [ ] I've added a unit test to test for potential regressions of this bug.
- [x] The changelog has been updated, if applicable.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #282, opened 2020-08-25
Run the CI on PR

NSeydoux:
In order to have the CI run for third-party contributions coming from a fork, they must run on PR, and not only on push, since technically a third-party contribution hasn't been pushed to a branch of the repo.

This should trigger the CI for https://github.com/inrupt/solid-client-authn-js/pull/277.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
+on: [push, pull_request]
 
 env:
   CI: true


Vinnl (approved):
Not much harm in enabling this, I guess, but I don't see why the Actions wouldn't run for the fork. After all, the `.workflows` directory is still present, so it should have run on push to the fork?

For example, you can see that this fork does run an action that's configured to only run on `push`: https://github.com/itme/solid-client-js/actions?query=workflow%3ACI

Vinnl:
Ah, I just forked a repository myself, and got this:

image

I guess running on a PR means that they will run in the context of _our_ repository, where they are enabled properly?

Vinnl:
Given this: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

> Note: Workflows do not run on private base repositories when you open a pull request from a forked repository.

I think this might need to be [`pull_request_target`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target)

> This event is similar to pull_request, except that it runs in the context of the base repository of the pull request, rather than in the merge commit.

Although that might mean that any changes in the PR are not actually included in the checks, as its use case appears to be different:

> For example, this event allows you to create workflows that label and comment on pull requests, based on the contents of the event payload.

NSeydoux:
I think the latter option would be ideal, to make it possible for external contributors to create a branch in the repo, and ask for this branch to be submitted to a PR, but I guess this could be a security issue, if the PR changes the workflows to expose the tokens for instance.

I don't think the first point is an issue for us, as the repository isn't private (I assume private here means the repo isn't visible, not that it hasn't public write access). I think as long as no secrets are made available in the workflow, which is the case for `ci`,  `pull_request` seems to be more suitable than `pull_request_target`, since we want the tests to run against the code within the PR, and not into the repo. But that's my understanding of `pull_request_target`... I'll create a throwaway account, and try to fork and see what happens, maybe I'll revert this if it doesn't do what I expect.

NSeydoux:
This doesn't work as intended. To run a PR on a forked PR, someone with write privilege must add the forked remote, and re-create a PR from within the repo so that the CI can run.

PR #289, opened 2020-09-02
Clear local storage on login

NSeydoux:
If the user did not log out, and tries to login in a different tab, data stored in local storage prevents expected behaviour, and the login cannot happen. This fixes the issue byy clearing the storage on login as well as on logout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### New features
 
+### Bugfixes
+
+- Browser
+  - Login now clears the local storage, so that you can log into a different server
+even if not logged out properly.
+
 ## [0.1.2] - 2020-09-07
 
 ### Internal refactor:
--- a/packages/browser/__tests__/ClientAuthn.spec.ts
+++ b/packages/browser/__tests__/ClientAuthn.spec.ts
@@ -26,14 +26,8 @@ import {
   RedirectHandlerMock,
   RedirectHandlerResponse,
 } from "../src/login/oidc/redirectHandler/__mocks__/RedirectHandler";
-import {
-  LogoutHandlerMock,
-  mockLogoutHandler,
-} from "../src/logout/__mocks__/LogoutHandler";
-import {
-  SessionInfoManagerMock,
-  SessionCreatorCreateResponse,
-} from "../src/sessionInfo/__mocks__/SessionInfoManager";
+import { LogoutHandlerMock } from "../src/logout/__mocks__/LogoutHandler";
+import { mockSessionInfoManager } from "../src/sessionInfo/__mocks__/SessionInfoManager";
 import {
   AuthenticatedFetcherMock,
   AuthenticatedFetcherResponse,
@@ -48,7 +42,7 @@ describe("ClientAuthentication", () => {
     loginHandler: LoginHandlerMock,
     redirectHandler: RedirectHandlerMock,
     logoutHandler: LogoutHandlerMock,
-    sessionInfoManager: SessionInfoManagerMock,
+    sessionInfoManager: mockSessionInfoManager(mockStorageUtility({})),
     authenticatedFetcher: AuthenticatedFetcherMock,
     environmentDetector: EnvironmentDetectorMock,
   };
@@ -95,7 +89,7 @@ describe("ClientAuthentication", () => {
         { secure: true }
       );
       const clientAuthn = getClientAuthentication({
-        logoutHandler: mockLogoutHandler(nonEmptyStorage),
+        sessionInfoManager: mockSessionInfoManager(nonEmptyStorage),
       });
       await clientAuthn.login("someUser", {
         clientId: "coolApp",
@@ -146,12 +140,24 @@ describe("ClientAuthentication", () => {
 
   describe("getSessionInfo", () => {
     it("creates a session for the global user", async () => {
-      const clientAuthn = getClientAuthentication();
+      const sessionInfo = {
+        isLoggedIn: "true",
+        sessionId: "mySession",
+        webId: "https://pod.com/profile/card#me",
+      };
+      const clientAuthn = getClientAuthentication({
+        sessionInfoManager: mockSessionInfoManager(
+          mockStorageUtility(
+            {
+              mySession: { ...sessionInfo },
+            },
+            true
+          )
+        ),
+      });
       const session = await clientAuthn.getSessionInfo("mySession");
-      expect(session).toBe(SessionCreatorCreateResponse);
-      expect(defaultMocks.sessionInfoManager.get).toHaveBeenCalledWith(
-        "mySession"
-      );
+      // isLoggedIn is stored as a string under the hood, but deserialized as a boolean
+      expect(session).toEqual({ ...sessionInfo, isLoggedIn: true });
     });
   });
 
--- a/packages/browser/__tests__/ClientAuthn.spec.ts
+++ b/packages/browser/__tests__/ClientAuthn.spec.ts
@@ -26,7 +26,10 @@ import {
   RedirectHandlerMock,
   RedirectHandlerResponse,
 } from "../src/login/oidc/redirectHandler/__mocks__/RedirectHandler";
-import { LogoutHandlerMock } from "../src/logout/__mocks__/LogoutHandler";
+import {
+  LogoutHandlerMock,
+  mockLogoutHandler,
+} from "../src/logout/__mocks__/LogoutHandler";
 import {
   SessionInfoManagerMock,
   SessionCreatorCreateResponse,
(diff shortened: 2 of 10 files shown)


__tests__/login/oidc/OidcLoginHandler.spec.ts
```diff
@@ -89,4 +93,30 @@ describe("OidcLoginHandler", () => {
...
+      oidcIssuer: new URL("https://arbitrary.url"),
+      redirectUrl: new URL("https://app.com/redirect"),
+      clientId: "coolApp",
+    });
+    expect(nonEmptyStorage.getForUser("someUser", "someKey", { secure: true }))
+      .toBeUndefined;
+    expect(nonEmptyStorage.getForUser("someUser", "someKey", { secure: false }))
+      .toBeUndefined;
+    // This test is only necessary until the key is stored safely
+    expect(nonEmptyStorage.get("clientKey", { secure: false })).toBeUndefined;
```
Vinnl, line 120: I don't think these actually assert anything right now :P Might be worth enabling the Jest plugin in ESLint if that isn't enabled yet, as I think that will warn you of tests without any assertions.

```suggestion
    expect(nonEmptyStorage.getForUser("someUser", "someKey", { secure: true }))
      .toBeUndefined();
    expect(nonEmptyStorage.getForUser("someUser", "someKey", { secure: false }))
      .toBeUndefined();
    // This test is only necessary until the key is stored safely
    expect(nonEmptyStorage.get("clientKey", { secure: false })).toBeUndefined();
```

Vinnl (approved):
The tests don't actually test anything right now, but other than that it looks good to me. That said, the changelog message is a bit worrying: does that mean that it is possible that you're not actually logged out before logging in? If so, should logging in really log you out anyway, rather than just not doing anything/throwing an error? And if not, maybe the changelog message could be made more reassuring that there's no need to worry about lingering sessions? :)

__tests__/login/oidc/OidcLoginHandler.spec.ts
NSeydoux: Ah good catch, that was actually coming from another test I had written for logout, both fixed in b3a44615fc30e195d9423a8414a929ba7ec1adc1

NSeydoux:
Regarding your comment on lingering sessions: right now, `session.login` doesn't check if the session was previously logged in. This PR makes it so that logging you in effectively logs you out first, because there are currently no mechanism preventing you from logging in twice. If a session which is logged in tries to log in again, should an error be thrown instead ?

Vinnl:
@NSeydoux That's a good question, maybe consult the wider team on that?

pmcb55:
Re. " If a session which is logged in tries to log in again, should an error be thrown instead ?", I think asking the other teams might be a good idea. I don't think it should be an error, but perhaps it has value to re-generate the auth token or something - i.e. perhaps it should be a valid action that should perform a specific action (i.e. an action that should be in the spec so that all clients can rely on it, and that an SDK could enforce - e.g. if a re-login should re-generate a new auth token, then our SDK could throw an error if a new, different auth token is not received).

NSeydoux:
We can put this on hold until Jack is back from his vacation, he is the one who has faced this issue the most often, and resolved it in the layer on top of solid-client-authn-browser. @pmcb55 I this think PR currently goes into the direction you suggest: the previous access token is discarded, and a new one is requested in a brand-new login process. Since the new login might be against a different IdP, that's the safest approach. An alternative would be to throw an error, or to just do nothing, but the latter is the current behaviour, and that has proven to be counter-intuitive in particular when you intend to log into a new IdP.

ajacksified:
To me, it makes sense to call `logout` to clear any previous existing sessions when `login` begins - the developer is explicitly asking for a new session. I think if a developer had a use case for re-generating a token for an existing token, the library could provide a `refresh` or `regenerate` method instead.

PB is doing this the cheap, hacky way of clearing localstorage on login and logout to fix an issue where calling `login` with idp information in localstorage re-uses localstorage rather then the parameters passed in, which seems like another bug that this happens to fix.

pmcb55:
@ajacksified - Ok, I've no problem with the `regenerate()/refresh()` idea.
But just being pedantic, I wouldn't **_literally_** call `logout()` on `login()`, since I'd expect any `logout()` implementation to log the fact that the logout-action was initiated by the user. But in this case the user did not request a `logout()` at all, all they did was a `login()` - so I think the `login()` could call a `clearSessionInfo()` function of something (which would also be called by the `logout()` function of course).

NSeydoux:
The token refresh flow is something that should currently be supported: on `401` after a fetch, the library should try to use a refresh token if available to get a new access token. I don't think there is a need to expose this feature in the API. I'll update the code so that `login` uses `logout`, instead of manually clearing the storage.

src/login/oidc/OidcLoginHandler.ts
```diff
@@ -39,6 +39,7 @@ import IIssuerConfig from "./IIssuerConfig";
 import { IDpopClientKeyManager } from "../../dpop/DpopClientKeyManager";
 import URL from "url-parse";
 import { IClientRegistrar } from "./ClientRegistrar";
+import { IStorageUtility } from "../../storage/StorageUtility";
```
pmcb55, line 42: 'Defined but never used' warnings on this line...? Original code didn't have this, so maybe a holdover from the Tsryinge testing...?!

src/logout/GeneralLogoutHandler.ts
```diff
@@ -28,6 +28,24 @@ import ILogoutHandler from "./ILogoutHandler";
 import { inject, injectable } from "tsyringe";
 import { IStorageUtility } from "../storage/StorageUtility";
 
+/**
+ * This function removes all session-related information from storage.
+ * @param userId the session identifier
+ * @param storage the storage where session info is stored
+ * @hidden
+ */
+export async function clearSession(
```
pmcb55, line 37: Should this function be in the `util/` folder (i.e. it's not really Logout-specific at all really)...?

src/ClientAuthentication.ts
```diff
@@ -69,6 +69,8 @@ export default class ClientAuthentication {
     sessionId: string,
     options: ILoginInputOptions
   ): Promise<void> => {
+    // In order to get a clean start, make sure that the session is logged out on login.
+    await this.logout(sessionId);
```
pmcb55, line 73: So you are calling `logout()` from within the `login()` function? I thought this would just call `clearSession()`...?

src/login/oidc/OidcLoginHandler.ts
NSeydoux: Ah right, I'm surprised the CI did not complain. Fixed in 9335a362c0ca70effbd05e7121a18e9dc8e273f0

src/ClientAuthentication.ts
NSeydoux: Currently logout is basically just calling `clearSession` , but this might change depending of how we store session-related information in the future, and how the underlying libraries that we use implement logout. If `login` implements the session-clearing logic, there will be some redundancy with `logout`.

pmcb55: And actually, perhaps this 'clearing out the current session' logic (which you're doing now by calling a 'logout' function) actually belongs inside the 'login handler' itself, and not here at this 'ClientAuthentication' level...? i.e. maybe I have an implementation of a login handler that wants access to the current session info (if any) before it goes ahead and actually performs it's login voodoo...?

src/logout/GeneralLogoutHandler.ts
NSeydoux: I'm not a big fan of a 'catch-all' `util` module. If clearing the session state is shared in the `util` module, then I see your point about just using `clearState` at the beginning of `login`, but given that a big refactoring is planned, I'd rather share high-level logic (e.g. logout at the beginning of login) rather than more specific implementation (e.g. clearing some specific  information from storage at login and logout).

src/ClientAuthentication.ts
pmcb55: | Currently logout is basically just calling clearSession 
Ok, but that's not really relevant here I don't think. What is the 'logic' we want here for a login() operation - i.e. is it that we want to perform a formal logout operation first, or is it that we want to perform that login with fresh 'state'? i.e. would our documentation state that login actually performs a logout operation first?
|  there will be some redundancy with logout.
But that's why you factored out the 'clear session' logic into it's own separate function, right (so they're be no redundancy)?

src/logout/GeneralLogoutHandler.ts
pmcb55: Sure, I agree (about having a catch-all `/util` module). I just meant at a high-level, conceptually. I guess I really meant moved to the `sessionInfo` folder, perhaps as a method on the `SessionInfoManager` class in fact...?

src/ClientAuthentication.ts
ajacksified: It occurred to me that logout _also_ emits an event, so maybe it makes more sense to just clear the session.

NSeydoux: This is implemented in a27e30bb1563bab3f5fc78e0fcfc0c7e21d2fbb8. It has a couple of side-effects, but I find it to make the code cleaner: the `LogoutHandler` no longer depends on the local storage, but rather on the `SessionInfoManager`, where the clean up logic is implemented. And initiating login also just calls the same `SessionInfoManager::clear` function.

pmcb55 (approved):
LGTM - yep, definitely an improvement now alright.

PR #300, opened 2020-09-07
Fix lerna publication to NPM

NSeydoux:
Instead of looking into the `package.json` based on the tag, lerna was trying to do something smart, but which doesn't match our pipeline. This flag enforces it to behave the way we expect.


pmcb55 (approved):
LGTM

PR #302, opened 2020-09-08
oidc-client-js is exported by oidc-dpop-client-browser

NSeydoux:
oidc-dpop-client-browser re-exports oidc-client-js so that consuming libraries don't have to manage the two dependencies separately, since the functions we expose will likely be related to the ones in oidc-client-js.


packages/browser/README.md
```diff
@@ -1,97 +1,11 @@
...
-
-The `solid-client-authn` libraries are compatible with [NSS](https://github.com/solid/node-solid-server/releases/tag/v5.3.0) 5.3.X and higher.
-
-## Documentation
-
-- [Using the libraries from within the browser](./docs/browser.md)
-- [Inrupt documentation Homepage](https://docs.inrupt.com/)
+`solid-client-authn-browser` is based on [`oidc-client-js`](https://github.com/IdentityModel/oidc-client-js). 
+However, the latter lacks some features that are necessary to provide the developer
+experience we specifically want for the Solid ecosystem, so we developped
```
Vinnl, line 10:
```suggestion
experience we specifically want for the Solid ecosystem, so we developed  
```

packages/browser/README.md
```diff
@@ -1,97 +1,11 @@
...
-The `solid-client-authn` libraries are compatible with [NSS](https://github.com/solid/node-solid-server/releases/tag/v5.3.0) 5.3.X and higher.
-
-## Documentation
-
-- [Using the libraries from within the browser](./docs/browser.md)
-- [Inrupt documentation Homepage](https://docs.inrupt.com/)
+`solid-client-authn-browser` is based on [`oidc-client-js`](https://github.com/IdentityModel/oidc-client-js). 
+However, the latter lacks some features that are necessary to provide the developer
+experience we specifically want for the Solid ecosystem, so we developped  
+[`oidc-dpop-client-browser`](../oidc-dpop-client-browser/README.md) to add these features.
```
Vinnl, line 11: The README is also visible on npmjs.com, for example, where a relative link won't work.

```suggestion
[`oidc-dpop-client-browser`](https://www.npmjs.com/package/@inrupt/oidc-dpop-client-browser) to add these features.
```

packages/oidc-dpop-client-browser/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@inrupt/oidc-dpop-client-browser",
+  "version": "0.0.1",
+  "description": "A module extending oidc-client-js with new features, such as dynamic client registration and DPoP support.",
+  "homepage": "https://github.com/inrupt/solid-client-authn-js/packages/oidc-dpop-client-browser/",
```
Vinnl, line 5: I think this is what you meant:

```suggestion
  "homepage": "https://github.com/inrupt/solid-client-authn-js/tree/master/packages/oidc-dpop-client-browser/",
```

That said, maybe it's better to just not include the `homepage` property, since there's not much more there to find.

packages/oidc-dpop-client-browser/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "format": "prettier --write \"src/**\" \"__tests__/**\"",
+    "licenses": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
+    "test": "npm run lint && npm run licenses && npm run test-unit",
+    "test-unit": "jest --coverage --verbose --passWithNoTests",
+    "build-api-docs": "npx typedoc --out docs/api/source/api",
+    "build-docs-preview-site": "npm run build-api-docs; cd docs/prose; make html; cd ../api; make html; cd ../; rm -r dist || true; mkdir -p dist/prose dist/api; cp -r prose/build/html/. dist/prose/; cp -r api/build/html/. dist/api/; echo 'Draft documentation: <a href=\"./prose/\">Prose</a>, <a href=\"./api/\">API docs</a>.' >> dist/index.html"
```
Vinnl, line 24: Given that we don't want to be responsible for a generic OIDC library, are we even going to have documentation for this? (Probably not, given that TypeDoc is not even listed as a dependency :) )

packages/oidc-dpop-client-browser/rollup.config.js
```diff
@@ -0,0 +1,32 @@
...
+      format: "cjs",
+    },
+    {
+      file: pkg.module,
+      format: "esm",
+    },
+    {
+      dir: "umd",
+      format: "umd",
+      name: "LitSolidCore",
```
Vinnl, line 18:
```suggestion
      name: "OidcDpopClient",
```

.gitignore
```diff
@@ -86,3 +86,6 @@ packages/**/docs/prose/build/
 packages/**/docs/api/build/
 packages/**/docs/api/source/api/
 packages/**/docs/dist/
+
+# Rollup-built code
+umd/
```
Vinnl, line 91: Shouldn't `packages/**/dist` and `packages/**/umd` also need to be in here?

packages/oidc-dpop-client-browser/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.build.json",
+
+  "compilerOptions": {
+    "lib": ["es2018", "dom"],
+    "outDir": "./dist"
+  },
+
+  "typedocOptions": {
+    "out": "website/docs/api/oidc-dpop-client-browser"
```
Vinnl, line 10: Likewise w.r.t. API docs.

Vinnl (approved):

.gitignore
NSeydoux: As I understand it correctly, `umd/` matches the directories that `packages/**/umd` would match, but it's more generic. 
> frotz/ matches frotz and a/frotz that is a directory
from https://git-scm.com/docs/gitignore.

I don't have a strong opinion on this, but I can change it if that makes it clearer. And `dist/` is already present upper in the file: if I change to `packages/**/umd`, i'll consistently change to `packages/**/dist` as well.

packages/oidc-dpop-client-browser/package.json
NSeydoux: Yeah I wasn't sure if it should be included or not. I ended up including it and having the target file link back to the root of the repository, so that one could get context, but since `oidc-dpop-client-browser` isn't really meant for wide public adoption anyway maybe I'll just remove it.

packages/oidc-dpop-client-browser/package.json
NSeydoux: I wasn't sure about this either. I figured we might still want to build a docs website in case people with a very specific use case end up requiring to contribute to the library. `TypeDoc` is not in the dependencies because it's in the devDependencies of the high-level library. You'll notice that `jest` isn't a dependency either, and yet I was able to run `npm run test` in the submodule folder (not from the top level).

packages/oidc-dpop-client-browser/rollup.config.js
NSeydoux: That's not the first time this happens ^^

.gitignore
Vinnl: Nope you're right, I don't think there will be any `umd` folder anywhere that we _do_ want to include, so just the op-level definition is fine then if that works.

PR #304, opened 2020-09-08
Added codesandbox to PRs

NSeydoux:
This adds a code sandbox to PRs, with a small web app performing login and authenticated fetch.

# Checklist


.codesandbox/sandbox/src/App.js
```diff
@@ -0,0 +1,100 @@
+import React, { useState, useEffect } from "react";
+import {
+  Session as AuthSession,
+  getClientAuthenticationWithDependencies
+} from "@inrupt/solid-client-authn-browser";
+
+// const REDIRECT_URL = "https://podpecker.zwifi.eu"
+// const REDIRECT_URL = "http://localhost:3000";
+const REDIRECT_URL = "https://ycos2.csb.app/";
```
Vinnl, line 9: I don't think this will work, given that a new sandbox is created for every PR. Possibly just `window.location`?

.codesandbox/ci.json
```diff
@@ -0,0 +1,3 @@
+{
+  "sandboxes": ["/.codesandbox/sandbox"]
```
Vinnl, line 2: If the examples in the browser package still exist, possibly you could just reuse those. But a separate demonstration works as well of course.

Vinnl (approved):
LGTM. Don't forget to set up [the GitHub app](https://github.com/apps/codesandbox) if you haven't already (see [docs](https://codesandbox.io/docs/ci)).

.codesandbox/ci.json
NSeydoux: The examples still exist, but I actually wanted to replace that code with the one featured in this PR eventually: the React hooks make the code much cleaner, both for `useEffect` and `useState`. 

Also for some reason that I didn't take the time to investigate yet, the parcel bundler complains about missing dependencies that don't prevent webpack from building the examples, so I'm not sure the exact same code would run in both contexts.

.codesandbox/sandbox/src/App.js
NSeydoux: Good idea, `window.location` does the trick. Implemented in 812e598c79760c20176484d92b3ee1eb219c788f.

PR #314, opened 2020-09-14
Align TypeDoc setups

Vinnl:
Since API documentation is generated per-package, they each need to
specify e.g. not to include non-exported files.

(Although TypeDoc allows you to specify its options in your
tsconfig, it does not appear to also follow its `extends`
statements in the way TypeScript does.)

Note: Creating as a draft so Vercel can deploy these API docs, so I can compare them to the current API docs.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -16,7 +16,7 @@
     "moduleResolution": "node",
     "outDir": "./dist"
   },
-  "exclude": ["**/node_modules", "**/dist"],
+  "exclude": ["node_modules", "dist", "**/__mocks__/**"],
 
   // We don't provide an 'out' value here, each sub-package should provide its
   // own.


pmcb55 (commented):
Yikes - that's a bit shitty that TSDoc doesn't support 'extends'! I guess it means we need to duplicate these config options in the `oidc-dpop-client-browser` sub-package too...?

I'm seeing loads of `Missing return type on function` warnings too (maybe they're already fixed though).

Vinnl:
@pmcb55 We haven't set anything up for TypeDoc in oidc-dpop-client-browser yet, but when we do, yes, it will need its own setup.

That said, I'm not sure if this is actually the error. Kay reported things getting better, but now that I can see the generated API docs with the Vercel deployment, there's still a lot in there that shouldn't be there, and I've yet to figure out why.

(All "missing return type" warnings should go away after #316 is merged into this PR.)

Vinnl:
@pmcb55 Good news: turns out TypeDoc _does_ support `extends`. It's just that the mocks directory got inadvertently removed from the list. So good to go now.

pmcb55 (approved):
LGTM - Nic removed the mocks folder here when we were trying to resolve the inter-package TSyringe issue, with the intent that dependent packages might want to re-use mocks from lower-level packages. We probably need to look at that notion (it's something I've done before in Java - i.e. having specific released artifacts for use purely by tests (e.g. 'test-support' JARs)), but I've no idea if the same notion is common in JavaScript-land.
Anyway, if everything still works fine when excluding the mocks, then we can look at the 'sharing mocks' idea later (if needed).

PR #315, opened 2020-09-14
Use a single place to exclude files from npm

Vinnl:
We already have a `.npmignore` file that specifies which files
should not be published to npm. The `files` key in `package.json`
specified which files _should_ be published there, i.e. files not
listed there would not be published.

This resulted in source files not being published, and source maps
failing. Since the source files can help in debugging (both for
developeres, and when we receive bug reports), we should publish
those as well.


pmcb55 (approved):
LGTM

PR #316, opened 2020-09-14
Annotate return types where ESLint wants us to

Vinnl:
There are a number of annotations listed in our CI: https://github.com/inrupt/solid-client-authn-js/actions/runs/253759683

They all come down to the same lint warning not being respected.

image

This PR fixes those lint warnings by making return types explicit.

### DIFF
--- a/packages/core/src/storage/__mocks__/StorageUtility.ts
+++ b/packages/core/src/storage/__mocks__/StorageUtility.ts
@@ -113,7 +113,7 @@ export const mockStorageUtility = (
       userId: string,
       key: string,
       options?: { errorIfNull?: boolean; secure?: boolean }
-    ) => {
+    ): Promise<string | undefined> => {
       const store = options?.secure ? secureStore : nonSecureStore;
       return store[userId]
         ? (store[userId] as Record<string, string>)[key]
@@ -123,7 +123,7 @@ export const mockStorageUtility = (
       userId: string,
       values: Record<string, string>,
       options?: { secure?: boolean }
-    ) => {
+    ): Promise<void> => {
       const store = options?.secure ? secureStore : nonSecureStore;
       store[userId] = values;
     },
@@ -131,14 +131,14 @@ export const mockStorageUtility = (
       userId: string,
       key: string,
       options?: { secure?: boolean }
-    ) => {
+    ): Promise<void> => {
       const store = options?.secure ? secureStore : nonSecureStore;
       delete (store[userId] as Record<string, string>)[key];
     },
     deleteAllUserData: async (
       userId: string,
       options?: { secure?: boolean }
-    ) => {
+    ): Promise<void> => {
       const store = options?.secure ? secureStore : nonSecureStore;
       delete store[userId];
     },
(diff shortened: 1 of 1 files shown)


pmcb55 (approved):
LGTM

NSeydoux:
Do you know why GitHub complained about this, but ESlint wasn't failing the CI ?

Vinnl:
@NSeydoux ESLint rules can be set to `off` (do not check for the rule), `warn` (check for the rule, but only log a warning) and `error` (actually return a non-zero exit code in addition to logging the error, failing the build). Apparently this rule is set to just be a warning. You could e.g. run ESLint with `--max-warnings=0` in CI if you want them to halt CI as well, without blocking during development.

pmcb55:
@Vinnl - so to prevent this popping up again later, should we suggest setting the rule to be `error` instead of the current `warning` (I'm assuming this rule, and it's current `warning` setting, are coming from our 'Inrupt-wide standard ESLint ruleset' guidelines)?
I don't really see much benefit in 'not blocking during development' to only get tripped up later at CI time, i.e. it's such a simple rule to comply with, and the addition of the rule itself makes sense, so why not just force developers to adhere to the rule as soon as possible...?

Vinnl:
@pmcb55 I think the rule comes from ESLint's own recommended ruleset, but I didn't check. That said, as mentioned I think it probably makes sense to set _all_ warnings to errors.

With "not blocking during development" I meant e.g. not failing live reload on it (which you would do for errors which point to potential bugs, like uninitialised variables). It would make sense to also do so in the pre-commit hook, for example, to make developers adhere as soon as possible, but not any sooner than that it is sure the code is actually going to be included in the codebase.

pmcb55:
@Vinnl - yep, I agree that it "makes sense to set all warnings to errors" (generally anyway), so more than happy for us to try that out and see how we get on. (I'm not sure how to do that, so is that something you can do easily (or just create a backlog ticket if you think it needs one)?)

On the live reload question - yeah, I guess the pre-commit hook is fine. I'm just a bit uncomfortable with the idea of '_It's Ok to be sloppy until you're sure_'. In this case, isn't it just a few keystrokes to be 'correct' as your writing the code in the first place (and in an IDE won't it complain as you type anyway)...? In other words, is there really much of an advantage to slightly delaying enforcement of, presumably, just this one ESLint rule (or would this 'breaking live reload' apply to lots of annoying little linty rules (like missing TSDoc, or stuff like that))...?

Vinnl:
As I said, just add the `--max-warnings=0` argument to where you're calling ESLint (e.g. for its entry for `lint-staged` in `package.json` to set the pre-commit hook, and in `ci.yml` for CI).

It's not about the number of keystrokes, it's about having to do context switching: when you're prototyping something and thinking about potential implementations, you don't want to be spending mental cycles on figuring out what ESLint rule you're breaking and how to fix it, when that rule will have no effect on your learning how best to implement something. (Note that your editor is still able to show you warnings in the gutter - it just allows you to ignore it until you commit it.)

(And again, there's no reason to limit it to just this one ESLint rule, other than that we just happened to run into that one now.)

PR #318, opened 2020-09-14
Make Dependabot aware of the current repo layout

Vinnl:
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #320, opened 2020-09-15
Hoist up devDependencies

Vinnl:
# New feature description

Hoisting up devDependencies to the top-level package.

This has several advantages:
- All packages use the same versions. (We will usually update them
  together anyway - if an upgrade causes problems in a single
  package, we can always re-add it as a separate dependency for
  that specific package at that point in time.)
- There is less duplication and thus less disk space usage, and
  installation takes less long.

This commit's contents was based on `npx lerna link convert`.

See https://github.com/lerna/lerna#common-devdependencies.

# Checklist

- [ ] All acceptance criteria are met. N/A


package.json
```diff
@@ -21,29 +21,32 @@
...
   },
   "dependencies": {
+    "@inrupt/oidc-dpop-client-browser": "file:packages/oidc-dpop-client-browser",
+    "@inrupt/solid-client-authn-browser": "file:packages/browser",
+    "@inrupt/solid-client-authn-core": "file:packages/core",
     "lerna": "^3.22.1"
   },
   "devDependencies": {
-    "@types/jest": "^26.0.13",
+    "@types/jest": "^26.0.12",
```
Vinnl, line 30: Note that some of these versions are lower, presumably because some of the child packages used to have a lower version. Dependabot should take care of that for future upgrades (also see #318).

packages/browser/package.json
```diff
@@ -95,5 +71,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "jest": "^26.4.2"
```
Vinnl, line 76: Dependencies that ship their own binaries, like Jest, will still need to be available in the relevant packages so they can be run from there.

packages/browser/package.json
```diff
@@ -45,30 +45,8 @@
...
-    "typedoc-plugin-markdown": "^2.4.2",
-    "typescript": "^3.9.7",
-    "webpack": "^4.44.1",
-    "webpack-bundle-analyzer": "^3.8.0",
-    "webpack-cli": "^3.3.12",
-    "webpack-merge": "^4.2.2"
-  },
   "dependencies": {
-    "@inrupt/solid-client-authn-core": "^0.1.4",
+    "@inrupt/solid-client-authn-core": "file:../core",
```
pmcb55, line 49: Yikes - this doesn't look right. Lerna should be using symbolic links, so I don't think local `file:` references shouldn't ever appear here (my local copy is `"@inrupt/solid-client-authn-core": "^0.1.4",`)...?

Vinnl: This is what `lerna link convert` generated, so I assumed that that was how it should be. Looking at [the published package](https://unpkg.com/browse/@inrupt/solid-client-authn-browser@0.1.5-chorededupe-devDependencies-257252491-828.0/package.json) it also seems like it gets set to the correct dependency on publication.

I can't find anything about whether Lerna is supposed to use symbolic links or something like this, so I don't know whether we should revert its own output. Do you know if there's docs anywhere on what this should look like?

Vinnl: @pmcb55 I've reverted the `file:` references, and child packages now only have devDependencies with binaries in them. (Except that `-browser` is the only package using Webpack, so all Webpack-related dependencies are in there as well.)

Vinnl: Never mind; tests are still failing, and this doesn't actually seem to hoist up that much anymore, so given that I shouldn't spend too much time on this, I'm closing this.

PR #338, opened 2020-09-22
Merge all succeeding Dependabot PRs

Vinnl:
After #318, Dependabot opened a whole bunch of PRs updating outdated PRs. Because merging them in order takes a long time (after every merge, the other PRs have to be updated again and have their pipelines run), this PR bundles all of those PRs whose pipelines succeeded. If the pipeline of this one still succeeds, we should be relatively safe in merging them all at once.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10825,9 +10825,9 @@
       }
     },
     "rollup": {
-      "version": "2.26.11",
-      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.26.11.tgz",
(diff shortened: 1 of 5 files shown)


pmcb55 (approved):
LGTM

NSeydoux:
What I usually do in similar situation is to comment `@dependabot merge` on the PRs. Dependabot automatically rebases the PRs, so it's not required to manually update PRs after each merge, and in the case where a previous merge fails the CI on a PR, dependabot  won't merge it. Are there additional reasons to do this work manually, besides the fact that letting dependabot do its thing can last for quite a while ?

Vinnl:
@NSeydoux Aha, I was wondering why you always left them as comments :) No, the reason is that I did not know that Dependabot would automatically rebase and then merge them in a row. That's good to know.

PR #356, opened 2020-10-01
Keep the access token in the fetch closure

NSeydoux:
This prevents the access token to be stored in-memory, and instead stores it in the closure of the session's `fetch` function. Currently, only bearer tokens are supported, but the same mechanism may be extended for DPoP support in a future PR.

# Checklist


packages/browser/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @returns A fetch function that adds an Authorization header with the provided
+ * bearer token.
+ * @hidden
+ */
+export function buildBearerFetch(authToken: string): typeof fetch {
+  return (init, options): Promise<Response> => {
+    return fetch(init, {
+      ...options,
+      headers: {
+        Authorization: `Bearer ${authToken}`,
```
Vinnl, line 38:
```suggestion
        ...options.headers,
        Authorization: `Bearer ${authToken}`,
```

or maybe (not sure if this is necessary):

```suggestion
        ...(options.headers ?? {}),
        Authorization: `Bearer ${authToken}`,
```

Otherwise you're overwriting all headers that the developer passes. (Move the `Authorization` header one line up if we want the developer to be able to override that, but I suppose we don't.)

(And maybe add a test that verifies this behaviour.)

packages/browser/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export async function buildDpopFetch(
+  authToken: string,
+  dpopKey: JSONWebKey
+): Promise<typeof fetch> {
+  return async (init, options): Promise<Response> => {
+    return fetch(init, {
+      ...options,
+      headers: {
+        Authorization: `DPoP ${authToken}`,
```
Vinnl, line 58: Likewise:

```suggestion
        ...options.headers,
        Authorization: `DPoP ${authToken}`,
```

packages/browser/src/authenticatedFetch/fetchFactory.ts
Vinnl: Although come to think of it, `headers` needn't be a plain object, it can also be a `Headers` instance. So maybe to be sure:

```javascript
const headers = new Headers(options.headers);
headers.set('Authorization', `Bearer ${authToken}`);
```

and then use that for `headers`.

packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import "reflect-metadata";
```
Vinnl, line 22: Given that there's no TSyringe in the fetch factory, I don't think this is necessary.

```suggestion
```

packages/browser/src/ClientAuthentication.ts
```diff
@@ -125,6 +112,10 @@ export default class ClientAuthentication {
   handleIncomingRedirect = async (
     url: string
   ): Promise<ISessionInfo | undefined> => {
-    return this.redirectHandler.handle(url);
+    const redirectInfo = await this.redirectHandler.handle(url);
+    // TODO: When handling DPoP, both the key and the token should be returned
+    // by the redirect handler.
+    this.fetch = buildBearerFetch(redirectInfo.accessToken);
+    return { ...redirectInfo } as ISessionInfo;
```
Vinnl, line 119: Is this not the same, or do you explicitly want a clone? And is the type assertion really necessary?

```suggestion
    return redirectInfo as ISessionInfo;
```

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -113,19 +110,10 @@ describe("ClientAuthentication", () => {
   describe("fetch", () => {
     it("calls fetch", async () => {
       const clientAuthn = getClientAuthentication();
-      const response = await clientAuthn.fetch(
-        "mySession",
+      await clientAuthn.fetch("https://zombo.com");
```
Vinnl, line 113: Unfortunately the old site uses Flash :P

```suggestion
      await clientAuthn.fetch("https://html5zombo.com");
```

packages/browser/src/Session.ts
```diff
@@ -134,7 +134,7 @@ export class Session extends EventEmitter {
    * @param init Optional parameters customizing the request, by specifying an HTTP method, headers, a body, etc. Follows the [WHATWG Fetch Standard](https://fetch.spec.whatwg.org/).
    */
   fetch = async (url: RequestInfo, init?: RequestInit): Promise<Response> => {
-    return this.clientAuthentication.fetch(this.info.sessionId, url, init);
+    return this.clientAuthentication.fetch(url, init);
```
Vinnl, line 137: Huh. How was this not a major bug already?

Vinnl (approved):
I'm not sure I followed everything that was going on - I might need a more detailed description of what's going on in the future :P As I understood it now, the redirect handlers no longer store the tokens received after redirection in the localstorage, but simply return it, and then those return values are passed to the fetch factory to create a fetch function that encloses the token. I'm just not really sure about what the changes to TokenSaver do.

Anyway, left some comments on the things that I did see. Mostly looks good :)

packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
NSeydoux: I actually had to add this because TSyringe complained: since `fetchFactory` imports files where TSyringe annotations are used, and the test file is an entry point to the application, the `import "reflect-metadata"` statement is required.

packages/browser/src/ClientAuthentication.ts
NSeydoux: That was actually laziness from my part: I did not want the `accessToken` to be returned visibly, hence the type assertion. But now that I think about it cloning the `redirectInfo` object would not prevent the property to be present in the instance, even if it is hidden in the type. I fixed this properly in cf94fc1e3c6ca8ad2aaca8b92b57a840da9cc224.

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: Oh my, I didn't realize that was an actual website. And what a website !

packages/browser/src/Session.ts
NSeydoux: The `clientAuthentication.fetch` used to take three parameters, and to use the sessionId to look up the access token, so it was not expected to match the standard `fetch` signature. Is that what's your raising here ?

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: Resolved in 01046978d8b6d895ed113c543d8dfc8d940c606f

packages/browser/src/Session.ts
Vinnl: Ah no I misread, I thought you replaced `this.info.sessionId` with `url`, i.e. that it used to just send a fetch to the session ID. Never mind!

packages/browser/src/authenticatedFetch/fetchFactory.ts
NSeydoux: Excellent point. Fixed in 643b73a3e943e8c3768b714b8b35965117173c42.

packages/browser/src/authenticatedFetch/fetchFactory.ts
NSeydoux: Resolved in 643b73a3e943e8c3768b714b8b35965117173c42 as well.

packages/browser/src/login/oidc/redirectHandler/AggregateRedirectHandler.ts
```diff
@@ -39,7 +39,7 @@ import {
  */
 @injectable()
 export default class AggregateRedirectHandler
-  extends AggregateHandler<[string], ISessionInfo | undefined>
+  extends AggregateHandler<[string], ISessionInfo & { accessToken: string }>
```
pmcb55, line 42: Is this an interim step, or is it intended that the `accessToken` always be passed about like this? (I was thinking instead we'd be able to push the ever-so-precious accessToken directly into the fetch closure **_immediately_** upon getting it from `oidc-client-js`, and never need to pass it around anywhere (not that I've thought through if that's feasible or not :) !?)

NSeydoux: We could push the call to `buildBearerFetch` one level down and return the `fetch` instead of the `accessToken`, but as it is right now the token is ever only accessible in the `handleRedirect` scope, from which it is stored in the `fetch` scope, so I don't think moving the call would make a difference. I find it slightly clearer that the fetch is built by the object that will then expose it, what do you think ?

pmcb55: Yeah, is certainly debatable. My intuition says move it down one level - I'm just a tad uncomfortable seeing exposure of the 'concept' of an `accessToken` across all the redirect handlers. I'm not saying that's wrong, but it just feels a tiny bit safer if redirect handlers all return fetch functions instead of super-secret accessTokens (not that them returning fetch functions is super-intuitive either of course). So unless you've any further misgivings about doing it, I'd move it down.

PR #374, opened 2020-10-02
Removed Usage and examples from README

nicolasmondada:
Removed Usage and examples from README


pmcb55 (approved):
LGTM (this is fine, but we should probably create a ticket to add simple docs on how to run the new Test Demo App). Also, you should edit the PR template before creating the PR itself!

nicolasmondada:
> LGTM (this is fine, but we should probably create a ticket to add simple docs on how to run the new Test Demo App). Also, you should edit the PR template before creating the PR itself!
https://inrupt.atlassian.net/browse/SDK-1522

PR #375, opened 2020-10-02
Fix/authorization headers

NSeydoux:
This PR fixes a bug where the Authorization headers are added correctly  in the test environment, but not in the production one. Due to the nature of the bug, it's hard to test for non-regression. Not being quite sure of what's going on, I created a dedicated branch (https://github.com/inrupt/solid-client-authn-js/tree/debug/wtf-headers) so that we can take a closer look at the issue without blocking this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### New features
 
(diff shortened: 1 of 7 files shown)


pmcb55 (approved):
LGTM - looks a good bit cleaner, although I don't know the 'difference' other than that (i.e. I don't see where the bug was either!).

PR #388, opened 2020-10-06
Failed registration response processing

NSeydoux:
This commit introduces cleaner error management when client registration fails.

# Checklist

### DIFF
--- a/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
@@ -150,7 +150,7 @@ describe("ClientRegistrar", () => {
     it("handles a failure to dynamically register elegantly", async () => {
       const mockFetch = jest.fn().mockResolvedValueOnce(
         /* eslint-disable @typescript-eslint/camelcase */
-        (new NodeResponse("bad stuff that's an error", {
+        (new NodeResponse('{"error":"bad stuff that\'s an error"}', {
           status: 400,
         }) as unknown) as Response
         /* eslint-enable @typescript-eslint/camelcase */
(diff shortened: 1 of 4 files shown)


jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   roots: ["<rootDir>"],
   clearMocks: true,
-  testMatch: ["**/__tests__/**/?(*.)+(spec|test).+(ts|tsx|js)"],
+  testMatch: ["**/(__tests__|src)/**/?(*.)+(spec|test).+(ts|tsx|js)"],
```
pmcb55, line 4: Does this mean we now have a mish-mash of tests in both the `___tests__` directory and the `src` directory...?

packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -48,33 +48,20 @@ describe("ClientRegistrar", () => {
...
         (new NodeResponse(
           JSON.stringify({
             client_id: "abcd",
             client_secret: "1234",
+            redirect_uris: ["https://example.com"],
           })
         ) as unknown) as Response
         /* eslint-enable @typescript-eslint/camelcase */
       );
+      const savedFetch = global.fetch;
```
pmcb55, line 63: This looks a tad scary (i.e. if this tests fails after doing this). Should this be a `try { } finally { /* Restore global fetch...*/ }` ...?

pmcb55 (approved):
LGTM (I tried to comment in a few places about delimiting vars in error messages (using '[' and ']'), but it complains about the diff being out-of-date).

jest.config.js
NSeydoux: That's correct: since one of the test file did not exist, I added it along the source code rather than in the `__tests__` folder.

packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
NSeydoux: Actually, I realized that we'd never want `global.fetch` to be called in our tests before being mocked first, so in 1bd4e8d120e1fa5d1b67904fcc11a92648f39c2d I cleaned up all the saving and restoring of `global.fetch`. It should only ever be replaced by the mock which is relevant in the current test.

PR #394, opened 2020-10-07
Feat/export types

pmcb55:
We might not merge this PR ever - it was really just a quick learning exercise while testing that our TS types could be exposed and used easily by external code.
It **_might_** be handy for automated testing, but it seems awkward to maintain the explicit interface references, and I suspect our other tests (integration and end-2-end in particular) make more sense for testing that more comprehensively (although this example is so simple and self-contained, it might still have some value...?!).


pmcb55:
Mistakenly created against `master` (instead of `1.x`!)

PR #395, opened 2020-10-07
Feat/export types

pmcb55:
We might not merge this PR ever - it was really just a quick learning exercise while testing that our TS types could be exposed and used easily by external code.
It **_might_** be handy for automated testing, but it seems awkward to maintain the explicit interface references, and I suspect our other tests (integration and end-2-end in particular) make more sense for testing that more comprehensively (although this example is so simple and self-contained, it might still have some value...?!).

### DIFF
--- a/packages/example/.eslintrc.js
+++ b/packages/example/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  extends: ["../../.eslintrc.js"],
+  // rules: {
+  //   "license-header/header": ["error", "../../../resources/license-header.js"],
+  // },
+};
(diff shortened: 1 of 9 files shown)


NSeydoux:
Should this PR be merged or closed ? At this point, I'm afraid the overhead of having an additional package be deployed on NPM (or the configuration to prevent said publication), plus maintaining the package over time, outwheights the benefits.

pmcb55:
Yeah, I agree - this example is really minimal, and so it's value doesn't outweigh the cognitive noise involved in understanding what it's for. Closing this now.

PR #396, opened 2020-10-07
Moved key generation to oidc-client-dpop-browser

NSeydoux:
The DPoP key is now generated in the oidc-client-dpop-browser module, and the solid-client-authn-browser modules only calls a high-level function to get a new key.
The key is currently still stored in memory, which we want to change eventually. Also, the crypto libraries relied upon to generate the key require some node builtins, which are provided by rollup plugins for now.

### DIFF
--- a/packages/browser/__tests__/jose/IsomorphicJoseUtility.test.ts
+++ b/packages/browser/__tests__/jose/IsomorphicJoseUtility.test.ts
@@ -23,25 +23,16 @@ import { describe, it } from "@jest/globals";
 import {
   decodeJWT,
   generateCodeVerifier,
-  generateJWK,
   signJWT,
 } from "../../src/jose/IsomorphicJoseUtility";
-
-describe("generateJWK", () => {
-  it("can generate a RSA-based JWK", async () => {
-    const key = await generateJWK("RSA");
-    expect(key.kty).toEqual("RSA");
-  });
-
-  it("can generate an elliptic curve-based JWK", async () => {
-    const key = await generateJWK("EC", "P-256");
-    expect(key.kty).toEqual("EC");
-  });
-});
+import {
+  generateDpopKey,
+  generateRsaKey,
+} from "@inrupt/oidc-dpop-client-browser";
 
 describe("signJWT/decodeJWT", () => {
   it("generates a JWT that can be decoded without signature verification", async () => {
-    const key = await generateJWK("RSA");
+    const key = await generateRsaKey();
     const payload = { testClaim: "testValue" };
     const jwt = await signJWT(payload, key, { algorithm: "RS256" });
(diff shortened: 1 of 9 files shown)


packages/browser/__tests__/jose/IsomorphicJoseUtility.test.ts
```diff
@@ -23,41 +23,32 @@ import { describe, it } from "@jest/globals";
...
-    const key = await generateJWK("EC", "P-256");
-    expect(key.kty).toEqual("EC");
-  });
-});
+import {
+  generateDpopKey,
+  generateRsaKey,
+} from "@inrupt/oidc-dpop-client-browser";
 
 describe("signJWT/decodeJWT", () => {
```
pmcb55, line 33: The unit tests in this `describe()` block look like they belong down in the DPop library (or just removed from here if there are similar/same tests down in that library already). (But then, is the RSA key generation really DPop (or DCR)-specific (i.e. does it belong in `oidc-client-dpop-browser` at all...)?

packages/oidc-dpop-client-browser/src/dpop/dpop.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("can generate an elliptic curve-based JWK", async () => {
+    const key = await generateJWK("EC", "P-256");
+    expect(key.kty).toEqual("EC");
+  });
+});
+
+describe("generateDpopKey", () => {
+  it("generates an elliptic curve-base key", async () => {
```
pmcb55, line 39: This seems to be implying that a DPoP key simply **_is_** an elliptic curve-based key, but that doesn't seem right. Question then is, so what is a `DPoP key` exactly (isn't it just a JWK that happens to be generated with certain params)...?

pmcb55 (approved):

packages/browser/__tests__/jose/IsomorphicJoseUtility.test.ts
NSeydoux: Signing and decoding the JWT will be moved to the `oidc-client-dpop-browser` module in a future PR, the current one only moves key generation around. The RSA key generation isn't DPoP/DCR-specific, but moving it in the oidc-specific module isolates the crypto code from the rest of the library.

packages/oidc-dpop-client-browser/src/dpop/dpop.spec.ts
NSeydoux: Fixed in 28296d3cc4a69bfaf52e4daa6d2f493bd21211b6, and the method name has been changed in cb2893aa360c62d8cb24e2da88f1e033558eb372

PR #401, opened 2020-10-08
Pass refresh_token to fetch builders

pmcb55:
Just the initial part of adding refresh token support (we need the fetch closure code to be extended first before we can complete this, since now we are only returning an augmented fetch function (so we have nowhere to store the refresh token, or any other sensitive info we may wish to associated with the fetch)).


NSeydoux (approved):

pmcb55:
Re-applied these changes in the new PR: https://github.com/inrupt/solid-client-authn-js/pull/446 (as we forgot to merge this PR back before major changes for DPoP)

PR #402, opened 2020-10-08
Feat/dpop/sign header

NSeydoux:
This PR moves the code required to:
- generically sign a JWT 
- specifically generate a DPoP header
from the `solid-client-authn-browser` module to the `oidc-client-dpop-browser` one. 

The dependency on crypto libraries used to sign the JWT that is used as a DPoP header has been pushed down one level.
However, this created some issues with rollup. For the time being, since the library is not exposed to be used directly in the browser, the bundler isn't used anymore, and the library is just transpiled before being consumed by the browser module.

Note that this PR makes a lot of the fetcher/dpop code in the codebase obsolete, it will be removed in a different PR (for ease of review).


packages/oidc-dpop-client-browser/package.json
```diff
@@ -13,7 +13,8 @@
     "clean": "npm run clean-module && npm run clean-browser && npm run clean-examples && rm -rf ./coverage",
     "clean-module": "rm -rf ./dist",
     "clean-browser": "rm -rf ./browserDist",
-    "build": "rollup --config rollup.config.js",
+    "rollup": "rollup --config rollup.config.js",
```
Vinnl, line 16: Maybe we can just remove rollup altogether?

packages/browser/__tests__/login/oidc/redirectHandler/GeneralRedirectHandler.spec.ts
```diff
@@ -28,8 +28,9 @@ import {
 import GeneralRedirectHandler from "../../../../src/login/oidc/redirectHandler/GeneralRedirectHandler";
 import {
   generateJWK,
-  signJWT,
+  // signJWT,
```
pmcb55, line 31: Just remove.

packages/oidc-dpop-client-browser/package.json
```diff
@@ -13,7 +13,8 @@
...
+    "build": "tsc -p tsconfig.json",
```
pmcb55, line 17: Just a niggle, but I prefer full switches, so here it'd be "tsc --project tsconfig.json".

NSeydoux:
Moving this to draft, as the bundler issue breaks the browser deployment.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
  * @hidden
  */
 export async function generateJWK(
```
pmcb55, line 45: Another niggle, but our naming convention means this should be named `generateJwk()`.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+/**
+ * Generates a Json Web Token (https://tools.ietf.org/html/rfc7519) containing
+ * the provided payload and using the signature algorithm specified in the options.
+ * @param payload The body of the JWT.
+ * @param key The key used for the signature.
+ * @param options
+ * @returns a 3-parts base64-encoded string, split by dots.
+ * @hidden
+ */
+export async function signJWT(
```
pmcb55, line 63: All named with `JWT` should be `Jwt`, etc.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+  return (await JWK.asKey(key as JWK.RawKey, "public")) as JSONWebKey;
+}
+
+/**
+ * Normalizes a URL in order to generate the DPoP token based on a consistent scheme.
+ * @param audience The URL to normalize.
+ * @returns The normalized URL as a string.
+ * @hidden
+ */
+export function normalizeHtu(audience: URL): string {
```
pmcb55, line 116: Can we rename this to English, e.g. `normalizeHttpUriClaim()` (I don't know if `htu` is technically a 'claim')? Unfortunately this function isn't inside a Class to help give context, but is it really DPoP-specific?

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+ * @param key The key used for the signature.
+ * @param options
+ * @returns a 3-parts base64-encoded string, split by dots.
+ * @hidden
+ */
+export async function signJWT(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  payload: Record<string, any>,
+  key: JWKECKey | JWKOKPKey | JWKRSAKey | JWKOctKey,
+  options?: JoseJWT.SignOptions
```
Vinnl, line 67:
```suggestion
  options: JoseJWT.SignOptions = {}
```

Would this allow you to remove the type assertion below? I can't see what the type would've been otherwise in the PR review interface, but it's making me nervous...

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+): Promise<string> {
+  const parsedKey = await JWK.asKey(key);
+  const convertedKey: string = parsedKey.toPEM(true);
+  const signed = JWT.sign(payload, convertedKey, {
+    ...(options as JWT.SignOptions),
+  });
```
Vinnl, line 73: Also, even if you need to keep the assertion, is there a need to clone it?

```suggestion
  const signed = JWT.sign(payload, convertedKey, (options as JWT.SignOptions));
```

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+
+/**
+ * Creates a DPoP header according to https://tools.ietf.org/html/draft-fett-oauth-dpop-04,
+ * based on the target URL and method, using the provided key.
+ * @param audience Target URL.
+ * @param method HTTP method allowed.
+ * @param key Key used to sign the token.
+ * @returns A JWT that can be used as a DPoP Authorization header.
+ */
+export async function createHeaderToken(
```
pmcb55, line 128: Again, without a Class to provide context here, is this function better named as `createDpopHeaderToken()` (given the `typ` value here, it does seem to be very DPoP-specific)?
Raises another point on the function naming across all this code - we seem to be using `function xxxxJkt()` in places, and `function xxxToken()` in others - should we just be consistent everywhere (I'd suggest we use `Jkt` instead of `Token`, as it's more specific and explicit)...?

pmcb55: ...and likewise, I think we use `function xxxKey()` in places and `functionJwk()` in others (again I'd favor the more explicit `Jwk` instead of `Key`).

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+  }
+  return JWT.decode(token) as Promise<Record<string, unknown>>;
+}
+
+/**
+ * @param key
+ * @hidden
+ */
+export async function privateJWKToPublicJWK(
+  key: JSONWebKey
```
Vinnl, line 105: Reading the assertion below, should this be the following?

```suggestion
  key: JWK.RawKey
```

pmcb55 (approved):
LGTM (just minor comments on naming)

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
```diff
@@ -30,14 +43,111 @@ import { JWK } from "node-jose";
...
+}
+
+/**
+ * Normalizes a URL in order to generate the DPoP token based on a consistent scheme.
+ * @param audience The URL to normalize.
+ * @returns The normalized URL as a string.
+ * @hidden
+ */
+export function normalizeHtu(audience: URL): string {
+  return `${audience.origin}${audience.pathname}`;
```
Vinnl, line 117: I realise this is existing code (probably holds for the rest as well), but shouldn't the DPoP token take into account everything but the hash? Most worryingly, this appears to ignore query params, even though that can result in completely different resources (e.g. `?ext=acl`)? (I guess not, since the tests explicitly verify for it, but just to make sure.)

An alternative may be:

```suggestion
  const cleanedAudience = new URL(audience.href);
  cleanedAudience.hash = "";
  return cleanedAudience.href;
```

packages/oidc-dpop-client-browser/src/dpop/dpop.spec.ts
```diff
@@ -48,3 +57,94 @@ describe("generateRsaKey", () => {
...
+    const key = await generateKeyForDpop();
+    const payload = { testClaim: "testValue" };
+    const jwt = await signJWT(payload, key, {
+      algorithm: "RS256",
+    });
+    const decoded = await decodeJWT(jwt);
+    expect(decoded.testClaim).toEqual(payload.testClaim);
+  });
+
+  it("can verify the ES256 signature of the generated JWT", async () => {
```
Vinnl, line 72: What is it supposed to do when the verification fails? Is it doable to have a test for that?

packages/browser/__tests__/login/oidc/redirectHandler/GeneralRedirectHandler.spec.ts
Vinnl, line 31:
```suggestion
```

Vinnl (approved):

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
pmcb55: Yeah, that's a great call @Vinnl - very important to ensure query params are included, so would be good to properly verify (i.e. that the tests aren't just 'mocking away' the query param consideration).

packages/browser/__tests__/login/oidc/redirectHandler/GeneralRedirectHandler.spec.ts
NSeydoux: Resolved in 6176710526f4f13537f4419ec557ba6bd2a9e849

packages/oidc-dpop-client-browser/package.json
NSeydoux: Since Vincent figured out an issue with the rollup config, the pure `tsc` script is no longer required. Resolved in 04c8dfbdcd27bfe611a839d92656fd2f8d915fd5.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: Fixed in 86755fc7ab85dc1b8cbbe6c6ea5c0d1d4c57e3b6

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: Fixed in 86755fc7ab85dc1b8cbbe6c6ea5c0d1d4c57e3b6

packages/oidc-dpop-client-browser/package.json
NSeydoux: I removed tsc, and kept rollup, since you found the config issue.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: Good catch, that would absolutely have been an issue. I fixed this in 9c1a7c8bb2feb95bc4d9766b652d257c86112307. You'll see that I still remove any username/password present in the URL, I think it's the right thing to do, do you agree ?

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: Changed in eb02277a9b53f21ce24688d549a5908ec816b2ac. I don't really think there is a need to specify that this is DPoP specific, as the function is only ever used in the `dpop.ts` file, and it is exported only for test purpose.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
pmcb55: Yeah, I guess so - but probably a good idea to run it past Aaron too.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: I refactored the function name to `createDpopHeader` in f41a097f64666f7e014dc9b118009514961b9bf9, since it is indeed very DPoP-specific.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
pmcb55: Yeah, my point was just that if it's **_not_** DPoP-specific then should it be moved to a Utils class, or a Claims class, or something. But it's perfectly fine here - I'd only consider moving it (perhaps into the Core package) if something **_else_** wanted to call this function too.

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: And the change from `key` to `jwk` is in d411c4359f6f201b3c216821106daf542fb507e4

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: No, actually the issue was an inconsistency in types expected by this function and in an interface that uses it. I had not paid attention to it so far, but it is now fixed: 0d4acef4280dc4a09297631c98b42422fe10d335

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: Indeed 0d4acef4280dc4a09297631c98b42422fe10d335

packages/oidc-dpop-client-browser/src/dpop/dpop.ts
NSeydoux: The assertion was actually unecessary: fd42481886940ecf40fc12e793312769f7ae2746

packages/browser/__tests__/login/oidc/redirectHandler/GeneralRedirectHandler.spec.ts
NSeydoux: Removed in 6176710526f4f13537f4419ec557ba6bd2a9e849

packages/oidc-dpop-client-browser/src/dpop/dpop.spec.ts
NSeydoux: If the verification fails, it throws. I added a test there : d855d58d7e5e8871e800597275e6a77844d6ea3f. That's the behaviour I'd expect for a low-level library like this, but one layer higher, when requesting a DPoP token (bound to the key) to the IdP, this exception should be catched, at least to add more context.

PR #407, opened 2020-10-09
Feat/dpop/cleanup

NSeydoux:
The refactoring involving oidc-client-authn-js enabled to delete a lot of code from solid-client-authn-browser.
Mostly, this removes all the code that manages in-memory tokens, which includes all the fetchers, and all the dpop package.
The joseutility is removed as well, and cleand up from the DI of some classes.


packages/browser/__tests__/login/oidc/oidcHandlers/LegacyImplicitFlowOIDCHandler.spec.ts
```diff
@@ -35,14 +33,35 @@ import {
...
+};
+
+jest.mock("@inrupt/oidc-dpop-client-browser", () => {
+  return {
+    generateJwkForDpop: async (): Promise<typeof mockJWK> => mockJWK,
+    createDpopHeader: async (
+      _audience: URL,
+      _method: string,
+      _jwt: JSONWebKey
+    ): Promise<string> => "someToken",
```
pmcb55, line 56: Very minor, but I'd move this string to a constant, e.g. `const mockDpopTokenValue = "Some opaque DPoP token value";`, so that `expect()` calls can refer to that same constant rather than hard-coding values.

packages/browser/src/login/oidc/TokenRequester.ts
```diff
@@ -96,18 +96,15 @@ export default class TokenRequester {
       throw new Error(`This issuer [${issuer}] does not have a token endpoint`);
     }
 
-    await this.dpopClientKeyManager.generateClientKeyIfNotAlready();
+    const key = await generateJwkForDpop();
```
pmcb55, line 99: Just a suggestion, but perhaps a nice opportunity to expand the JWK acronym here, making the code slightly more readable, i.e.
`const jsonWebKey = await generateJwkForDpop();`

pmcb55 (approved):
LGTM - fantastic to see so much code disappearing (even though that was always the intent of this whole SAF refactoring effort) - great job though!

package.json
```diff
@@ -5,6 +5,7 @@
     "demo-app": "cd packages/browser; npm run demo-app",
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build",
+    "clean": "rm -rf packages/*/dist/ packages/*/node_modules/",
```
Vinnl, line 8: No need to use the shorthand in a command, and `-f` shouldn't actually be necessary (and thus is probably best left out).

```suggestion
    "clean": "rm --recursive packages/*/dist/ packages/*/node_modules/",
```

packages/browser/__tests__/login/oidc/TokenRequester.spec.ts
```diff
@@ -248,20 +249,22 @@ describe("TokenRequester", () => {
...
+  //   const TokenRefresher = getTokenRequester();
+  //   await expect(
+  //     /* eslint-disable @typescript-eslint/camelcase */
+  //     TokenRefresher.request("global", {
+  //       grant_type: "refresh_token",
+  //       refresh_token: "thisIsARefreshToken",
+  //     })
+  //     /* eslint-enable @typescript-eslint/camelcase */
+  //   ).rejects.toThrowError("The idp returned a bad token without a sub.");
+  // });
```
Vinnl, line 269:
```suggestion
   it.skip("Fails elegantly if the access token does not have a sub claim", async () => {
     setUpMockedReturnValues({
       jwt: {
         iss: "https://idp.com",
       },
     });
     const TokenRefresher = getTokenRequester();
     await expect(
       /* eslint-disable @typescript-eslint/camelcase */
       TokenRefresher.request("global", {
         grant_type: "refresh_token",
         refresh_token: "thisIsARefreshToken",
       })
       /* eslint-enable @typescript-eslint/camelcase */
     ).rejects.toThrowError("The idp returned a bad token without a sub.");
   });
```

Vinnl (approved):
I love the smell of dead code burning in the morning.

Btw, this emphasises why it's important not to add `getClientAuthnWithDependencies` in our public API documentation: once we do (in a v1+ release), changes like these would be a breaking change, since they break for anyone passing their own `dpopClientKeyManager`. Every interface we allow developers to inject, now becomes part of our public API.

(This is also why I'm guessing it actually made sense for some of our unit tests to test whether the right arguments were passed to injected dependencies: because that was not testing the internal implementation, but an actual public API.)

packages/browser/__tests__/login/oidc/oidcHandlers/LegacyImplicitFlowOIDCHandler.spec.ts
NSeydoux: I would have preferred to do this as well, but I don't know how to do this/if it's possible due to `jest` behaviour: the mocks are hoisted at the top of the file when `jest` runs, which makes any constant in the mock closure uninitialized.

packages/browser/__tests__/login/oidc/TokenRequester.spec.ts
NSeydoux: Fixed in bbf7d447ecb4e6ee1b180e8cf7d839c3c991e9b9

package.json
NSeydoux: Fixed in bbf7d447ecb4e6ee1b180e8cf7d839c3c991e9b9

packages/browser/src/login/oidc/TokenRequester.ts
NSeydoux: Fixed in bbf7d447ecb4e6ee1b180e8cf7d839c3c991e9b9

package.json
pmcb55: This is funny - I did actually consider saying basically the same thing as Vinnie here. In general I totally agree, but I thought that in Linux-land, `rm -rf` is soooo common, that actually using `--recursive` might actually be making this **_less_** readable, rather than more readable :) ! Anyway, I'm totally fine with the change - it just struck me as funny.

PR #411, opened 2020-10-12
Implemented the DPoP token exchange

NSeydoux:
This implements the OAuth2 code exchange step: after an auth code has been returned by the IdP through redirection, the client can use the obtained auth code to get an access token. If the request has a DPoP header, the returned token is bound to the DPoP key.


packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import URL from "url-parse";
+import { IClient, IIssuerConfig } from "@inrupt/solid-client-authn-core";
```
Vinnl, line 23: I think I've mentioned this before, but ideally this package does not depend on SCAC.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+import { it, describe } from "@jest/globals";
```
Vinnl, line 24:
```suggestion
import { jest, it, describe } from "@jest/globals";
```

Maybe set `injectGlobals` to `false` in the Jest config so this can be done consistently. (Like https://github.com/inrupt/solid-client-js/pull/493.)

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+        x: "0dGe_s-urLhD3mpqYqmSXrqUZApVV5ZNxMJXg7Vp-2A",
+        y: "-oMe9gGkpfIrnJ0aiSUHMdjqYVm5ZrGCeQmRKoIIfj8",
+        d: "yR1bCsR7m4hjFCvWo8Jw3OfNR4aiYDAFbBD9nkudJKM",
+      };
+    },
+  };
+});
+
+// Note that mockJwk and the mocked generateJwkForDpop from keygen return the
+// same mocked key, but due to jest initialisation, one cannot call the other.
```
Vinnl, line 54: I think you might be able to call `jest.requireMock` in the mock initialisation function?

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+
+async function generateMockJwt(): Promise<void> {
+  const payload = {
+    sub: "https://my.webid",
+    iss: mockIssuer().issuer.toString(),
+  };
+  const jwt = await signJwt(payload, mockJwk(), {
+    algorithm: "ES256",
+  });
+  console.log(jwt.toString());
```
Vinnl, line 105: This might generate a bit too much noise during testing:

```suggestion
  console.log(jwt.toString());
```

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+
+const mockFetch = (payload: string, status: number): typeof fetch => async (
+  _input: RequestInfo,
+  _init?: RequestInit
+): Promise<Response> => new Response(payload, { status });
+
+it.skip("manually generate JWT", async () => {
+  await generateMockJwt();
+});
+
```
Vinnl, line 170:
```suggestion
```

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,206 @@
...
+  let dpopJwk: JSONWebKey | undefined = undefined;
+  if (dpop) {
+    dpopJwk = await generateJwkForDpop();
+    headers["DPoP"] = await createDpopHeader(
+      issuer.tokenEndpoint,
+      "POST",
+      dpopJwk
+    );
+  }
+  // TODO: is this necessary ? it's present in OAuth2 in action book, but not in spec
```
Vinnl, line 169: Are we following up on this anywhere? If so, maybe log that somewhere public and link to that issue here.

Vinnl (approved):
Very clear code, good job :+1:

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: Initially that's the approach I was trying (probably because you suggested it), but that resulted in the types `IClient` and `IIssuerConfig` being exported by this browser-specific module, and imported by the generic `core` module, which didn't feel right. I hesitated to create a small module that would just export these types, making `core` and `oidc-client-dpop-browser` independant, but that felt a bit overkill, even if I guess it would be the 'right' thing to do. What do you think ?

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: Right, but what I meant is that I would like to do the following: 
```
const mockJwk = (): JWKECKey => {
  return {
    kty: "EC",
    kid: "oOArcXxcwvsaG21jAx_D5CHr4BgVCzCEtlfmNFQtU0s",
    alg: "ES256",
    crv: "P-256",
    x: "0dGe_s-urLhD3mpqYqmSXrqUZApVV5ZNxMJXg7Vp-2A",
    y: "-oMe9gGkpfIrnJ0aiSUHMdjqYVm5ZrGCeQmRKoIIfj8",
    d: "yR1bCsR7m4hjFCvWo8Jw3OfNR4aiYDAFbBD9nkudJKM",
  };
};
jest.mock("./keyGen", () => {
  return {
    generateJwkForDpop: mockJwk
});
```
but since `jest` hoists the mocked modules up, that results in `ReferenceError: Cannot access 'mockJwk' before initialization`. And the other, more "regular" issue is that generateJwkForDpop is `async`, which prevents doing `const mockJwk = generateJwkForDpop`, because `mockJwk` is used to initialize mocks at the top level (I think top-level `await` is on the wishlist for next ecmascript release, but not quite there yet).

Did you have another approach in mind ?

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
NSeydoux: There isn't any tickets dedicated to this per se, I wanted to test it as part of SDK-1342 (which would enable end-to-end testing of the DPoP features). I preferred to keep it as a comment to avoid forgetting about it and searching around why requests are failing in case it is actually needed ^^

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: Fixed in e6751b898d0afafee8dbf01b69e9f9711256ea0b

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
Vinnl: Since it's an interface that describes the parameters you have to pass to oidc-dpop-client, that does feel right to me. If it so happens to be that SCAC accepts equal arguments in its own API, then I think it's fair to have it define its own interface as well, that just so happens to be equal to this one. That would also mean that you would be warned if you simply pass on arguments given to SCAC that are no longer compatible with oidc-dpop-client.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
Vinnl: Right, `mockJwk` is not defined when initialising the mocks, but I think it _is_ defined when functions defined in those mocks are executed. So you could do this, I think:

```
jest.mock("./keyGen", () => {
  return {
    generateJwkForDpop: () => mockJwk()
});
```

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
Vinnl: OK, if you're going to get to it in due time, then that's fine I think.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: :bulb: riiiight, very good point ! Fixed in f3b0e84f617039ea8ebf9c4a991a011765f2eb64

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: Since the jest config is shared across the different modules, I'll do it once for all in a separate PR.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
acoburn: Depending on how the client was registered (static/dynamic), the token endpoint may require authentication via headers (as shown here in the commented-out code), or authentication via form params, or no authentication at all. The client would presumably know which mode of operation is in place.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * Extracts a Webid from an ID token based on https://github.com/solid/webid-oidc-spec.
+ * The upcoming spec is still a work in progress.
+ *
+ * Note: this function does not implement the userinfo webid lookup yet.
+ * @param idToken
+ */
+async function deriveWebidFromIdToken(idToken: string): Promise<string> {
```
pmcb55, line 93: I'd change the name to `deriveWebIdFromIdToken` (capital 'I').

packages/oidc-dpop-client-browser/src/dpop/keyGen.ts
```diff
@@ -0,0 +1,53 @@
+/*
```
pmcb55, line 1: There no need to abbreviate the filename here - I'd just call it `keyGeneration.ts`.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+};
+
+export type TokenEndpointInput = {
+  grantType: string;
+  redirectUri: string;
+  code: string;
+  codeVerifier: string;
+};
+
+type WebidOidcIdToken = {
```
pmcb55, line 67: Actually, there are multiple references to names with `webidXXX` - I'd just use `webIdXXX` as a general naming rule.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+        decoded
+      )} is missing 'sub' or 'iss' claims`
+    );
+  }
+  if (decoded.webid) {
+    return decoded.webid;
+  }
+  if (!decoded.sub.match(/^https?:\/\/.+\..+$/)) {
+    throw new Error(
+      `Cannot extract WebID from ID token: the ID token returned by ${decoded.iss} has no webid claim, nor an IRI-like sub claim: [${decoded.sub}]`
```
pmcb55, line 107: Suggest: `Cannot extract WebID from ID token: the ID token returned by [${decoded.iss}] has no 'webid' claim, nor an IRI-like 'sub' claim: [${decoded.sub}]`

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+
+function validateTokenEndpointResponse(
+  tokenResponse: Record<string, unknown>,
+  dpop: boolean
+): Record<string, unknown> & { access_token: string; id_token: string } {
+  if (!hasAccessToken(tokenResponse)) {
+    throw new Error(
+      `Invalid token endpoint response: ${JSON.stringify(
+        tokenResponse
+      )} is missing an access_token.`
```
pmcb55, line 143: Suggest changing to:
`Invalid token endpoint response (missing the field 'access_token'): ${JSON.stringify(
        tokenResponse
      )}`

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+    );
+  }
+
+  if (!hasIdToken(tokenResponse)) {
+    throw new Error(
+      `Invalid token endpoint response: ${JSON.stringify(
```
pmcb55, line 149: Ditto comment above...

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+        tokenResponse
+      )} is missing an id_token.`
+    );
+  }
+
+  if (!hasTokenType(tokenResponse)) {
+    throw new Error(
+      `Invalid token endpoint response: ${JSON.stringify(
```
pmcb55, line 157: Ditto comment above...

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+        tokenResponse
+      )} is missing an token_type.`
+    );
+  }
+
+  if (dpop && tokenResponse.token_type.toLowerCase() !== "dpop") {
+    throw new Error(
+      `Invalid token endpoint response: requested a [DPoP] token, got a token_type [${tokenResponse.token_type}].`
```
pmcb55, line 165: Suggest:
`Invalid token endpoint response: requested a [DPoP] token, but got a 'token_type' value of: [${tokenResponse.token_type}].`

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -0,0 +1,233 @@
...
+    );
+  }
+
+  if (!dpop && tokenResponse.token_type.toLowerCase() !== "bearer") {
+    throw new Error(
+      `Invalid token endpoint response: requested a [Bearer] token, got a token_type [${tokenResponse.token_type}].`
```
pmcb55, line 171: Ditto comment above.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
pmcb55: Interesting. @acoburn - would client registration (i.e. static or dynamic) **_necessarily_** have any bearing on whether the token endpoint requires auth or not? (i.e. I would have thought the server may, or may not, require auth on any endpoint it offers based on a whatever criteria it wants, e.g. always required, never required, only required at certain times of the day, required for requests from certain incoming IP ranges, only required for dynamically registered clients, only required for statically registered clients, etc. etc.)
I assume there's no spec-defined way for a client to 'discover' at runtime whether any endpoint our client is calling requires auth? I'm guessing the book is just showing a simple example of where a server does happen to require Basic auth on it's token endpoint, but since our library can't really know beforehand (and can't determine at runtime either) whether the authorization server we're hitting **_requires_** it or not, I kinda think "just adding it anyway" can't really cause any harm (i.e. if the server doesn't require it, then it'll just ignore it (I think!)), and if it does happen to require it (and it expects the value to be "clientId:clientSecret", which seems reasonable to assume), then we're well-and-good too! So does anyone see any problems with "just adding it anyway" if we have a client secret (i.e. just uncommenting this code)...?

pmcb55 (approved):
LGTM - just minor comments, and a question on the Basic Auth issue.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
NSeydoux: Resolved in 77e1054e4d2abbb463c971b2bde0748739f212aa

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
NSeydoux: Resolved in 77e1054e4d2abbb463c971b2bde0748739f212aa

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
NSeydoux: Resolved in 77e1054e4d2abbb463c971b2bde0748739f212aa

packages/oidc-dpop-client-browser/src/dpop/keyGen.ts
NSeydoux: Resolved in 44808e6e3ada27958ac7fa996102200c21f0f0ea, with a small fix in 77e1054e4d2abbb463c971b2bde0748739f212aa for the mock

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
NSeydoux: Resolved in 759e2b302601170397fd612883e42464a6d3dbf2

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
NSeydoux: Resolved in bf8dd3ebaaff79eecfe21bcd6849caad240879df

PR #414, opened 2020-10-13
Base setup for TestCafe

Vinnl:
This merges Kevin's TestCafe codebase into the SCA repository, and sets it up to run in CI.

A couple of important things to keep in mind compared to Kevin's repository:

- Unfortunately, while TestCafe provides a GitHub Action to run it in CI, it doesn't allow us to specify our own dependencies. Hence, I've had to remove testing-library, unfortunately.
- The tests weren't succeeding for me yet, partly because there was no account yet that could freely run in CI without being greylisted, i.e. without asking for 2FA every once in a while after many automated sign-ins. @pmcb55 will merge in his properly working tests later.
- I extracted the credentials into repository secrets. Note that the credentials that were hardcoded in are still present in the Git history (they weren't too sensitive anyway, I think), but this should be useful for more sensitive credentials.
- Currently only `ess.test.ts` is run. Once we have NSS tests working as well, make sure to modify the GitHub workflow file to execute all tests.
- The "End-to-end tests (browser)" check at the bottom of this PR is how you'll see whether the E2E test was successful.


pmcb55 (approved):
LGTM

PR #415, opened 2020-10-13
Support requesting a DPoP token

NSeydoux:
This integrates the DPoP support implemented in the oidc module into the browser module. For consistency, the bearer token exchange is moved to the oidc module as well, so that obtaining either type of token "feels" the same.
After this, the TokenRequester class could be removed, but it will be done in a different PR so that it appears in the git history, so that it is easier to look for.

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "demo-app": "cd packages/browser; npm run demo-app",
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build",
-    "clean": "rm --recursive packages/*/dist/ packages/*/node_modules/",
+    "clean": "rm --recursive --force packages/*/dist/ packages/*/node_modules/",
     "build-api-docs": "lerna run build-api-docs",
     "format-all": "prettier --write \"packages/*/src/**\" \"packages/*/__tests__/**\"",
     "licenses-all": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
--- a/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -20,62 +20,142 @@
  */
 
 import "reflect-metadata";
+import URL from "url-parse";
 import {
   StorageUtilityMock,
   mockStorageUtility,
+  IClient,
+  IClientRegistrar,
+  IClientRegistrarOptions,
+  IIssuerConfigFetcher,
 } from "@inrupt/solid-client-authn-core";
-import AuthCodeRedirectHandler from "../../../../src/login/oidc/redirectHandler/AuthCodeRedirectHandler";
+import AuthCodeRedirectHandler, {
+  exchangeDpopToken,
+} from "../../../../src/login/oidc/redirectHandler/AuthCodeRedirectHandler";
 import { RedirectorMock } from "../../../../src/login/oidc/__mocks__/Redirector";
 import { SessionInfoManagerMock } from "../../../../src/sessionInfo/__mocks__/SessionInfoManager";
-import { SigninResponse } from "@inrupt/oidc-dpop-client-browser";
+import {
+  IIssuerConfig,
+  TokenEndpointResponse,
+} from "@inrupt/oidc-dpop-client-browser";
+import { JSONWebKey } from "jose";
 
 jest.mock("cross-fetch");
 
-// The following key has been used to sign the mock access token. It is given
-// for an information purpose.
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-const _mockJwk = {
-  kty: "EC",
-  kid: "oOArcXxcwvsaG21jAx_D5CHr4BgVCzCEtlfmNFQtU0s",
-  alg: "ES256",
-  crv: "P-256",
-  x: "0dGe_s-urLhD3mpqYqmSXrqUZApVV5ZNxMJXg7Vp-2A",
-  y: "-oMe9gGkpfIrnJ0aiSUHMdjqYVm5ZrGCeQmRKoIIfj8",
-  d: "yR1bCsR7m4hjFCvWo8Jw3OfNR4aiYDAFbBD9nkudJKM",
+const mockJwk = (): JSONWebKey => {
+  return {
+    kty: "EC",
+    kid: "oOArcXxcwvsaG21jAx_D5CHr4BgVCzCEtlfmNFQtU0s",
+    alg: "ES256",
+    crv: "P-256",
+    x: "0dGe_s-urLhD3mpqYqmSXrqUZApVV5ZNxMJXg7Vp-2A",
+    y: "-oMe9gGkpfIrnJ0aiSUHMdjqYVm5ZrGCeQmRKoIIfj8",
+    d: "yR1bCsR7m4hjFCvWo8Jw3OfNR4aiYDAFbBD9nkudJKM",
+  };
 };
 
-// Payload: { sub: "https://my.webid" }
-const mockAccessToken =
-  "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0ZXN0Q2xhaW0iOiJ0ZXN0VmFsdWUiLCJpYXQiOjE2MDIxNTg2NDJ9.wGZ49jU3wNSAFvWvZsjjulmbfRjlIQMp0VY0Q5u2--5vyzeKwfGUmssOW8kftIXG1ikm2iqMb6YRXCO4KGEctQ";
+const mockWebId = (): string => "https://my.webid";
 
-jest.mock("@inrupt/oidc-dpop-client-browser", () => {
+// result of generateMockJwt()
+const mockIdToken = (): string =>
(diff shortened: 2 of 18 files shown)


NSeydoux:
Manually tested against ESS and NSS: DPoP support is back :D

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -170,11 +270,64 @@ describe("AuthCodeRedirectHandler", () => {
...
+describe("exchangeDpopToken", () => {
+  it("requests a key-bound token", async () => {
+    const mockedIssuer = mockIssuer();
+    const tokens = await exchangeDpopToken(
+      "mySession",
+      mockedIssuer.issuer,
+      mockIssuerConfigFetcher(mockedIssuer),
+      mockClientRegistrar(mockClient()),
+      "some code",
+      "some pkcd token",
```
pmcb55, line 324: I assume PCKE token (typo)...?

packages/browser/src/login/oidc/OidcLoginHandler.ts
```diff
@@ -127,7 +127,7 @@ export default class OidcLoginHandler implements ILoginHandler {
     const OidcOptions: IOidcOptions = {
       issuer: options.oidcIssuer as URL,
       // TODO: differentiate if DPoP should be true
-      dpop: true,
+      dpop: options.tokenType === "DPoP",
```
pmcb55, line 130: Are we really case-sensitive, or should this be a case-insensitive conditional...?

pmcb55 (approved):
LGMT - reviewed it quickly, but looks good.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -71,46 +113,60 @@ export default class AuthCodeRedirectHandler implements IRedirectHandler {
...
+    let tokens: TokenEndpointResponse | TokenEndpointDpopResponse;
+    let authFetch: typeof fetch;
+
+    if (isDpop) {
+      // Since we throw if not found, the type assertion is okay
+      const issuer = (await this.storageUtility.getForUser(
+        storedSessionId,
+        "issuer",
+        { errorIfNull: true }
+      )) as string;
```
Vinnl, line 129: Here too seems like `getForUser` could have type overloads that indicate that it returns a `string` if `errorIfNull` is set to `true`.

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -170,11 +270,64 @@ describe("AuthCodeRedirectHandler", () => {
       await redirectInfo.fetch("https://some.other.url");
       // @ts-ignore
       const header = fetch.fetch.mock.calls[0][1].headers["Authorization"];
-      // We test that the Authorization header matches the structure of a JWT.
       expect(
         // @ts-ignore
         header
-      ).toMatch(/^Bearer .+\..+\..+$/);
+      ).toMatch(/^Bearer some token$/);
```
Vinnl, line 276: I was a bit confused by this test. I was assuming it would do an unauthenticated fetch by default, but apparently some token is being passed in here (`some token`), but it's not clear to me where it comes from. Presumably from one of the mocks, but if it's not too much effort, maybe that value ("some token") can be explicitly passed to the mocks from this test, so it's clear where the assertion came from?

Vinnl (approved):
Looks good!

package.json
```diff
@@ -5,7 +5,7 @@
     "demo-app": "cd packages/browser; npm run demo-app",
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build",
-    "clean": "rm --recursive packages/*/dist/ packages/*/node_modules/",
+    "clean": "rm --recursive --force packages/*/dist/ packages/*/node_modules/",
```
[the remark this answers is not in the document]
NSeydoux: There are some files that appear to be write-protected in `node_modules/.cache` :)

packages/browser/src/login/oidc/OidcLoginHandler.ts
NSeydoux: Good point: TS users will have that checked for them (the expected type for tokenType is `"DPoP"|"Bearer"`, so you get auto-completion for this casing), but plain JS users won't necessarily notice that. Fixed in 03f44746ffd14f9f41c62c3e5b702ae386b4b1ea. In this case though, we rely on TS to prevent any other value to be passed. Should we throw on something with is neither dpop or bearer, casing aside ?

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.spec.ts
```diff
@@ -91,10 +96,10 @@ async function generateMockJwt(): Promise<void> {
     sub: "https://my.webid",
     iss: mockIssuer().issuer.toString(),
   };
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
NSeydoux: Absolutely, I had a bunch of debug cosole.log and this one got removed by accident, and then I mindlessly added the eslint comment ^^. Fixed in e905752f829366001a60118f65a992d61e72227c.

package.json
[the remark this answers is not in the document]
Vinnl: Ah clear, thanks.

packages/browser/src/login/oidc/OidcLoginHandler.ts
Vinnl: Yeah, that sounds like a good idea.

pmcb55: Yep - makes sense to throw alright.

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -228,7 +239,62 @@ export async function getTokens(
...
+ * been used to issue the initial redirect.
+ * @param redirectUrl The URL to which the user has been redirected
+ */
+export async function getBearerToken(
+  redirectUrl: URL
+): Promise<TokenEndpointResponse> {
+  let signinResponse;
+  try {
+    signinResponse = await new OidcClient({
+      // TODO: We should look at the various interfaces being used for storage,
```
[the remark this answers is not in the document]
NSeydoux: There is now :p https://inrupt.atlassian.net/browse/SDK-1477

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: That is a great idea. However, I just gave it a try, and the fact that the functions to overload are class methods rather than regular functions give a rather strange behaviour, as if only the latest type applied. I created a ticket to follow this up.

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
NSeydoux: Since the value is returned by the mocked module, it's a bit hard to explicitly pass the value as part of the test, but I added a comment to clarify things a bit in ba04ef918e13a08a019f8d5cb1556c808cb8a415

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -228,7 +239,62 @@ export async function getTokens(
...
+      //  (which has an interface Storage), and our own proprietary interface
+      //  IStorage - i.e. we should really just be using the browser Web Storage
+      //  API, e.g. "stateStore: window.localStorage,".
+
+      // We are instantiating a new instance here, so the only value we need to
+      // explicitly provide is the response mode (default otherwise will look
+      // for a hash '#' fragment!).
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      response_mode: "query",
+      // The userinfo endpoint on NSS fails, so disable this for now
```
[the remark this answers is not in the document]
NSeydoux: Fixed in ba04ef918e13a08a019f8d5cb1556c808cb8a415

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -161,11 +167,13 @@ function validateTokenEndpointResponse(
     );
   }
 
-  if (dpop && tokenResponse.token_type.toLowerCase() !== "dpop") {
-    throw new Error(
-      `Invalid token endpoint response: requested a [DPoP] token, but got a 'token_type' value of [${tokenResponse.token_type}].`
-    );
-  }
+  // TODO: Due to what seems to be a bug in the ID broker, a DPoP token is returned
```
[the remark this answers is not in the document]
NSeydoux: Fixed in 41d010d394610b1fb7b1ea8cb1557f8164b3929d

packages/oidc-dpop-client-browser/src/dpop/tokenExchange.ts
```diff
@@ -228,7 +239,62 @@ export async function getTokens(
...
+
+export async function getDpopToken(
+  issuer: IIssuerConfig,
+  client: IClient,
+  data: TokenEndpointInput
+): Promise<TokenEndpointDpopResponse> {
+  // The type assertion is okay, because the absence of the jwk would throw an error
+  return getTokens(issuer, client, data, true) as Promise<
+    TokenEndpointDpopResponse
+  >;
```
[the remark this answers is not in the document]
NSeydoux: Good point, fixed in ba04ef918e13a08a019f8d5cb1556c808cb8a415

NSeydoux:
Dammit jest cache

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
Vinnl: Ah that's weird. Just to make sure, when you add overloads, you also need to add one for the actual function definition, as that gets overridden.

So for example:

```
function aFunction(paramA: true): string;
function aFunction(paramA: boolean): string | null;
function aFunction(paramA: boolean): string | null {
  /* Return a string on true, or either when it's not. */
}
```

NSeydoux: Ah, I wasn't doing that, so that may have caused the issue (Did I shamelessly blame the tool because I did not know how to use it properly ? Maybe.) But even writing
```
async getForUser(
    userId: string,
    key: string,
    options: { errorIfNull: true; secure?: boolean }
  ): Promise<string>;
  async getForUser(
    userId: string,
    key: string,
    options?: { errorIfNull?: boolean; secure?: boolean }
  ): Promise<string | undefined>;
  async getForUser(
    userId: string,
    key: string,
    options?: { errorIfNull?: boolean; secure?: boolean }
  ): Promise<string | undefined>
```
the type assertion is required, otherwise the TS compiler complains.

Vinnl: Hmm, might be because the nested object parameter is too hard for TS to pattern match? :shrug:

PR #428, opened 2020-10-15
Return unauthenticated session on invalid redirect url

NSeydoux:
If the URL does not contain redirection info as query params, just return an unauthenticated session instead of throwing an error.

Changing this prevents the consuming app to have to worry about the type of IRI it is passing to the library.

### DIFF
--- a/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
+++ b/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
@@ -47,8 +47,7 @@ import {
   buildDpopFetch,
 } from "../../../authenticatedFetch/fetchFactory";
 import { JSONWebKey } from "jose";
-import { v4 } from "uuid";
-import { fetch } from "cross-fetch";
+import { getUnauthenticatedSession } from "../../../sessionInfo/SessionInfoManager";
 
 export async function exchangeDpopToken(
   sessionId: string,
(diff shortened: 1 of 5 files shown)


packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -98,9 +99,13 @@ export default class AuthCodeRedirectHandler implements IRedirectHandler {
     redirectUrl: string
   ): Promise<ISessionInfo & { fetch: typeof fetch }> {
     if (!(await this.canHandle(redirectUrl))) {
-      throw new ConfigurationError(
-        `Cannot handle redirect url [${redirectUrl}]`
-      );
+      // If the received IRI does not have redirection information, we can only
+      // return an unauthenticated session.
+      return {
```
ajacksified, line 104: Should the "logged out session" be a constant somewhere?

NSeydoux: I initially meant to have it as a constant, but it was so small that I figured it wasn't necessary. If you feel like it could be useful, I'll change it.

Vinnl (approved):
Makes sense to me.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Fixed in e21486131070bc33b3f09d7a7f55e3030dda9771. While I was at it, I made sure the implicit flow behaves consistently.

ajacksified (approved):

PR #431, opened 2020-10-16
Prevents DCR from hanging in React components

NSeydoux:
This PR fixes two bugs impacting podbrowser:

- The DCR hanged without any error message. Now, it uses fetch instead of window.fetch, the bundler fills the appropriate fetch in, and things seem to work out.
- Handling an IRI with no querystrings resulted in a throw. This had been fixed in cc939e9, but the fix actually introduced another bug preventing the implicit flow to execute. This new version of the fix is more in line with the handler pattern: a fallback handler is introduced, and returns an unauthenticated session if neither the auth code or the access token are present in the URL.

### DIFF
--- a/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
@@ -168,7 +168,7 @@ describe("ClientRegistrar", () => {
           }
         )
       ).rejects.toThrowError(
-        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - ]"
+        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - undefined]"
       );
     });
 
--- a/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
@@ -168,7 +168,7 @@ describe("ClientRegistrar", () => {
           }
         )
       ).rejects.toThrowError(
-        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - undefined]"
+        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - ]"
       );
     });
 
--- a/packages/browser/__tests__/login/oidc/redirectHandler/ImplicitRedirectHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/redirectHandler/ImplicitRedirectHandler.spec.ts
@@ -25,7 +25,7 @@ import {
   SessionInfoManagerMock,
   SessionCreatorCreateResponse,
 } from "../../../../src/sessionInfo/__mocks__/SessionInfoManager";
-import GeneralRedirectHandler from "../../../../src/login/oidc/redirectHandler/GeneralRedirectHandler";
+import { ImplicitRedirectHandler } from "../../../../src/login/oidc/redirectHandler/ImplicitRedirectHandler";
 
 jest.mock("cross-fetch");
 
@@ -46,16 +46,16 @@ const _mockJWK = {
 const mockAccessToken =
   "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0ZXN0Q2xhaW0iOiJ0ZXN0VmFsdWUiLCJpYXQiOjE2MDIxNTg2NDJ9.wGZ49jU3wNSAFvWvZsjjulmbfRjlIQMp0VY0Q5u2--5vyzeKwfGUmssOW8kftIXG1ikm2iqMb6YRXCO4KGEctQ";
 
-describe("GeneralRedirectHandler", () => {
+describe("ImplicitRedirectHandler", () => {
   const defaultMocks = {
     tokenSaver: TokenSaverMock,
     sessionInfoManager: SessionInfoManagerMock,
   };
 
-  function getGeneralRedirectHandler(
+  function getImplicitRedirectHandler(
     mocks: Partial<typeof defaultMocks> = defaultMocks
-  ): GeneralRedirectHandler {
-    return new GeneralRedirectHandler(
+  ): ImplicitRedirectHandler {
+    return new ImplicitRedirectHandler(
       mocks.tokenSaver ?? defaultMocks.tokenSaver,
       mocks.sessionInfoManager ?? defaultMocks.sessionInfoManager
     );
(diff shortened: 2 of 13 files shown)


packages/browser/__tests__/login/oidc/redirectHandler/ImplicitRedirectHandler.spec.ts
```diff
@@ -89,15 +91,15 @@ describe("GeneralRedirectHandler", () => {
...
   describe("handle", () => {
-    it("returns an unauthenticated session on non-redirect URL", async () => {
-      const authCodeRedirectHandler = getGeneralRedirectHandler();
-      const mySession = await authCodeRedirectHandler.handle("https://my.app");
-      expect(mySession.isLoggedIn).toEqual(false);
-      expect(mySession.webId).toBeUndefined();
+    it("throws on non-redirect URL", async () => {
+      const redirectHandler = getImplicitRedirectHandler();
+      await expect(redirectHandler.handle("https://my.app")).rejects.toThrow(
+        "ImplicitRedirectHandler cannot handle [https://my.app]"
```
pmcb55, line 97: Can this message explain 'why' it can't handle that valid IRI? (There's nothing worse than a message like "Internal Server Error" :( !)

packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -168,7 +168,7 @@ describe("ClientRegistrar", () => {
           }
         )
       ).rejects.toThrowError(
-        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - ]"
+        "Client registration failed: [Error: Dynamic client registration failed: bad stuff that's an error - undefined]"
```
pmcb55, line 171: I haven't seen the code that throws yet, but our test expecting an error message with 'undefined' in it doesn't 'feel' right...?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -86,23 +85,26 @@ export default class AuthCodeRedirectHandler implements IRedirectHandler {
...
-    );
+    try {
+      const myUrl = new URL(redirectUrl);
+      return (
+        myUrl.searchParams.get("code") !== null &&
+        myUrl.searchParams.get("state") !== null
+      );
+    } catch {
+      throw new Error(
+        `[${redirectUrl}] is not a valid URL, and cannot be used as a redirect URL.`
```
pmcb55, line 96: Can't we catch the actual error thrown (you just have `} catch {` there now), and report the description within that exception in the message here (I don't think it would be a security problem to report that)...?

packages/browser/src/login/oidc/redirectHandler/FallbackRedirectHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+@injectable()
+export class FallbackRedirectHandler implements IRedirectHandler {
+  async canHandle(redirectUrl: string): Promise<boolean> {
+    try {
+      new URL(redirectUrl);
+      return true;
+    } catch {
+      throw new Error(
+        `[${redirectUrl}] is not a valid URL, and cannot be used as a redirect URL.`
```
pmcb55, line 49: Ditto to comment before...

packages/browser/src/login/oidc/redirectHandler/ImplicitRedirectHandler.ts
```diff
@@ -29,41 +29,43 @@ import {
...
+    }
   }
   async handle(
     redirectUrl: string
   ): Promise<ISessionInfo & { fetch: typeof fetch }> {
     if (!(await this.canHandle(redirectUrl))) {
-      // If the received IRI does not have redirection information, we can only
-      // return an unauthenticated session.
-      return getUnauthenticatedSession();
+      throw new Error(`ImplicitRedirectHandler cannot handle [${redirectUrl}]`);
```
pmcb55, line 66: I'd be more explicit in this message. If I understand correctly, the reason we throw here it that we're assuming we are the last handler in the chain of all handlers - i.e. that we are the handler of last-resort - is that correct? If so, I'd somehow allude to that in this message, as it seems like important logic to understand.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
ajacksified: I think the additional context is more useful to a developer. (That said, eventually we'll want to consider typed errors instead of strings, and preferably use strings from a vocab.)

NSeydoux: Typed errors sound like a great idea, we should definitively right a couple of stories to have that in the backlog. The Java LIT has an Exception support module if I recall correctly, that's something that we might want to look into, because here the additional context is rather poor (`TypeError: Invalid URL: beep boop I am a robot`), which is why I wasn't sure about adding it in the first place ^^

packages/browser/src/login/oidc/redirectHandler/ImplicitRedirectHandler.ts
NSeydoux: To the contrary, we are throwing there because we are **not** the last handler: the newly created `FallbackRedirectHandler` is the one who should get called last, because as long as the URL you give to it is valid, it says it can handle it, and return an unauthenticated session. 

The `ImplicitRedirectHandler` and `AuthCodeRedirectHandler` throwing in their `handle` method is  possible because in the way we implement the handler pattern, `handle` should never be called if `canHandle` returns false. That's why I added a `FallbackRedirectHandler`, so that the redirect handler aggregator would always find a suitable handler, and a user can always use `handleRedirect` on any IRI, without having it throw an error in case the IRI is not a redirect one. In such case, you'd just get an unauthenticated session.

NSeydoux: This context is (tentatively) given in the `FallbackRedirectHandler` class and in the `dependencies` file, where the handler is registered.

packages/browser/src/login/oidc/redirectHandler/FallbackRedirectHandler.ts
NSeydoux: Resolved in 413e960a5c9fe7aca0ffad42c68d62293196dc25

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 413e960a5c9fe7aca0ffad42c68d62293196dc25

packages/browser/__tests__/login/oidc/ClientRegistrar.spec.ts
NSeydoux: Fixed in eb47a0c19491f23bfbffe99a06c61634d39300ae

pmcb55 (commented):
LGTM (was this merged before being approved)

NSeydoux:
Oh my bad I didn't realized it hadn't been approved: since we are not working against master, the branch protection rules don't apply.

pmcb55:
No worries - I was just wondering how, so good to know it was the non-master-branch-ness that allowed the merge!

PR #438, opened 2020-10-19
Allow using testing-library in our end-to-end tests

Vinnl:
This PR makes TestCafe in CI run by just installing our npm dependencies and running the TestCafe command, rather than relying on the pre-built GitHub Action by TestCafe. This means that we're in charge of its installation, and so when testing-library points to our installation of TestCafe, that's the correct one that's actually being used.


kevin-inrupt (approved):
Successfully configured and ran the tests locally.

pmcb55 (approved):
LGTM (assuming it works!)

e2e/browser/package.json
```diff
@@ -4,9 +4,13 @@
   "description": "",
   "main": "index.ts",
   "scripts": {
-    "test": "npx testcafe all ess.test.ts"
+    "test": "testcafe firefox:headless,chrome:headless testSuiteRunner.test.ts"
```
pmcb55, line 7: Just curious - why replace `all`? Is this just 'cos only firefox and chrome can run headlessly? And is the change to headless just for the speed improvement (which I'd totally buy into), but it seems a bit dangerous to leave our Edge and Safari, no...?

Vinnl: We need to run it headless in CI as there's no display server there. I'd have preferred to do `all:headless`, but that's not possible yet: https://github.com/DevExpress/testcafe/issues/5641

(I'm not sure whether Safari can do headless. Both it and Edge are not available in every OS, so we'd have to modify our CI to run in different OS's and then adjust the list based on the OS. Which is a good idea, but for a future PR - especially if that issue gets implemented before that time.)

pmcb55: @Vinnl - I see, very interesting - cheers (great job creating the TestCafe issue too!).

PR #445, opened 2020-10-20
Chore/add eslint for testcafe

pmcb55:
Chore: just internal refactor to allow ESLint to be run (with license checker, and Prettier).

NOTE: this is not yet integrated into the full Lerna scripts - this is still a separate project that needs manual linting.


NSeydoux (approved):

NSeydoux:
Since we are moving away from testcafé and have significantly refactored (and still are experimenting with) the shared linting config, I'll close this one, but will make sure that the intent is reflected using the shared config.

PR #459, opened 2020-10-21
URL and localStorage cleanup

NSeydoux:
This PR fixes two bugs: the query strings not being removed of the redirect IRI, and the local storage not being cleared by oidc-client.

Note: still missing some tests, that's why this is a draft.

### DIFF
--- a/packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
+++ b/packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
@@ -110,7 +110,7 @@ describe("SessionInfoManager", () => {
       const sessionManager = getSessionInfoManager({
         storageUtility: mockStorageUtility({}, true),
       });
-      await sessionManager.clear("mySession");
+      await sessionManager.clear("Value of sessionId doesn't matter");
       expect(mockClearFunction).toHaveBeenCalled();
     });
 
@@ -132,7 +132,7 @@ describe("SessionInfoManager", () => {
       ).toBeUndefined();
     });
 
-    it("clears local storage from user data", async () => {
+    it("clears local unsecure storage from user data", async () => {
       const mockStorage = mockStorageUtility(
         {
           mySession: {
--- a/packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
+++ b/packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
@@ -24,7 +24,15 @@ import { UuidGeneratorMock } from "../../src/util/__mocks__/UuidGenerator";
 import { AuthenticatedFetcherMock } from "../../src/authenticatedFetch/__mocks__/AuthenticatedFetcher";
 import { LogoutHandlerMock } from "../../src/logout/__mocks__/LogoutHandler";
 import { mockStorageUtility } from "@inrupt/solid-client-authn-core";
-import SessionInfoManager from "../../src/sessionInfo/SessionInfoManager";
+import { SessionInfoManager } from "../../src/sessionInfo/SessionInfoManager";
+
+const mockClearFunction = jest.fn();
+
+jest.mock("@inrupt/oidc-dpop-client-browser", () => {
+  return {
+    clearOidcPersistentStorage: async (): Promise<void> => mockClearFunction(),
+  };
+});
 
 describe("SessionInfoManager", () => {
   const defaultMocks = {
@@ -94,6 +102,61 @@ describe("SessionInfoManager", () => {
       });
       const session = await sessionManager.get("commanderCool");
       expect(session).toBeUndefined();
+    });
+  });
+
+  describe("clear", () => {
+    it("clears oidc data", async () => {
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorageUtility({}, true),
+      });
+      await sessionManager.clear("mySession");
+      expect(mockClearFunction).toHaveBeenCalled();
+    });
+
+    it("clears local secure storage from user data", async () => {
+      const mockStorage = mockStorageUtility(
+        {
+          mySession: {
+            key: "value",
+          },
+        },
+        true
+      );
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorage,
+      });
+      await sessionManager.clear("mySession");
+      expect(
+        await mockStorage.getForUser("mySession", "key", { secure: true })
+      ).toBeUndefined();
+    });
+
+    it("clears local storage from user data", async () => {
+      const mockStorage = mockStorageUtility(
+        {
+          mySession: {
+            key: "value",
+          },
+        },
+        false
+      );
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorage,
+      });
+      await sessionManager.clear("mySession");
+      expect(
+        await mockStorage.getForUser("mySession", "key", { secure: false })
+      ).toBeUndefined();
+    });
+  });
+
+  describe("getAll", () => {
+    it("is not implemented", async () => {
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorageUtility({}),
+      });
+      await expect(sessionManager.getAll).rejects.toThrow("Not implemented");
     });
   });
 });
--- a/packages/browser/src/ClientAuthentication.ts
+++ b/packages/browser/src/ClientAuthentication.ts
@@ -34,8 +34,9 @@ import {
   ISessionInfo,
   ISessionInfoManager,
 } from "@inrupt/solid-client-authn-core";
-import URL from "url-parse";
+import UrlParse from "url-parse";
 import { IFetcher } from "./util/Fetcher";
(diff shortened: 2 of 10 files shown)


packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
```diff
@@ -96,4 +104,59 @@ describe("SessionInfoManager", () => {
       expect(session).toBeUndefined();
     });
   });
+
+  describe("clear", () => {
+    it("clears oidc data", async () => {
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorageUtility({}, true),
+      });
+      await sessionManager.clear("mySession");
```
pmcb55, line 113: I don't think the value of the session param matters for this test, so I'd suggest:
```suggestion
      await sessionManager.clear("Value of sessionId doesn't matter");
```

packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
```diff
@@ -96,4 +104,59 @@ describe("SessionInfoManager", () => {
...
+      const sessionManager = getSessionInfoManager({
+        storageUtility: mockStorage,
+      });
+      await sessionManager.clear("mySession");
+      expect(
+        await mockStorage.getForUser("mySession", "key", { secure: true })
+      ).toBeUndefined();
+    });
+
+    it("clears local storage from user data", async () => {
```
pmcb55, line 135:
```suggestion
    it("clears local unsecure storage from user data", async () => {
```

packages/browser/src/ClientAuthentication.ts
```diff
@@ -34,8 +34,9 @@ import {
   ISessionInfo,
   ISessionInfoManager,
 } from "@inrupt/solid-client-authn-core";
-import URL from "url-parse";
+import UrlParse from "url-parse";
```
pmcb55, line 37: I'm not sure about this, but it seems to me that the only export from the `url-parse` library is actually named `URLParse` (note the capitalization), so changing that capitalization seems a little disingenuous(!): 
```suggestion
import URLParse from "url-parse";
```

packages/browser/src/ClientAuthentication.ts
```diff
@@ -54,12 +55,12 @@ export default class ClientAuthentication {
     private environmentDetector: IEnvironmentDetector
   ) {}
 
-  private urlOptionToUrl(url?: URL | string): URL | undefined {
+  private urlOptionToUrl(url?: UrlParse | string): UrlParse | undefined {
```
pmcb55, line 58: This function is still badly named - but hopefully it'll disappear completely if we can get rid of `URLParse` - so no point in suggesting a better name right now (like `convertToUrlParse()` !).

packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -59,13 +60,14 @@ export async function clear(
     // FIXME: This is needed until the DPoP key is stored safely
     storage.delete("clientKey", { secure: false }),
   ]);
+  await clearOidcPersistentStorage();
```
pmcb55, line 63: Won't this clear **_all_** `oidc-client-js` state, yet this method is supposed to be session-specific? Anyway, there are bigger problems right now with this whole area...

packages/oidc-dpop-client-browser/src/cleanup/cleanup.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import { OidcClient, WebStorageStateStore } from "oidc-client";
+
+/**
+ * Removes OIDC-specific query parameters from a given URL (state, code...)
+ * @param redirectUrl The URL to clean up.
+ * @returns A copy of the URL, without OIDC-specific query params.
+ */
+export function cleanupRedirectUrl(redirectUrl: string): string {
```
pmcb55, line 29: This function name could be more specific (e.g. it's not really redirect-URL-specific at all (i.e. it just takes in any string (and I guess could throw if that string is an invalid URL!?)), and it specifically and explicitly removes very specific query params (as opposed to all possible OIDC-specific query params). I know this whole PR is a horrific hack, but I'd rename this function just in case this code survives more than a few days!
```suggestion
export function removeQueryParamCodeAndState(url: string): string {
```

packages/oidc-dpop-client-browser/src/cleanup/cleanup.ts
```diff
@@ -0,0 +1,53 @@
...
+  const cleanedUrl = new URL(redirectUrl);
+  cleanedUrl.searchParams.delete("code");
+  cleanedUrl.searchParams.delete("state");
+  return cleanedUrl.toString();
+}
+
+/**
+ * Clears any OIDC-related data lingering in the local storage.
+ */
+export async function clearOidcPersistentStorage(): Promise<void> {
```
pmcb55, line 39: This is specifically referring to `oidc-client-js`, right? If so, I'd try and make that a little bit more specific:
```suggestion
export async function clearOidcClientPersistentStorage(): Promise<void> {
```
Regardless, this seems like a very heavy-handed, and a very 'implicit' thing to be doing - i.e instantiating a new instance of an object just to call a 'clear state' method on that instance (i.e. we both know how `oidc-client-js` works, but this behaviour seems very counter-intuitive to me). But anyway, this whole PR requires more serious consideration to fix up 'properly' later I think (i.e. an `OidcClient` instance per session maybe, or just using their `UserManager` from the start...).

pmcb55 (approved):
I'll approve this PR to get the fix in - but the whole thing seems to be a horrific hack - i.e. I think the whole PR requires more serious consideration to fix the overall code up 'properly' later (i.e. an `OidcClient` instance per session maybe (which I originally tried to do, but the whole redirect stage kinda suppered!), or just using their `UserManager` from the start...).
But I can certainly appreciate now why it took a lot of effort to first understand what the hell was going on, and then to construct this work-around - so this is a really fantastic job, well done :) !

packages/oidc-dpop-client-browser/src/cleanup/cleanup.ts
NSeydoux: I kept the name generic, because I expect the same issue will apply in the implicit flow with `id_token` and `access_token`, but the function could be named `removeOidcQueryParam`

packages/browser/src/ClientAuthentication.ts
NSeydoux: We are using the default import from `url-parse`, so the name we give it is up to us, and in its current form, it is aligned with our naming convention.

packages/browser/src/sessionInfo/SessionInfoManager.ts
NSeydoux: Given that `oidc-client-js` is browser-specific, I reckon they don't expect multiple sessions to coexist in the same browser, so I guess they don't really care about making the distinction.

packages/oidc-dpop-client-browser/src/cleanup/cleanup.ts
NSeydoux: Internally, yes, it is specific to `oidc-client-js`, but since this function is exposed as part of the API of `oidc-dpop-client-js` I guessed it wasn't needed to be specific about the underlying library.

Moreover, I think their approach is motivated by the fact that in the browser, it's impossible to have one `OidcClient` instance per session, since there is a redirection in the OIDC flow that necessarily will lead to the page being reloaded. They must assume that there is only one session active at a time, and their architecture is based on this assumption.

packages/oidc-dpop-client-browser/src/cleanup/cleanup.ts
NSeydoux: Fixed in ea0e3a72b23994927eb8b8869daf85e5352fbabe

packages/browser/src/ClientAuthentication.ts
pmcb55: Hhhmmm.... I can't say I agree with that argument (but then again, maybe this genuinely is just idiomatic JavaScript common practice - but if so, I'd be surprised (and disappointed!)).
Right now, we're simply using this 3rd-party library to perform functionality that we need, specifically to be able to parse an IRI into it's constituent parts. Sure, we can rename this class to whatever we like, but why would we want to rename a 3rd-party class (and such a popularly used one at that)? Isn't that just going to introduce potential confusion (i.e. it certainly confused me already!)...?
Given it's a popular 3rd-party library (actually, regardless of it's popularity, but being popular emphasizes the point I suppose), I see no reason to rename it just to conform to our particular naming conventions - in fact I'd see that as a distinctly bad idea ('cos doing so makes it appear to new developers as one of 'our classes', when in fact it's not).
I don't know if this helps, but it strikes me as an argument for arbitrarily renaming something super-popular like Moment as `import DateUtils from "moment";`, just 'cos I have a date-centric naming convention in my particular codebase.
I'd be interested to hear @Vinnl 's or @ajacksified 's thoughts on this...?

PR #460, opened 2020-10-21
first phase - remove ILoginInputOptions

pmcb55:
# Internal refactor

Really just the first phase of an internal tidy up that removes the unnecessary `ILoginInputOptions` interface. A much bigger problem is the use of `URLParse` throughout the codebase, but also the use of `IOidcOptions` (which seems overly complex). So I'm thinking we merge this first phase as a first step (also due to recent changes from @NSeydoux that touch on this area).


Vinnl:
Yep, url-parse should not be necessary in a browser-based environment at all (except in IE, but then developers should include their own polyfills), so if you can get rid of it entirely, that'd be perfect.

NSeydoux (commented):
Actually, I'm not sure we want to remove the distinction between `InputOptions` and internal `Options`: the idea is that input options are exposed to developers, so they should be minimal, to prevent them from having to make any choice that is not absolutely required. The internal options can be a bigger object, with more parameters that we fill in according to additional information. The idea of the ticket was to align the URL types between these two option types, rather than replacing one with the other. What do you think ?

NSeydoux:
After looking at the code in more details, `ILoginOptions` has one additional property compared to  `ILoginInputOptions`: `sessionId`, which is required. I don't think we want `sessionId` to be required when calling `mySession.login(...)`, so I would declare `ILoginOptions` as an extension of `ILoginInputOptions` with the one additional property. What do you think ?

pmcb55:
Yeah, closing this now and in a separate PR just going to extend one interface from the other (to remove the duplication of most of the fields). Ideally I'd like to get rid of the extra interface completely, and just pass the session ID as a required parameter along with instances of the `ILoginInputOptions` object (rather than extending and adding a session ID field), but that's more work right now, and not worth the effort.

PR #465, opened 2020-10-22
Coverage 100% for Core (mostly secure storage) (#368)

NSeydoux:
This re-applies https://github.com/inrupt/solid-client-authn-js/commit/a28b0b6e2c29b7e72d8f20dd99536425568906b9, which seems to have been lost in when merging the diverged `1.x` branch.


pmcb55 (approved):
LGTM - yep, this seems to be the correct commit!

PR #475, opened 2020-10-23
Prevents handleIncomingRedirect to hit token endpoint twice

NSeydoux:
Some token endpoints only accept the auth code once, so this prevents from hitting it twice for the same session.

Note: this won't include the non-regression tests, as currently Session has no tests at all (!), and I'm writing the tests from scratch in a differnet branch. I'll add this test there.

- [x] I've added a unit test to test for potential regressions of this bug. **See commit notes**


NSeydoux:
This does not prevent two requests to be submitted in parallel. Fix to come.

pmcb55 (approved):
LGTM

PR #482, opened 2020-10-26
Workaround react components issues with localstorage

NSeydoux:
This is a stopgap solution, until a cleaner fix is found. For some reason, the functions used to clear the local storage that are working in the demo app aren't working for the react components, which is a problem for podbrowser. This enforces a clean slate, preventing previous client ids to be reused.


ajacksified (approved):
This fix should be necessary for all environments, as by default the clearStaleState function only clears keys > 900s old.

PR #483, opened 2020-10-26
Feat/no resolve dependency

NSeydoux:
Building a session no longer requires calling `buildClientAuthnWithDependencies`: the call is done internally by the Session constructor. Additionnally, tests for the Session class are created to reach 100% code coverage.

While I was at it, I also fixed a long-standing bug: the `login` event could never be emitted, since the user got redirected away from the page as part of the login method. Now, the `login` event is fired by `handleIncomingRedirect`, if and only if a session is logged in.

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -153,11 +153,38 @@ describe("Session", () => {
           done();
         }
       };
-      const mySession = new Session({
-        clientAuthentication: mockClientAuthentication(),
-      });
+      const clientAuthentication = mockClientAuthentication();
+      clientAuthentication.handleIncomingRedirect = jest.fn(
+        async (_url: string) => {
+          return {
+            isLoggedIn: true,
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
       mySession.onLogin(myCallback);
-      await mySession.login({});
+      await mySession.handleIncomingRedirect("https://some.url");
+    });
+
+    it("does not call the registered callback if login isn't successful", async () => {
+      const failCallback = (): void => {
+        fail();
+      };
+      const clientAuthentication = mockClientAuthentication();
+      clientAuthentication.handleIncomingRedirect = jest.fn(
+        async (_url: string) => {
+          return {
+            isLoggedIn: false,
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      mySession.onLogin(failCallback);
+      await mySession.handleIncomingRedirect("https://some.url");
     });
   });
 
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2020 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// The following is required by tsyringe
+import "reflect-metadata";
+import { it, describe } from "@jest/globals";
+import { mockClientAuthentication } from "../src/__mocks__/ClientAuthentication";
+import { Session } from "../src/Session";
+import { mockStorage } from "../../core/src/storage/__mocks__/StorageUtility";
+
+describe("Session", () => {
+  describe("constructor", () => {
+    it("accepts an empty config", async () => {
+      const mySession = new Session({});
+      expect(mySession.info.isLoggedIn).toEqual(false);
+      expect(mySession.info.sessionId).toBeDefined();
+    });
+
+    it("accepts no config", async () => {
+      const mySession = new Session();
+      expect(mySession.info.isLoggedIn).toEqual(false);
+      expect(mySession.info.sessionId).toBeDefined();
+    });
+
+    it("does not generate a session id if one is provided", () => {
+      const mySession = new Session({}, "mySession");
+      expect(mySession.info.sessionId).toEqual("mySession");
+    });
+
+    it("accepts input storage", async () => {
+      const insecureStorage = mockStorage({});
+      const secureStorage = mockStorage({});
+      const mySession = new Session({
+        insecureStorage,
+        secureStorage,
+      });
+      const clearSecureStorage = jest.spyOn(secureStorage, "delete");
+      const clearInsecureStorage = jest.spyOn(insecureStorage, "delete");
+      await mySession.logout();
+      expect(clearSecureStorage).toHaveBeenCalled();
+      expect(clearInsecureStorage).toHaveBeenCalled();
+    });
+
+    it("accepts session info", () => {
+      const mySession = new Session({
+        sessionInfo: {
+          sessionId: "mySession",
+          isLoggedIn: false,
+          webId: "https://some.webid",
+        },
+      });
+      expect(mySession.info.isLoggedIn).toEqual(false);
+      expect(mySession.info.sessionId).toEqual("mySession");
+      expect(mySession.info.webId).toEqual("https://some.webid");
+    });
+  });
+
+  describe("login", () => {
+    it("wraps up ClientAuthentication login", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnLogin = jest.spyOn(clientAuthentication, "login");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.login({});
+      expect(clientAuthnLogin).toHaveBeenCalled();
+    });
+  });
+
+  describe("logout", () => {
+    it("wraps up ClientAuthentication logout", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnLogout = jest.spyOn(clientAuthentication, "logout");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.logout();
+      expect(clientAuthnLogout).toHaveBeenCalled();
+    });
+  });
+
+  describe("fetch", () => {
+    it("wraps up ClientAuthentication fetch", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnFetch = jest.spyOn(clientAuthentication, "fetch");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.fetch("https://some.url");
+      expect(clientAuthnFetch).toHaveBeenCalled();
+    });
+  });
+
+  describe("handleincomingRedirect", () => {
+    it("wraps up ClientAuthentication handleIncomingRedirect", async () => {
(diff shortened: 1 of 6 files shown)


packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      expect(mySession.info.sessionId).toBeDefined();
+    });
+
+    it("accepts no config", async () => {
+      const mySession = new Session();
+      expect(mySession.info.isLoggedIn).toEqual(false);
+      expect(mySession.info.sessionId).toBeDefined();
+    });
+
+    it("does not generate a session id if one is provided", () => {
```
pmcb55, line 43: In comments and messages, ID should be capitalized:
```suggestion
    it("does not generate a session ID if one is provided", () => {
```

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      const insecureStorage = mockStorage({});
+      const secureStorage = mockStorage({});
+      const mySession = new Session({
+        insecureStorage,
+        secureStorage,
+      });
+      const clearSecureStorage = jest.spyOn(secureStorage, "delete");
+      const clearInsecureStorage = jest.spyOn(insecureStorage, "delete");
+      await mySession.logout();
+      expect(clearSecureStorage).toHaveBeenCalled();
```
pmcb55, line 58: I don't like these mock tests that are dependent on implementation details (i.e. that specific internal methods are being called). The name of test ("accepts input storage") seems to imply this test is just testing that the those options are 'acceptable', so what's that got to do with logging out and clearing storage? If the test is testing that the Session does in fact **_use_** the storage instances being passed into the constructor, then shouldn't it put something in that storage, and then assert that it's actually there (and that it's then 'gone' after a logout)? But that sounds like separate tests to me, which you probably have below, so I'm not sure I'd have a specific test at all for any classes constructor that it accepts certain input params, unless that constructor was performing some action on those params (i.e. what 'behaviour' is this test testing...?).

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    });
+  });
+
+  describe("login", () => {
+    it("wraps up ClientAuthentication login", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnLogin = jest.spyOn(clientAuthentication, "login");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.login({});
+      expect(clientAuthnLogin).toHaveBeenCalled();
```
pmcb55, line 82: Again, I'd expect the mocked session `login()` method to set the `isLoggedIn` flag to true, so shouldn't this test be asserting on that flag, rather than the fact that some method on an internal instance (i.e. `clientAuthentication`'s `login` method was invoked)?

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    });
+  });
+
+  describe("logout", () => {
+    it("wraps up ClientAuthentication logout", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnLogout = jest.spyOn(clientAuthentication, "logout");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.logout();
+      expect(clientAuthnLogout).toHaveBeenCalled();
```
pmcb55, line 92: Ditto

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    });
+  });
+
+  describe("fetch", () => {
+    it("wraps up ClientAuthentication fetch", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnFetch = jest.spyOn(clientAuthentication, "fetch");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.fetch("https://some.url");
+      expect(clientAuthnFetch).toHaveBeenCalled();
```
pmcb55, line 102: Ditto

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  describe("handleincomingRedirect", () => {
+    it("wraps up ClientAuthentication handleIncomingRedirect", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnHandle = jest.spyOn(
+        clientAuthentication,
+        "handleIncomingRedirect"
+      );
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.url");
+      expect(clientAuthnHandle).toHaveBeenCalled();
```
pmcb55, line 115: Ditto - i.e. is there specific behaviour (from our mocked `clientAuthentication`) that we can assert on in the Session instance here instead of expecting a specific method to have been called?

packages/browser/__tests__/Session.spec.ts
NSeydoux: My thinking was the following: for testing purpose, one way to use the constructor (namely passing in a ClientAuthentication) is much more convenient, so the other ways of using the constructor (e.g. here passing the `storage`) wouldn't be covered if the constructor wasn't tested independently. Here, more than testing that the specific methods are called, what I want to test is that these methods are called **on the objects passed to the constructor**, which verifies that the input parameters are properly used.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+            isLoggedIn: true,
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      mySession.onLogin(myCallback);
+      await mySession.handleIncomingRedirect("https://some.url");
+    });
```
pmcb55, line 169: No assertion here - could you just set a simple boolean flag in the `myCallback` function and then assert that it's set here...?

packages/browser/__tests__/Session.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+            isLoggedIn: false,
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      mySession.onLogin(failCallback);
+      await mySession.handleIncomingRedirect("https://some.url");
+    });
```
pmcb55, line 188: Yeah, again just setting a simple boolean in the callback, and then asserting that it's **_not_** set here might be clearer...?

packages/browser/src/Session.ts
```diff
@@ -157,8 +154,12 @@ export class Session extends EventEmitter {
       url
     );
     if (sessionInfo) {
+      if (sessionInfo.isLoggedIn) {
```
pmcb55, line 157: Maybe just a comment to explain why this event is being fired here (i.e. due to the redirect)?

packages/browser/__tests__/Session.spec.ts
NSeydoux: `login` doesn't set the `isLoggedIn` flag: that's done by `handleincomingRedirect`. Basically, `Session` is a thin layer on top of `ClientAuthentication`, which is the core class here. Moreover, when calling `login` on a `ClientAuthentication` object, depending on the parameters, different behaviours are expected. Should all the possible behaviours be tested at this layer (e.g. the different OIDC redirect requests), knowing that these tests will be very redundant with the unit tests of the actual classes handling each of these flows (e.g. `AuthCodeWithPkceLoginHandler`?

packages/browser/src/dependencies.ts
```diff
@@ -205,6 +205,11 @@ container.register<ILogoutHandler>("logoutHandler", {
   useClass: GeneralLogoutHandler,
 });
 
+/**
+ *
+ * @param dependencies
+ * @deprecated This function will be removed from the external API in an upcomiong release.
```
pmcb55, line 211:
```suggestion
 * @deprecated This function will be removed from the external API in an upcoming release.
```

packages/browser/__tests__/Session.spec.ts
NSeydoux: The assertion is that the `done()` function gets called: if not, the test will time out.

pmcb55 (approved):
LGTM - comments here are just nit-picky stuff really, this looks really great!

packages/browser/__tests__/Session.spec.ts
pmcb55: Ah ok. Well, if this is just a thin wrapper, and all the behaviour is actually in a wrapped class, then I'd just change this test to:
```
expect(await mySession.login({}).not.toThrow();
```
...given it returns (a promise) on `void`. If the method under test returned anything at all I'd probably do:
```
expect(await mySession.login({}).not.isNull();
```
...or something.
So yeah, I agree not to test the expected flows, since as you say, they'll all be tested elsewhere, but I don't think this test should be aware of the internal details of the wrapper class (i.e. that it's calling a specific method on the wrapped class) - the only behaviour of the wrapper to test is that it doesn't fail (since it doesn't have any **_behaviour_** of it's own to test really).

packages/browser/__tests__/Session.spec.ts
pmcb55: Yeah, sure - but it's the behaviour on **_those objects passed to the constructor_** we want to test, and not the internal methods being used to enact that behaviour. So yeah, in this case I would try and store something in those mocked storage instances, then logout and then check that that storage has been cleared out (that might require some implementation knowledge, e.g. that a specific key was used to store data in that storage, but that key would be part of the documentation (i.e. our SDK uses the key 'XYZ' to store information in secure storage), so that would be Ok for this test to know about too, as it isn't a hidden internal implementation detail).

packages/browser/__tests__/Session.spec.ts
NSeydoux: Maybe it's being pedantic, but isn't a test just checking if something doesn't fail, as opposed to testing for actual output or for internal behaviour, a code smell ? As in that's just going to bump the code coverage, but it won't provide any value, and even worse in the case of a wrapper class like here it could hide the fact that the wrapped class isn't called at all (even if that's unlikely). What do you think ?

packages/browser/__tests__/Session.spec.ts
pmcb55: Yeah I know, I got that. But would a simple boolean flag just be clearer (and also I'm kinda surprised that our linter doesn't complain about the lack of an explicit assertion in this test (or is it just clever enough to recognise the use of `done()`)...?

packages/browser/src/Session.ts
NSeydoux: Fixed in 7c074ba3c132fef316c992aa33d11445ca311e90

packages/browser/__tests__/Session.spec.ts
NSeydoux: If the purpose of the test is to check that the callback isn't called, isn't it even clearer to have the callback simpley `fail()` the test ?

packages/browser/__tests__/Session.spec.ts
pmcb55: > a code smell ?
No, I don't think so. Yes, it might be needed to get coverage (although maybe not in this case, as other tests will exercise the `login` method), but I do think it's still valid to test that a wrapper method doesn't fail when invoked with correct the params. I generally wouldn't **_also_** test for failure modes if I think the wrapped class itself has full coverage (and if the wrapper class really is **_just_** a wrapper (i.e. that it doesn't catch exceptions from the wrapped class and add it's own extra context to that exception for instance - but of course, if it **_did_** do that, then it **_would_** be adding behaviour, and therefore would absolutely have to be tested for here too)).

packages/browser/__tests__/Session.spec.ts
pmcb55: Yeah, arguably (so don't change if you prefer your approach). But personally I'd write this as:
```
it("does not call the registered callback if login isn't successful", async () => {
      const failCallback = (): void => {
        throw new Error("Should *NOT* call callback - this means test has failed!");
      };
      const clientAuthentication = mockClientAuthentication();
      clientAuthentication.handleIncomingRedirect = jest.fn(
        async (_url: string) => {
          return {
            isLoggedIn: false,
            sessionId: "a session ID",
            webId: "https://some.webid#them",
          };
        }
      );
      const mySession = new Session({ clientAuthentication });
      mySession.onLogin(failCallback);
      expect(() => await mySession.handleIncomingRedirect("https://some.url")).not.toThrow();
    });
```

packages/browser/__tests__/Session.spec.ts
NSeydoux: Yeah, I was also quite surprised that the linter didn't complain. Looking into it, it seems like there's an issue with the linter config: it just doesn't detect empty tests. I installed locally `eslint-plugin-jest`, but that didn't make a difference. I'll investigate as part of a separate issue.

packages/browser/__tests__/Session.spec.ts
NSeydoux: Actually, I thought that this could fail to capture the case where the callback is called after the promise resolves, but I could not produce such outcome, so I changed to your suggestion in 92e8ae0b14dc6a0577a72838a2cb4bf306fffe85.

packages/browser/__tests__/Session.spec.ts
NSeydoux: Added an explicit expect in 92e8ae0b14dc6a0577a72838a2cb4bf306fffe85

PR #493, opened 2020-10-27
Gracefully handle redirect for DPoP

### LINKED ISSUE #89: Does not handle redirects well
Upon redirect, Auth Fetcher does not create a new DPoP Token

NSeydoux:
Since a DPoP header is scoped to one target IRI, if the request is redirected, the DPoP header is invalid, which results in a 401 unauthenticated. To fix the issue, the fetch must capture the redirection, and replay the request to the actual target IRI.

This is particularly useful when requests target a container, but omit to include the trailing `/`.

Fixes #89.

# Checklist

### DIFF
--- a/packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
+++ b/packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
@@ -112,7 +112,7 @@ describe("buildDpopFetch", () => {
   it("builds the appropriate DPoP header for a given HTTP verb.", async () => {
     mockFetch(mockNotRedirectedResponse());
     const key = await generateJwkForDpop();
-    const myFetch = await buildDpopFetch("myToken", undefined,  key);
+    const myFetch = await buildDpopFetch("myToken", undefined, key);
     await myFetch("http://some.url", {
       method: "POST",
     });
@@ -222,7 +222,7 @@ describe("buildDpopFetch", () => {
     );
 
     const key = await generateJwkForDpop();
-    const myFetch = await buildDpopFetch("myToken", key);
+    const myFetch = await buildDpopFetch("myToken", undefined, key);
     const response = await myFetch("https://my.pod/container");
 
     expect(
@@ -245,7 +245,7 @@ describe("buildDpopFetch", () => {
     );
 
     const key = await generateJwkForDpop();
-    const myFetch = await buildDpopFetch("myToken", key);
+    const myFetch = await buildDpopFetch("myToken", undefined, key);
     const response = await myFetch("https://my.pod/resource");
 
     expect(
--- a/packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
+++ b/packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
@@ -27,15 +27,45 @@ import {
 } from "../../src/authenticatedFetch/fetchFactory";
 import { decodeJwt, generateJwkForDpop } from "@inrupt/oidc-client-ext";
 
+type MockedRedirectResponse = {
+  redirected: boolean;
+  url: string;
+};
+
+const mockRedirectedResponse = (): MockedRedirectResponse => {
+  return {
+    redirected: true,
+    url: "https://my.pod/container/",
+  };
+};
+
+const mockNotRedirectedResponse = (): MockedRedirectResponse => {
+  return {
+    redirected: false,
+    url: "https://my.pod/container/",
+  };
+};
+
+const mockFetch = (
+  response: MockedRedirectResponse
+): typeof window.fetch => {
+  window.fetch = jest.fn().mockReturnValueOnce(
+    new Promise((resolve) => {
+      resolve(response as Response);
+    })
+  ) as jest.Mock<
+      ReturnType<typeof window.fetch>,
+      [RequestInfo, RequestInit?]
+    >;
+    return window.fetch;
+};
+
 // We use ts-ignore comments here only to access mock call arguments
 /* eslint-disable @typescript-eslint/ban-ts-ignore */
 
 describe("buildBearerFetch", () => {
   it("returns a fetch holding the provided token", async () => {
-    window.fetch = jest.fn() as jest.Mock<
-      ReturnType<typeof window.fetch>,
-      [RequestInfo, RequestInit?]
-    >;
+    mockFetch(mockNotRedirectedResponse());
     const myFetch = buildBearerFetch("myToken", undefined);
     await myFetch("someUrl");
 
@@ -46,10 +76,7 @@ describe("buildBearerFetch", () => {
   });
 
   it("returns a fetch preserving the optional headers", async () => {
-    window.fetch = jest.fn() as jest.Mock<
-      ReturnType<typeof window.fetch>,
-      [RequestInfo, RequestInit?]
-    >;
+    mockFetch(mockNotRedirectedResponse());
     const myFetch = buildBearerFetch("myToken", undefined);
     await myFetch("someUrl", { headers: { someHeader: "SomeValue" } });
 
@@ -65,10 +92,7 @@ describe("buildBearerFetch", () => {
   });
 
   it("returns a fetch overriding any pre-existing authorization headers", async () => {
-    window.fetch = jest.fn() as jest.Mock<
-      ReturnType<typeof window.fetch>,
-      [RequestInfo, RequestInit?]
-    >;
+    mockFetch(mockNotRedirectedResponse());
     const myFetch = buildBearerFetch("myToken", undefined);
     await myFetch("someUrl", { headers: { Authorization: "some token" } });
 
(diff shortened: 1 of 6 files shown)


NSeydoux:
Note that unlike it was initially planned for, this does not throw on redirect: the DPoP token is rebuilt, and the request re-sent to the target IRI.

packages/browser/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -53,17 +73,17 @@ export async function buildDpopFetch(
...
-          dpopKey
-        ),
-      },
-    });
+    const response = await fetch(
+      init,
+      await buildDpopFetchOptions(init.toString(), authToken, dpopKey, options)
+    );
+    if (response.url === init.toString()) {
+      console.log(JSON.stringify(response, null, "  "));
```
Vinnl, line 81: I presume this was just for debugging?

```suggestion
```

packages/browser/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -42,6 +42,26 @@ export function buildBearerFetch(authToken: string): typeof fetch {
...
 }
 
+async function buildDpopFetchOptions(
+  targetUrl: string,
+  authToken: string,
+  dpopKey: JSONWebKey,
+  defaultOptions?: RequestInit
+): Promise<RequestInit> {
+  return {
+    ...defaultOptions,
```
Vinnl, line 52: So how does this work? I'd have expected there to be a `redirect: 'manual'` in here somehow, but this looks like a redirect returns a successful response with a different `.url` property?

packages/browser/__tests__/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -32,17 +32,54 @@ import {
...
 
 describe("buildBearerFetch", () => {
   it("returns a fetch holding the provided token", async () => {
-    const fetch = jest.requireMock("cross-fetch") as {
-      fetch: jest.Mock<
-        ReturnType<typeof window.fetch>,
-        [RequestInfo, RequestInit?]
-      >;
-    };
+    // const fetch = mockFetch();
```
Vinnl, line 81:
```suggestion

```

Vinnl (commented):
I'm not quite following how this works yet, so I left a question inline.

Also, it seems like this might be a good candidate to add and end-to-end test for? Since it's so particular to server behaviour.

packages/browser/src/authenticatedFetch/fetchFactory.ts
NSeydoux: I initially included `redirect: "manual"`, but that actually did not make a difference in the implementation, because the `redirected` property of the Response wasn't set how I expected it to be. So it's as you say: `fetch` follows the redirects "normally", and eventually returns a Response. You don't have access to the intermediate redirected responses, so from the client's point of view you'll never see a 301 status code for instance, but the final response does have an `.url` property that makes it possible to detect the actual IRI of the target resource. From there, we can just re-issue the request.

NSeydoux:
Good point, an end-to-end test would be a nice addition to this feature. i'll add one to the current PR.

packages/browser/src/authenticatedFetch/fetchFactory.ts
Vinnl: Hmm OK, I suppose technically that would work, but I'm still not 100% comfortable with it. I suppose this does mean that the target of the redirect will receive two requests rather than one, with the first one failing?

I did some tests locally, and although it seems like `redirected` is always `false` if you set `redirect` to `"manual"`, but the `type` of the Response does indicate whether you need to do a redirect or not, i.e. it will be `"opaqueredirect"` if you do.

To reproduce the tests, I created an empty dir, ran `npm install express`, then created the following `index.js`:

```js
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  res.sendStatus(200);
})

app.get('/redirectfail', (req, res) => {
  res.redirect(301, "http://localhost:3000/fail")
})

app.get('/fail', (req, res) => {
  res.sendStatus(403);
})

app.get('/redirectsuccess', (req, res) => {
  res.redirect(301, "http://localhost:3000/success")
})

app.get('/success', (req, res) => {
  res.send("Hoi");
})

app.listen(port, () => {
  console.log(`Example app listening at http://localhost:${port}`)
})
```

You can then see what happens in the different scenarios using your browser console:

1a:
```
await fetch("http://localhost:3000/success", { redirect: "manual" });
// redirected: false
// type: "basic"
// url: "http://localhost:3000/success"
```


1b:
```
await fetch("http://localhost:3000/redirectsuccess", { redirect: "manual" });
// redirected: false
// type: "opaqueredirect"
// url: "http://localhost:3000/redirectsuccess"
```

1c:
```
await fetch("http://localhost:3000/redirectfail", { redirect: "manual" });
// redirected: false
// type: "opaqueredirect"
// url: "http://localhost:3000/redirectfail"
```

2a:
```
await fetch("http://localhost:3000/success");
// redirected: false
// type: "basic"
// url: "http://localhost:3000/success"
```

2b:
```
await fetch("http://localhost:3000/redirectsuccess");
// redirected: true
// type: "basic"
// url: "http://localhost:3000/success"
```

2c:
```
await fetch("http://localhost:3000/redirectfail");
// redirected: true
// type: "basic"
// url: "http://localhost:3000/fail"
```

Essentially, what we do is distinguish b and c (i.e. redirects) from a. Without `redirect: "manual" (i.e. the scenarios labeled 2)`, indeed the fact that the `url` on the Response differs from the requested URL distinguishes them.

However, when the target is not accessible (i.e. 2c) the first redirect call to `fetch` already logs an error, and then the DPoP retry will log another one. This does not happen when `redirect: "manual"` is set (1c).

Then we'd need to distinguish 1b and 1c (redirects) from 1a (a regular request). That, apparently, can be done by checking for `response.type === "opaqueredirect"`.

NSeydoux: Thanks for the explanation, that makes it very clear ! 

I have a concern though: in the case where we don't set `redirect: manual`, the browser handles the redirections. In the case we handle them manually, we have to manage the case where multiple redirections happen in a row (`http://movedOnce.net/resource` -> `http://movedTwice.net/resource` -> `http://target.com/resource`). However, what happens if there is a cycle ? Should we have a maximum number of jumps ? If you add 
```
app.get('/redirectloop', (req, res) => {
  res.redirect(301, "http://localhost:3000/redirectloop")
})
```
to the previous server, it looks like without `redirect: manual`, the browser throws an error after 21 requests, so 20 redirections. If that's good enough for Firefox, that's good enough for me :p

NSeydoux: Proposed solution: 85dc450819d9cad8fa2959016a2b57061de0a4ec

NSeydoux: Wait, I just realised there was is issue: when using `redirect: manual`, you don't have the IRI to which you should be redirected, `response.url` isn't updated to show the new target, so there's no way to know what the request should be redirected to. I'll change the code back to how it was before, but I'll see if I can be smartest with the server's response in order not to replay a request that is failing for non-auth related issues.

ajacksified: https://github.com/whatwg/fetch/issues/763 🤦‍♂️

NSeydoux: Fixed in 5a9b9cd22fce6f79e5e5a9cb59ed272e6df7bd2f

pmcb55 (approved):
This does look a bit shaky, but I'll approve to get PB moving (I assume test coverage is good...!?)

NSeydoux:
Coverage is 100% for the code impacted by this PR, and an e2e test is also added to verify that fetching `https://my.pod/private` ends up fetching `https://my.pod/private/` successfully. The bits added in the fetch basically aim at not re-issuing any unnecessary request, meaning the initial request hadn't been redirected and had an auth issue, or it had a non-auth issue. `401` and `403` are the only errors that I can see making sense in this context: a request rejected because the provided credentials aren't accepted by the server because of the redirection. Actually typing this, it seems that only 403 makes sense: you are authenticated, but the credentials aren't accepted. If you aren't providing any authentication, retrying the request won't help.

NSeydoux:
Ah yeah right, after a quick test I verified that NSS does return 401 for an authenticated request that gets rejected for DPoP mismatch reasons. I'll log a bug ticket in NSS, and add a link in the code.

packages/browser/src/authenticatedFetch/fetchFactory.ts
Vinnl: Dammit. OK, well, if the e2e test succeeds, that should give enough confidence then, I suppose. Sorry for the extra work!

PR #499, opened 2020-10-28
Simplified the sample app

NSeydoux:
This updates the single/bundle app in order to improve React practices. It also adds a logout button.

This sample app has a different purpose than the demoClient app: the latter has a richer UI and more features, while the former basically provides code snippets serving as documentation.

# Checklist

### DIFF
--- a/packages/browser/examples/single/bundle/package.json
+++ b/packages/browser/examples/single/bundle/package.json
@@ -27,7 +27,6 @@
   "dependencies": {
     "react": "^16.13.0",
     "react-dom": "^16.13.0",
-    "regenerator-runtime": "^0.13.3",
-    "@inrupt/solid-client-authn-browser": "file://../../../"
+    "regenerator-runtime": "^0.13.3"
   }
 }
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -19,134 +19,92 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import React, { Component } from "react";
+import React, { useState, useEffect } from "react";
 import "regenerator-runtime/runtime";
 
-import {
-  Session,
-  getClientAuthenticationWithDependencies,
-} from "../../../../dist/index";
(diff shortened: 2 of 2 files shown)


matthieubosquet (approved):
Currently broken because of [@inrupt/oidc-client-ext](https://www.npmjs.com/package/@inrupt/oidc-client-ext)'s version (0.2.3-chorerename-oidc-module-331577249-1609.0), but once it's fixed, all works like a charm.
LGTM.

NSeydoux:
I'm not sure what issue you are experiencing with oidc-client-ext. If you run locally `npm ci` and then `npm run dev`, things don't work ?

matthieubosquet:
@NSeydoux exactly that. If you follow the link as per my comment: https://www.npmjs.com/package/@inrupt/oidc-client-ext
You'll see that the current version tags are pretty weird (for example: 0.2.3-chorerename-oidc-module-331577249-1609.0).
I think that prevents npm from installing @inrupt/oidc-client-ext.
It's to do with deploying that package to npm with a proper version tag. Haven't looked into it.
The error you get from `npm ci` is pretty standard and explicit:
```bash
npm ERR! Missing: @inrupt/oidc-client-ext@^0.2.2
```

NSeydoux:
These version numbers are pre-release tags, they shouldn't prevent from installing the package: they are generated by the CD for each git branch, so that we can test the impact of the changes in one branch before merging/releasing it. By any chance, do the change in https://github.com/inrupt/solid-client-authn-js/pull/503 solve your issue ? The demo app doesn't import the packages from NPMJS, but it resolves the dependencies locally. Did you run `npm ci` at the top level to build all the packages ?

matthieubosquet:
That makes sense!
I'm not familiar with how lerna works, so didn't know it symlinks "lerna packages", which oidc-client-ext is since it's matched by `"packages": ["packages/*"]`, instead of fetching them from npmjs.
Running `npm ci` at the top level works perfectly. Running `npm ci` on the `package/browser` fails because `https://registry.npmjs.org/@inrupt%2foidc-client-ext`^0.2.2 doesn't exist yet.
Thanks for the explanation.

PR #503, opened 2020-10-29
Fix module name in oidc package-lock

NSeydoux:


Vinnl (approved):
Hmm, this one is also weird (like #506) - the browser lockfile _adds_ all the `dev` properties. What did you do here?

(Or are these the conflicts with `master`?)

NSeydoux:
I have no idea what went wrong. I noticed that the package-lock contained references to `localhost:4872`, which matches the default port used by Verdaccio. To have a clean start, I removed the package-lock, and used `lerna bootstrap` to recreate it properly. I hope that will be the end of it ^^

PR #504, opened 2020-10-29
Enable Static Application Security Testing

Vinnl:
See https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/enabling-code-scanning-for-a-repository.


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python' ]
```
NSeydoux, line 27: Is supporting python required here for the API documentation ?

NSeydoux (approved):

.github/workflows/codeql-analysis.yml
Vinnl: I thought so, although I didn't think of it when I enabled SAST on solid-client. If it doesn't help there's no harm in including it either, I think.

Vinnl:
Btw @NSeydoux, would you mind taking a look at the reported alerts? https://github.com/inrupt/solid-client-authn-js/security/code-scanning?query=ref%3Arefs%2Fheads%2Fenable-sast

I think they can mostly be dismissed (the use of `"some token"` in tests, for example), but I trust you more than myself on those :)

PR #505, opened 2020-10-29
Set up CD to delete preview tags

NSeydoux:


.github/workflows/cd-teardown.yml
```diff
@@ -0,0 +1,33 @@
+name: CD
+
+on: delete
+
+env:
+  CI: true
+jobs:
+  unpublish-npm:
+    runs-on: ubuntu-18.04
```
Vinnl, line 9: Might as well use the latest version if we're setting this up anew :)

```suggestion
    runs-on: ubuntu-20.04
```

Vinnl (approved):
Looks good, only worry I have is how we can make sure that we don't forget adding new packages (e.g. `-node`). Unfortunately I don't have any ideas on how to do that, but hopefully me just mentioning this will trigger a brilliant insight on your side ;)

NSeydoux:
Unfortunately I have the same concern and no satisfactory solution ^^. I initially hoped there would be a lerna command to avoid the manual aspect of this, but that is not the case. I guess one of us will have to remind the other about this when this becomes relevant :p

Vinnl:
Deal :)

PR #506, opened 2020-10-29
Removed dependency on cross-fetch

### LINKED ISSUE #180: Use Webpack to not bundle `cross-fetch`
Considering the conversation in #174, in the browser environment `cross-fetch` won't be required. It should therefore not be included by the bundler to reduce the bundle size.

NSeydoux:
Fixes #180 .

Since the package does not intend to be isomorphic, it can make the assumption that the fetch should default to the window's.
# Checklist

### DIFF
--- a/packages/browser/package-lock.json
+++ b/packages/browser/package-lock.json
@@ -14,48 +14,155 @@
       }
     },
     "@babel/core": {
-      "version": "7.12.3",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.3.tgz",
-      "integrity": "sha512-0qXcZYKZp3/6N2jKYVxZv0aNCsxTSVCiK72DTiTYZAu7sjg73W0/aynWjMbiGd87EQL4WyA8reiJVh92AVla9g==",
+      "version": "7.11.5",
+      "resolved": "http://localhost:4873/@babel%2fcore/-/core-7.11.5.tgz",
+      "integrity": "sha512-fsEANVOcZHzrsV6dMVWqpSeXClq3lNbYrfFGme6DE25FQWe7pyeYpXyx9guqUnpy466JLzZ8z4uwSr2iv60V5Q==",
       "dev": true,
       "requires": {
         "@babel/code-frame": "^7.10.4",
-        "@babel/generator": "^7.12.1",
-        "@babel/helper-module-transforms": "^7.12.1",
-        "@babel/helpers": "^7.12.1",
-        "@babel/parser": "^7.12.3",
+        "@babel/generator": "^7.11.5",
+        "@babel/helper-module-transforms": "^7.11.0",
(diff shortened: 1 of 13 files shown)


packages/browser/package-lock.json
```diff
@@ -8,7 +8,6 @@
       "version": "7.10.4",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
       "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
-      "dev": true,
```
Vinnl, line 11: I'm a bit surprised at there being a whole lot of these lines in here. I'm not sure exactly what they mean, but I suppose it might be something like a devDependency suddenly being moved into regular dependencies? Or possible using a different version of npm? Either, just wanted to ask for a double-check that this is still fine.

(I suppose if `npm ci` doesn't fail it should be fine.)

packages/browser/src/util/Fetcher.ts
```diff
@@ -27,20 +27,18 @@
...
 /**
  * @hidden
  */
 export default class Fetcher implements IFetcher {
   async fetch(url: RequestInfo, init?: RequestInit): Promise<Response> {
-    if (typeof window !== "undefined" && typeof window.fetch !== "undefined") {
-      return window.fetch(url, init);
-    }
-    return _fetch(url, init);
+    return global.fetch(url, init);
```
Vinnl, line 42: `global` is a Node-specific global, IIRC. Either `window` or `globalThis` (no IE support, but :shrug: ) would be more appropriate, I think: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

Vinnl (approved):
I'm wondering if we need to add to the review that this package might need polyfills (like for `fetch`) on Internet Explorer, like we do on solid-client? (Also, the README currently mentions a required Node version, which doesn't seem like it should apply other than to contributors...)

packages/browser/package-lock.json
NSeydoux: Yeah there's something wrong going on with the package-lock... I think the root cause is something that wasn't correctly commited when cutting the release. My intuitiion is that https://github.com/inrupt/solid-client-authn-js/pull/503 would solve it, could you have a look/ approve it so that I can rebase this on it ?

packages/browser/src/util/Fetcher.ts
NSeydoux: Good point, that's what I've been using everywhere else I don't know what I made an exception here. Fixed in b3fc9e4ce68dc0264b70f5ee23e959eb76f03d07

PR #526, opened 2020-11-02
Added README for sample apps

NSeydoux:


packages/browser/examples/single/script/README.md
```diff
@@ -0,0 +1,11 @@
+# `solid-client-authn-browser` sample app
+
+This application provides a snippet of code showcasing usage of the 
+`solid-client-authn-browser` library in React components, with a bundler.
```
Vinnl, line 4:
```suggestion
`solid-client-authn-browser` library in React components, without a bundler.
```

Am I correct in guessing that this is the difference between the two projects?

Vinnl (approved):

packages/browser/examples/single/script/README.md
NSeydoux: Absolutely. Actually, it's not even using React: fixed in e847a26893b0166bb047c03d361e001d808bba01

PR #528, opened 2020-11-03
remove url-parse completely, updating package-lock.json

pmcb55:
#Internal refactor

Removed `url-parse` from the codebase.


packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -106,8 +105,9 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
         `AuthCodeRedirectHandler cannot handle [${redirectUrl}]: it is missing one of [code, state].`
       );
     }
-    const url = new URLParse(redirectUrl, true);
-    const oauthState = url.query.state as string;
+
+    const url = new URL(redirectUrl);
+    const oauthState = url.searchParams.get("state") as string;
```
NSeydoux, line 110: Here, as in `OidcLoginHandler`, I'd rather use proper type guards to make the check for the property explicit, rather than use an equivalent type assertion.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -147,7 +147,7 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
         this.issuerConfigFetcher,
         this.clientRegistrar,
         // the canHandle function checks that the code is part of the query strings
-        url.query["code"] as string,
+        url.searchParams.get("code") as string,
```
NSeydoux, line 150: ditto

packages/browser/src/login/oidc/redirectHandler/ImplicitRedirectHandler.ts
```diff
@@ -67,18 +66,26 @@ export class ImplicitRedirectHandler implements IRedirectHandler {
...
+
+    const url = new URL(redirectUrl);
 
     await this.tokenSaver.saveSession(
-      url.query.state as string,
-      url.query.id_token as string,
-      url.query.access_token
+      url.searchParams.get("state") as string,
+      url.searchParams.get("id_token") as string,
+      url.searchParams.get("access_token") as string
```
NSeydoux, line 75: ditto

NSeydoux (approved):
I'll run manual tests to be sure about the change regarding IRIs, but the Pod root being public by default in NSS, I'm not sure any issue could be uncovered.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
pmcb55: Not possible to add type guards on a Map lookup.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
pmcb55: Not possible to add type guards on a Map lookup.

packages/browser/src/login/oidc/redirectHandler/ImplicitRedirectHandler.ts
pmcb55: Not possible to add type guards on a Map lookup.

PR #531, opened 2020-11-03
Hoisted devDependencies using `lerna link convert`

NSeydoux:
Working on #517 showed that updating webpack on the core module broke the build, while the core module should not have dependend on webpack in the first place. This hoists all the dev dependencies to the top module, where only one version of each dependency will be used. Incidently, it resolves the issue in #517.


NSeydoux:
Actually it looks like it creates a huge mess. Nevermind !

Vinnl:
Haha, this looks familiar: #320.

PR #587, opened 2020-11-12
Moved error-related code in the core module

NSeydoux:
There is no environment-specificness to this code, so it should be located in the core module to avoid duplication.


Vinnl (approved):
LGTM, but consider re-exporting the error classes from the `-browser` package as well, so developers can use it both as a type and in an `instanceof` check, if that is desirable.

NSeydoux:
Good point, fixed in 0ee5e78855517ae49c17f1dd0a02717408ec9a45.

PR #592, opened 2020-11-14
Core module uses cross-fetch

NSeydoux:
This aligns the type definitions of the core module with the new node module, while remaining compatible with the browser module. Also, this removes unused code.


packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
```diff
@@ -26,6 +26,7 @@
 
 import IHandleable from "../../../util/handlerPattern/IHandleable";
 import ISessionInfo from "../../../sessionInfo/ISessionInfo";
+import { fetch } from "cross-fetch";
```
pmcb55, line 29: It seems you're only importing this for the type information on Fetch, which seems a heavy dependency just to get type information into the Core package (which shouldn't have any browser or NodeJS awareness, ideally). I would have thought `fetch` was part of the web standards, and therefore should already be in `lib.dom.d.ts`, see this question: https://github.com/DefinitelyTyped/DefinitelyTyped#what-about-type-definitions-for-the-dom, which I got to from looking for just types for fetch (https://www.npmjs.com/package/@types/node-fetch). E.g. what about just importing this (which I think would just be a Dev Dependency, which would be much better than a Dependency): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-fetch/index.d.ts

pmcb55 (commented):
Lots of skipped tests in here still - are they due to be removed soon too, or are they still needed here for some reason...?

packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
Vinnl: You wouldn't want to enable `dom` in `-core` though, since the DOM is not available when `-core` is used in `-node`. Since the dependency doesn't affect the actual code that gets run or shipped to the client, I'd say it's fine to add it.

If you want to make sure that you do only import the types, you could turn this into a [type-only import](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export), I think:

```suggestion
import type { fetch } from "cross-fetch";
```

Vinnl (approved):

NSeydoux:
@pmcb55 it depends on the tests: one of them is skipped because of an NSS issue, some of theme were never implemented in the first place, and should be, and one of them should be removed soon.

PR #593, opened 2020-11-14
Feat/non oidc node

NSeydoux:
This ports the code which does not rely on an OIDC library from soilc-client-authn-browser. `window.fetch` is replaced by "cross-fetch", and the jest environment is node (not jsdom).

Note that the core of the code is missing, and will be implemented in (hopefully) smaller PRs using a dedicated node OIDC library. Currently, all the tests that would require such library just test that the code throws a "not implemented" error. Also, this requires #592. 

PS: Sorry for the über-PR.

### DIFF
--- a/packages/browser/package-lock.json
+++ b/packages/browser/package-lock.json
@@ -8,6 +8,7 @@
       "version": "7.10.4",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
       "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
+      "dev": true,
       "requires": {
         "@babel/highlight": "^7.10.4"
       }
@@ -16,6 +17,7 @@
       "version": "7.11.5",
       "resolved": "http://localhost:4873/@babel%2fcore/-/core-7.11.5.tgz",
       "integrity": "sha512-fsEANVOcZHzrsV6dMVWqpSeXClq3lNbYrfFGme6DE25FQWe7pyeYpXyx9guqUnpy466JLzZ8z4uwSr2iv60V5Q==",
+      "dev": true,
       "requires": {
         "@babel/code-frame": "^7.10.4",
         "@babel/generator": "^7.11.5",
@@ -39,6 +41,7 @@
           "version": "7.12.1",
           "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.12.1.tgz",
           "integrity": "sha512-DB+6rafIdc9o72Yc3/Ph5h+6hUjeOp66pF0naQBgUFFuPqzQwIlPTm3xZR7YNvduIMtkDIj2t21LSQwnbCrXvg==",
+          "dev": true,
           "requires": {
             "@babel/types": "^7.12.1",
             "jsesc": "^2.5.1",
@@ -48,7 +51,8 @@
             "source-map": {
               "version": "0.5.7",
               "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
-              "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w="
+              "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
+              "dev": true
             }
           }
         },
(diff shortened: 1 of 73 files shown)


packages/node/package.json
```diff
@@ -14,13 +14,12 @@
     "lint": "eslint \"src/**/*.ts\"",
     "lint-fix": "eslint --fix \"src/**\"",
     "licenses": "license-checker --production --out license.csv --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
-    "test": "npm run lint-fix && npm run licenses && jest --coverage --verbose --passWithNoTests",
+    "test": "npm run lint-fix && npm run licenses && jest --coverage --verbose --passWithNoTests --no-cache",
```
pmcb55, line 17: Was this just a temporary thing, or is there a genuine reason for turning off the cache (according to the Jest docs: "**_On average, disabling the cache makes Jest at least two times slower_**")...?

pmcb55 (approved):
LGTM - (looks like a wholesale copy-and-paste from the Browser codebase, as opposed to trying to start by just implementing, for example, `login()` and seeing what code might be shareable with the Browser code and perhaps refactored into the 'Core' module instead of being copied over directly...? But I guess this approach might be much quicker.)

packages/browser/package-lock.json
```diff
@@ -8,6 +8,7 @@
       "version": "7.10.4",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
       "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
+      "dev": true,
```
Vinnl, line 11: This is suspicious again - no changes to `package.json`, but lots of `"dev": true` added.

packages/node/package.json
```diff
@@ -30,7 +29,13 @@
     "typescript": "^4.0.5"
   },
   "dependencies": {
-    "@inrupt/solid-client-authn-core": "^1.0.0"
+    "@inrupt/solid-client-authn-core": "^1.0.0",
+    "@types/uuid": "^8.3.0",
+    "cross-fetch": "^3.0.6",
```
Vinnl, line 34: Since it only needs to work in Node, might as well just use `node-fetch` here?

Vinnl (commented):
Lots of warnings in the new code from GitHub Actions, quite a few of which are easily resolve I think - might want to consider resolving them now, before future warnings no longer jump out at you.

Since Pat already approved this and this is such an enormous PR, other than that I just had a high-level look at it, hope you don't mind.

packages/node/package.json
NSeydoux: That was my initial intent, but the type definitions of `node-fetch` are a bit weird (`fetch` is both a function and a namespace, and it has an `isRedirected` property absent of the canonical `fetch`), and were creating inconsistencies that disappear when using `cross-fetch`.

NSeydoux:
Good point @Vinnl regarding the GH actions warning, they were all due to one ESlint rule set to warning instead of error, which is fixed in 74126c37fd298a0c4fd0e1ef1c543feb91eb6ae9.

packages/browser/package-lock.json
NSeydoux: I'm really not sure why this keeps happening... I assume it is somehow related to Lerna, but I don't know what causes this.

packages/node/package.json
NSeydoux: Yup, that's an oversight, my mistake. Fixed in b556401d2d5c01986f1c6f1f2718b38827ea813b

PR #602, opened 2020-11-17
Feat/use error class

pmcb55:
# New feature description

Added a new `InruptError` class that supports simple error message strings, but also supports messages coming from RDF vocabs. All use-cases in Core package switched over, but just using simple strings (except for one in the Schema code that would just cause a merge conflict I think!).

Just a draft PR for now, as needs discussion before rolling out further with actual errors in RDF vocabs (it also pulls in N3, and just a Dev Dep on the Inrupt bundle of vocabs).

# Checklist


packages/core/src/errors/InruptError.ts
```diff
@@ -0,0 +1,116 @@
...
+import { VocabTerm } from "@inrupt/solid-common-vocab";
+
+export default class InruptError extends Error {
+  public readonly statusCode?: number;
+
+  public readonly statusText?: string;
+
+  constructor(
+    messageOrIri: string | NamedNode | VocabTerm,
+    messageParams?: string[],
```
NSeydoux, line 41: I'm not a fan of unnamed template parameters. I find it doesn't make the code easy to read, and doesn't provide context about how each parameter is going to be used. A `Record<string, string>` would be clearer, e.g.
```
const vocabError = UI_COMMON.errFileUpload_exceededSizeLimit;
const params = ["one", "two", "three"];
const error = new InruptError(vocabError, params);
```
would be easier to read as 
```
const vocabError = UI_COMMON.errFileUpload_exceededSizeLimit;
const params = {
  currentSize: "one",
  maxSize:  "two", 
  filename: "three"
};
const error = new InruptError(vocabError, params);
```

packages/core/src/errors/InruptError.ts
```diff
@@ -0,0 +1,116 @@
...
+    httpResponse?: Response & { ok: false },
+    appendHttpDetailsToMessage = true,
```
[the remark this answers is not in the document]
pmcb55: Yep, I agree. I don't like having multiple Error classes though (instead we can encode the 'type' of error, and any extra context-specific data within the error itself, using the extensibility of RDF to keep evolving that over time if needed). But in the next PR I'll rework this to remove the HTTP response from the ctor, and add it instead to an instance method, e.g. `new InruptError(iri, params).addHttpResponseDetails(response);`. That would mean the InruptError instance isn't immutable and fully initialised in the ctor anymore, but I think that's fine (and we could change it around to keep immutability too if we really wanted, e.g. by adding an optional 3rd string param like `new InruptError(iri, params, InruptError.extractHttpResponseDetails(response));`).

packages/core/src/errors/InruptError.ts
pmcb55: Yep, totally agree - this is really the result of the original versions of this idea being written in Java, where constructing Maps is very verbose and clunky. Using positional params work quite well in any programming language, and a big advantage is that the JavaDoc/TSDoc/etc. generated by the Artifact Generator already includes the English message, thereby allowing developers (who use IDEs!) to see tooltip descriptions of the error message, from which determining the meaning and order of params is relatively straightforward (although I agree it's not as nice as fully named params). We discussed the idea of generating TypeScript types that could explicitly list the required param names, one type per error message, which could provide really nice IDE support, but would still leave the Java side in trouble I think.

NSeydoux (approved):

PR #606, opened 2020-11-17
Dynamic client registration with openid-client

NSeydoux:
This implements DCR using "openid-client". With the code in this commit, the login process progresses all the way to the point where the client can redirect the user to the IdP.

Two main files are updated: 
- `OidcLoginHandler`, which depending on the options decides whether to go dynamically register the client or not
- `ClientRegistrar`, which handles the client registration.

### DIFF
--- a/packages/node/src/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/node/src/login/oidc/ClientRegistrar.spec.ts
@@ -48,7 +48,7 @@ describe("ClientRegistrar", () => {
   }
 
   describe("getClient", () => {
-    it("Fails if there is not registration endpoint", async () => {
+    it("fails if there is not registration endpoint", async () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
       const mockedIssuerConfig = mockIssuerConfig({
@@ -76,13 +76,14 @@ describe("ClientRegistrar", () => {
       );
     });
 
-    it("retrieves client id and secret from storage if they are present", async () => {
+    it("retrieves client information from storage if they are present", async () => {
       const clientRegistrar = getClientRegistrar({
         storage: mockStorageUtility(
           {
             mySession: {
               clientId: "an id",
               clientSecret: "a secret",
+              clientName: "my client name",
             },
           },
           false
@@ -99,9 +100,10 @@ describe("ClientRegistrar", () => {
       );
       expect(client.clientId).toEqual("an id");
       expect(client.clientSecret).toEqual("a secret");
+      expect(client.clientName).toEqual("my client name");
     });
 
-    it("properly performs dynamic registration", async () => {
+    it("properly performs dynamic registration and saves client information", async () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
       const mockedIssuer = {
@@ -113,9 +115,12 @@ describe("ClientRegistrar", () => {
         },
       };
       Issuer.discover = jest.fn().mockResolvedValueOnce(mockedIssuer);
+      const mockStorage = mockStorageUtility({});
 
       // Run the test
-      const clientRegistrar = getClientRegistrar();
+      const clientRegistrar = getClientRegistrar({
+        storage: mockStorage,
+      });
       const client = await clientRegistrar.getClient(
         {
           sessionId: "mySession",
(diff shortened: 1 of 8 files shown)


packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -41,23 +48,239 @@ describe("ClientRegistrar", () => {
   }
 
   describe("getClient", () => {
-    it("isn't implemented", async () => {
+    it("Fails if there is not registration endpoint", async () => {
```
pmcb55, line 51: Lowercase first char of `it("")` messages (it's supposed to read as a sentence, i.e. "this test, it **_fails if there is_* ..."):
```suggestion
    it("fails if there is not registration endpoint", async () => {
```

packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -41,23 +48,239 @@ describe("ClientRegistrar", () => {
...
           },
-          {
-            ...IssuerConfigFetcherFetchConfigResponse,
-            registrationEndpoint: registrationUrl,
-          }
+          IssuerConfigFetcherFetchConfigResponse
         )
-      ).rejects.toThrow();
+      ).rejects.toThrow(
+        "Dynamic client registration cannot be performed, because issuer does not have a registration endpoint"
```
pmcb55, line 75: Can we make this message a little more helpful by providing the Issuer URL or something, e.g. something like (just to give you the idea):
```suggestion
        "Dynamic client registration cannot be performed, because issuer at [${issuer}] does not have a registration endpoint (expected a valid HTTP response from something like [${issuer}/register])."
```

packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -41,23 +48,239 @@ describe("ClientRegistrar", () => {
...
+      );
+    });
+
+    it("retrieves client id and secret from storage if they are present", async () => {
```
pmcb55, line 79: Should we be getting the Client Name (as a string) too...? Is this description too implementation-specific (e.g. maybe change to 'client registration details' or something instead)? But at least if we **_are_** going to be explicit about what the test does, be very explicit, e.g.:
```suggestion
    it("retrieves the OIDC 'client_id' and 'client_secret' from storage if they are present", async () => {
```

packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -41,23 +48,239 @@ describe("ClientRegistrar", () => {
...
+        {
+          redirect_uris: ["https://example.com"],
+        },
+        {
+          initialAccessToken: "a token",
+        }
+      );
+    });
+
+    it("saves the registered client information in storage", async () => {
```
pmcb55, line 213: I'm not sure, but this test setup and execution seems identical to previous tests, with just the final two assertions being different. If that is the case, I wouldn't duplicate so much code - I'd just move these two assertions up into a previous test. So yes, it would be a single test testing for multiple behaviours, but I'd be fine with that (over code duplication), and if we really wanted to split it out into multiple independent tests (although I don't see a justification yet), then I'd refactor the test setup and execution into it's own callable static function.

packages/node/src/login/oidc/OidcLoginHandler.ts
```diff
@@ -84,6 +85,35 @@ export default class OidcLoginHandler implements ILoginHandler {
...
+      options.oidcIssuer
+    );
+    let clientInfo: IClient;
+    if (options.clientId !== undefined && options.clientId !== undefined) {
+      clientInfo = {
+        clientId: options.clientId,
+        clientSecret: options.clientSecret,
+      };
+    } else {
+      // If client id and secret aren't specified in the options, performe dynamic
```
pmcb55, line 99: Since these are OIDC spec-defined values, I'd be explicit in the comment. But again, are you being too implementation-specific here (and missing `client_name` too), e.g. we could just say 'the OIDC standard client registration details' or something instead?
```suggestion
      // If 'client_id' and 'client_secret' aren't specified in the options, perform dynamic
```

pmcb55 (approved):
LGTM (just minor comments really).

packages/node/src/login/oidc/ClientRegistrar.spec.ts
NSeydoux: The error message does include additional information (``Dynamic client registration cannot be performed, because issuer does not have a registration endpoint: ${JSON.stringify(issuer.metadata)}``), but I omitted that part of the error message in the test to make it easier to read.

packages/node/src/login/oidc/OidcLoginHandler.ts
NSeydoux: The `client_name` is left out on purpose: only the client id and secret are required in the process, the client name just being a "nice to have". I find it more explicit to specifically mention client id and secret in the comment, rather than just ' OIDC standard client registration details', because this behaviour is spec-driven, so I wouldn't regard it as an implementation detail.

packages/node/src/login/oidc/ClientRegistrar.spec.ts
NSeydoux: Resolved in 723613e09a9a80001374bf3a8be9f5c37cb62ad1

packages/node/src/login/oidc/ClientRegistrar.spec.ts
NSeydoux: Yes, the client name could be retrieved from memory. Fixed in 723613e09a9a80001374bf3a8be9f5c37cb62ad1

packages/node/src/login/oidc/OidcLoginHandler.ts
NSeydoux: Fixed comment in 723613e09a9a80001374bf3a8be9f5c37cb62ad1

PR #620, opened 2020-11-19
Feat/node auth code redirect

NSeydoux:
This retrieves data from storage (including an auth code), and exchanges said auth code for tokens at the issuer's token endpoint. Both DPoP and Bearer tokens may be retrieved, as well as refresh tokens if the IdP returns one.

NB: currently, the DPoP token isn't used to build a DPoP fetch, this will be done in an upcomig PR.

Also, note that commit 0df53ad6794ec862c2b71aaacea79a0a73e8314b should ideally be kept around, so I'll try to think about squashing only the others when merging. It contains work on `jose@3`.

### DIFF
--- a/packages/node/src/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/node/src/login/oidc/ClientRegistrar.spec.ts
@@ -24,8 +24,8 @@ import { mockStorageUtility } from "@inrupt/solid-client-authn-core";
 import ClientRegistrar from "./ClientRegistrar";
 import {
   IssuerConfigFetcherFetchConfigResponse,
-  mockDefaultIssuerConfig,
-  mockIssuerConfig,
+  mockDefaultIssuerMetadata,
+  mockIssuerMetadata,
 } from "./__mocks__/IssuerConfigFetcher";
 import {
   mockClientConfig,
@@ -51,7 +51,7 @@ describe("ClientRegistrar", () => {
     it("fails if there is not registration endpoint", async () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
-      const mockedIssuerConfig = mockIssuerConfig({
+      const mockedIssuerConfig = mockIssuerMetadata({
         registration_endpoint: undefined,
       });
       const mockedIssuer = {
@@ -107,7 +107,7 @@ describe("ClientRegistrar", () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
       const mockedIssuer = {
-        metadata: mockDefaultIssuerConfig(),
+        metadata: mockDefaultIssuerMetadata(),
         Client: {
           register: jest.fn().mockResolvedValueOnce({
             metadata: mockDefaultClientConfig(),
@@ -150,7 +150,7 @@ describe("ClientRegistrar", () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
       const mockedIssuer = {
-        metadata: mockDefaultIssuerConfig(),
+        metadata: mockDefaultIssuerMetadata(),
         Client: {
           register: jest.fn().mockResolvedValueOnce({
             metadata: mockDefaultClientConfig(),
@@ -194,7 +194,7 @@ describe("ClientRegistrar", () => {
       // Sets up the mock-up for DCR
       const { Issuer } = jest.requireMock("openid-client");
       const mockedIssuer = {
-        metadata: mockDefaultIssuerConfig(),
+        metadata: mockDefaultIssuerMetadata(),
         Client: {
           register: jest.fn().mockResolvedValueOnce({
             metadata: mockDefaultClientConfig(),
@@ -232,7 +232,7 @@ describe("ClientRegistrar", () => {
         client_secret: undefined,
       });
       const mockedIssuer = {
-        metadata: mockDefaultIssuerConfig(),
+        metadata: mockDefaultIssuerMetadata(),
         Client: {
           register: jest.fn().mockResolvedValueOnce({
             metadata: mockedClientConfig,
--- a/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
+++ b/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
@@ -26,16 +26,135 @@
 
 import { inject, injectable } from "tsyringe";
 import {
+  IClient,
   IClientRegistrar,
+  IIssuerConfig,
   IIssuerConfigFetcher,
   IRedirectHandler,
   ISessionInfo,
   ISessionInfoManager,
   IStorageUtility,
-  NotImplementedError,
 } from "@inrupt/solid-client-authn-core";
 import { URL } from "url";
-import { fetchType } from "../../../authenticatedFetch/fetchFactory";
+import { Issuer, TokenSet } from "openid-client";
+import type { KeyObject } from "crypto";
+import { JWTPayload } from "jose/jwt/verify";
+import { configToIssuerMetadata } from "../IssuerConfigFetcher";
+import {
+  buildBearerFetch,
+  buildDpopFetch,
+  fetchType,
+} from "../../../authenticatedFetch/fetchFactory";
+// import generateKeyPair from "jose/util/generate_key_pair";
+const { default: generateKeyPair } = require("jose/util/generate_key_pair");
+
+type OidcContext = {
+  sessionId: string;
+  issuerConfig: IIssuerConfig;
+  codeVerifier: string;
+  redirectUri: string;
+  dpop: boolean;
+};
+
+/**
+ * Based on the provided state, this looks up contextual information stored
+ * before redirecting the user to the OIDC issuer.
+ * @param oauthState The state (~ correlation ID) of the OIDC request
+ * @param storageUtility
+ * @param configFetcher
+ * @returns Information stored about the client issuing the request
+ */
+async function retrieveContextFromStorage(
+  oauthState: string,
+  storageUtility: IStorageUtility,
+  configFetcher: IIssuerConfigFetcher
+): Promise<OidcContext> {
+  // Since we throw if not found, the type assertion are ok too
+  const storedSessionId = (await storageUtility.getForUser(
+    oauthState,
+    "sessionId",
+    {
+      errorIfNull: true,
+    }
+  )) as string;
+
+  const [issueIri, codeVerifier, storedRedirectIri, dpop] = (await Promise.all([
+    storageUtility.getForUser(storedSessionId, "issuer", { errorIfNull: true }),
+    storageUtility.getForUser(storedSessionId, "codeVerifier", {
+      errorIfNull: true,
+    }),
+    storageUtility.getForUser(storedSessionId, "redirectUri", {
+      errorIfNull: true,
+    }),
+    storageUtility.getForUser(storedSessionId, "dpop", { errorIfNull: true }),
+  ])) as string[];
+
+  // Unlike openid-client, this looks up the configuration from storage
+  const issuerConfig = await configFetcher.fetchConfig(issueIri);
+  return {
+    sessionId: storedSessionId,
+    codeVerifier,
+    redirectUri: storedRedirectIri,
+    issuerConfig,
+    dpop: dpop === "true",
+  };
+}
+
+async function saveSessionInfoInStorage(
+  storageUtility: IStorageUtility,
+  sessionId: string,
+  idToken: string,
+  webId: string,
+  isLoggedIn: string,
+  refreshToken?: string
+): Promise<void> {
+  if (refreshToken !== undefined) {
+    await storageUtility.setForUser(
+      sessionId,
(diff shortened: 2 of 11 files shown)


packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -24,8 +24,8 @@ import { mockStorageUtility } from "@inrupt/solid-client-authn-core";
 import ClientRegistrar from "./ClientRegistrar";
 import {
   IssuerConfigFetcherFetchConfigResponse,
-  mockDefaultIssuerConfig,
-  mockIssuerConfig,
+  mockDefaultIssuerMetadata,
```
pmcb55, line 27: I'm just curious - what's the ratonale behind the name change from `Config` to `Metadata`? (This is referring to the JSON returned from the `.well-known/openid-configuration` endpoint, right (in which case I would have thought `Config` was more appropriate)...?)

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+    const storedSessionId = (await storageUtility.getForUser(
+      oauthState,
+      "sessionId",
+      {
+        errorIfNull: true,
+      }
+    )) as string;
+
+    const [
+      issueIri,
```
pmcb55, line 80:
```suggestion
      issuerIri,
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+    return {
+      sessionId: storedSessionId,
+      codeVerifier,
+      redirectUri: storedRedirectIri,
+      issuerConfig,
+      dpop: dpop === "true",
+    };
+  } catch (e) {
+    throw new Error(
+      `Could not retrieve the OIDC context from storage for request associated to state ${oauthState}: ${e.toString()}`
```
pmcb55, line 108:
```suggestion
      `Could not retrieve the OIDC context from storage for request associated with state [${oauthState}]: ${e.toString()}`
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+    );
+  }
+}
+
+async function saveSessionInfoInStorage(
```
pmcb55, line 113: We don't save session info anywhere else apart from Storage (right?), so can this name be simplified, i.e.:
```suggestion
async function saveSessionInfo(
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+): Promise<string> {
+  if (idToken.webid !== undefined && typeof idToken.webid === "string") {
+    return idToken.webid;
+  }
+  try {
+    const webid = new URL(idToken.sub);
+    return webid.href;
+  } catch (e) {
+    throw new Error(
+      `The ID token has a malformed 'sub' claim ([${idToken.sub}]), and no 'webid' claim.`
```
pmcb55, line 160:
```suggestion
      `The ID token has no 'webid' claim, and it's 'sub' claim of [${idToken.sub}] is invalid as a URL - error [${error}] .`
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+  isLoggedIn: string,
+  refreshToken?: string
+): Promise<void> {
+  if (refreshToken !== undefined) {
+    await storageUtility.setForUser(
+      sessionId,
+      {
+        refreshToken,
+      },
+      { secure: true }
```
pmcb55, line 127: Is the notion of 'secure' and 'insecure' storage relevant anymore (i.e. now that we use the fetch closure for 'secure'' data)...?

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -26,16 +26,141 @@
...
+
+/**
+ * Based on the provided state, this looks up contextual information stored
+ * before redirecting the user to the OIDC issuer.
+ * @param oauthState The state (~ correlation ID) of the OIDC request
+ * @param storageUtility
+ * @param configFetcher
+ * @returns Information stored about the client issuing the request
+ */
+async function retrieveContextFromStorage(
```
pmcb55, line 64: 'Context' is a very overloaded term - in this case we're specifically talking about OIDC context data, right? If so, I'd just call that out, i.e.:
```suggestion
async function retrieveOidcContextFromStorage(
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -74,8 +199,89 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+    // The type assertion is ok, because we checked in canHandle for the presence of a state
+    const oauthState = url.searchParams.get("state") as string;
+
+    const {
+      sessionId,
+      issuerConfig,
+      codeVerifier,
+      redirectUri,
+      dpop,
+    } = await retrieveContextFromStorage(
```
pmcb55, line 212:
```suggestion
    } = await loadOidcContextFromStorage(
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -74,8 +199,89 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+        // @ts-ignore
+        dpopKey as any
+      );
+    } else {
+      authFetch = buildBearerFetch(tokens.access_token, tokens.refresh_token);
+    }
+
+    const webid = await deriveWebidFromTokenPayload(tokens.claims());
+
+    await saveSessionInfoInStorage(
```
pmcb55, line 269:
```suggestion
    await saveSessionInfoToStorage(
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -74,8 +199,89 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+        // TS thinks dpopKey isn't initialized, when it is.
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        dpopKey as any
+      );
+    } else {
+      authFetch = buildBearerFetch(tokens.access_token, tokens.refresh_token);
+    }
+
+    const webid = await deriveWebidFromTokenPayload(tokens.claims());
```
pmcb55, line 267:
```suggestion
    const webid = await deriveWebIdFromTokenPayload(tokens.claims());
```

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -74,8 +199,89 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+      sessionId,
+      tokens.id_token,
+      webid,
+      "true",
+      tokens.refresh_token
     );
+
+    const sessionInfo = await this.sessionInfoManager.get(sessionId);
+    if (!sessionInfo) {
+      throw new Error(`Could not retrieve session: [${sessionId}].`);
```
pmcb55, line 280: Why not? Since we don't get any information as to why (i.e. no exception, just a null/undefined response), is the only possible reason that the Session ID wasn't found, i.e.:
```suggestion
      throw new Error(`Could not find any session information associated with SessionID [${sessionId}] in storage.`);
```

pmcb55 (approved):
LGTM (just minor naming suggestions and a question or two).

packages/node/src/login/oidc/ClientRegistrar.spec.ts
NSeydoux: It's because internally, we have a `IIssuerConfig` class, which we use in our interfaces, and there exists a similar class from `openid-client`, `IssuerMetadata`. I started mixing them up, so I used a clearer name for the mocks :)

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Not quite, and that's something that I'd like to sort out in a ticket on the current sprint about token and storage management.

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: I used 'derive' because it is used in the [WebID-OIDC spec](https://github.com/solid/webid-oidc-spec#deriving-webid-uri-from-id-token), but I'm fine with anything else. I used "payload" specifically, because we don't parse the entire ID Token, which also has headers and a signature, so payload here comes from JWT. I'd prefer to keep it in, because we on purpose don't parse the whole ID Token (that's what `.claims()` does, and it also verifies the signature).

I changed `tokens` to `tokenSet`, as well as a comment about what `.claims()` returns as you suggested.

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Yep, if the session cannot be retrieved at this point it means that looking up the storage for the given session ID failed.

packages/node/src/login/oidc/ClientRegistrar.spec.ts
pmcb55: Ah. Ok, well, in that case (i.e. if it confused me, and even you were getting confused!), then I'd try to make it even clearer again in the mocks, i.e.
```suggestion
  mockDefaultOpenIdClientIssuerMetadata,
```
...although that doesn't work so well either, mainly 'cos 'openid-client' is such a terribly unwieldy name (much like Solid Client Library is an unwieldy name, and why something like LIT works so much better instead... oh well...).
So whatever you think yourself - maybe it just looks strange here side-by-side with the our IIssuerConfig in this PR's diff, but if we can't come up with a better name that distinguishes our IssuerConfig and openid-client's one, then maybe just a comment on the mock definition, or something.

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Resolved in 687516df762ab9f8dbbaae9605bf250ac66de6dd

PR #627, opened 2020-11-20
Disabled ESS e2e tests until Cognito is supported

NSeydoux:
Testcafé seems not to play nicely with Cognito, so while we figure it out, ESS e2e tests are disabled.


pmcb55 (approved):
LGTM

PR #643, opened 2020-11-23
Read and write DCR info to/from consistent storage

NSeydoux:
Client registration information was written in one storage, and read from another, which prevented a registered client to be reused. Also, instead of re-discovering the Issuer, it is now rebuilt from cached information to avoid issueing multiple requests to the issuer config in one grant.

- [ ] The changelog has been updated, if applicable. *N/A: the changelog does not mention yet node support*

### DIFF
--- a/packages/node/src/login/oidc/ClientRegistrar.spec.ts
+++ b/packages/node/src/login/oidc/ClientRegistrar.spec.ts
@@ -57,7 +57,7 @@ describe("ClientRegistrar", () => {
       const mockedIssuer = {
         metadata: mockedIssuerConfig,
       };
-      Issuer.discover = jest.fn().mockResolvedValueOnce(mockedIssuer);
+      Issuer.mockReturnValue(mockedIssuer);
 
       // Run the test
       const clientRegistrar = getClientRegistrar({
@@ -114,7 +114,7 @@ describe("ClientRegistrar", () => {
           }),
         },
       };
-      Issuer.discover = jest.fn().mockResolvedValueOnce(mockedIssuer);
+      Issuer.mockReturnValue(mockedIssuer);
       const mockStorage = mockStorageUtility({});
 
       // Run the test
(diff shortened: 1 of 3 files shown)


packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -139,13 +139,58 @@ describe("ClientRegistrar", () => {
...
+      let client = await clientRegistrar.getClient(
+        {
+          sessionId: "mySession",
+          redirectUrl: "https://example.com",
+        },
+        {
+          ...IssuerConfigFetcherFetchConfigResponse,
+        }
+      );
+      // Re-request for the client. If not returned from memory, the mock throws.
```
pmcb55, line 179: Comment says the mock throws, but I don't see any `expect().toThrow();` to test for an exception being thrown...? And shouldn't it be testing for the expected reason for that exception too (i.e. the reasons being `Cannot register more than once`)...?

pmcb55 (approved):
LGTM (just one test seems a little unclear).

packages/node/src/login/oidc/ClientRegistrar.spec.ts
NSeydoux: The `openid-client` mock throws on purpose, to make sure that it isn't called twice, but we never expect the error to be thrown. If the test completes, it means that the client has been retrieved from storage, as intended, because otherwise the error would have been thrown.

pmcb55: Oh! Well then I'd rewrite the comment to be clearer, e.g.:
```
// Re-request for the client (which we expect to be returned from memory, as `openid-client` deliberately throws if we attempt to call it twice (since that would indeed be a mistake because...???).
```
(I'm not sure exactly what the 'because ...' part would say, but I know you know :) !)

NSeydoux: Just to be sure of the content of the comment before re-writing it, `openid-client` does not throw on purpose if we try to register a client twice, but the expected behaviour is to retrieve the already registered client from memory. The mock throws just to make sure that we don't make a second call, but that's not how the actual `openid-client` behaves. What do you think of the following comment: 
```
// Re-request the client information. The client being registered on the first call, the registration information are expected to be retrieved from storage on the second `getClient` call. The `openid-client` mock throws if it is called a second time to ensure that the client isn't registered again.
```

pmcb55: Perfect (apart from two teeny weeny tweaks :) !):
```
// Re-request the client information. The client being registered on the first call, the registration information is expected to be retrieved from storage on the second `getClient()` call. The `openid-client` mock throws if it is called a second time to ensure that the client isn't registered again.
```

PR #651, opened 2020-11-24
Moved storage utility functions to core module

NSeydoux:
Some storage functions are generic enough that they can be shared in the core module, so that they aren't duplicated between the browser and node modules.
This moves the oidc context read/write function from the node to the core module, ensures 100% code coverage on the moved code, and moves the tests alongside the code in the core.


packages/core/src/storage/StorageUtility.spec.ts
```diff
@@ -438,3 +443,185 @@ describe("StorageUtility", () => {
...
+        "someState",
+        mockedStorage,
+        mockIssuerConfigFetcher(mockIssuerConfig())
+      )
+    ).rejects.toThrow(
+      "Failed to retrieve OIDC context from storage for login request associated with state"
+    );
+  });
+
+  it("throws if no redirect uri is stored for the user", async () => {
```
pmcb55, line 515:
```suggestion
  it("throws if no redirect URI is stored for the user", async () => {
```

pmcb55 (approved):
LGTM

PR #657, opened 2020-11-24
Feat/node/refresh flow

NSeydoux:
Two commits: 
- When looking up storage to retrieve some OAuth context, one can have one of two entry points: either the session ID (e.g. when using a refresh token from a known session), or the OAuth state, e.g. when handling a redirect.
This splits the OIDC context retrieval to cater to these two use cases.
- This implements the refresh token flow, based on openid-client. The OIDC context is retrieved from storage, and used to get a new access token from a refresh token.
This PR also trims some code that was duplicated in the AuthCodeRedirectHandler, based on the mock examples used for the refresh flow.
Note that at this point, the token refreshed isn't "used" anywhere, and still needs to be fed into the fetch function by the AuthCodeRedirectHandler.


packages/core/src/storage/StorageUtility.spec.ts
```diff
@@ -444,33 +445,19 @@ describe("StorageUtility", () => {
...
-    const mockedStorage = mockStorageUtility({
-      "solidClientAuthenticationUser:mySession": {
-        issuer: "https://my.idp/",
-        codeVerifier: "some code verifier",
-        redirectUri: "https://my.app/redirect",
-        dpop: "true",
-      },
-    });
+describe("getSessionIdFromOauthState", () => {
+  it("returns undefined if no stored state matches the current request's", async () => {
```
pmcb55, line 449:
```suggestion
  it("returns undefined if we have no stored OIDC 'state' matching the current request's OIDC 'state' value", async () => {
```

packages/core/src/storage/StorageUtility.spec.ts
```diff
@@ -444,33 +445,19 @@ describe("StorageUtility", () => {
...
     await expect(
-      loadOidcContextFromStorage(
-        "some unexisting state",
-        mockedStorage,
-        mockIssuerConfigFetcher(mockIssuerConfig())
-      )
-    ).rejects.toThrow(
-      "Failed to retrieve OIDC context from storage for login request associated with state"
-    );
+      getSessionIdFromOauthState(mockedStorage, "some unexisting state")
```
pmcb55, line 453:
```suggestion
      getSessionIdFromOauthState(mockedStorage, "some non-existent 'state' value")
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
-      mocks.tokenRequester ?? defaultMocks.tokenRequester
+      mocks.issuerConfigFetcher ?? defaultMocks.issuerConfigFetcher,
+      mocks.clientRegistrar ?? defaultMocks.clientRegistrar
     );
   }
 
-  it("Refreshes the token properly", async () => {
-    const key = "global";
-    const refreshTokenValue = "refresh token value";
+  it("throws if no oidc issuer can be retrieved from storage", async () => {
```
pmcb55, line 144:
```suggestion
  it("throws if no OIDC issuer can be retrieved from storage", async () => {
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+  it("throws if a refresh token isn't provided", async () => {
+    setupDefaultOidcClientMock();
+    const mockedStorage = mockRefresherDefaultStorageUtility();
+
+    const refresher = getTokenRefresher({
+      storageUtility: mockedStorage,
+    });
+
+    await expect(refresher.refresh("mySession")).rejects.toThrow(
+      "Missing a refresh token to refresh the access token associated to the session [mySession]"
```
pmcb55, line 187:
```suggestion
      "Session [mySession] has no refresh token to allow it refresh it's access token."
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+    );
+  });
+
+  it("throws if a dpop token is expected, but no dpop key is provided", async () => {
```
pmcb55, line 191:
```suggestion
  it("throws if a DPoP token is expected, but no DPoP key is provided", async () => {
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+    const mockedStorage = mockRefresherDefaultStorageUtility();
+
+    const refresher = getTokenRefresher({
+      storageUtility: mockedStorage,
+    });
+
+    await expect(
+      refresher.refresh("mySession", "some refresh token")
+    ).rejects.toThrow(
+      "The key bound to the access token associated to the session [mySession] must be provided to refresh it."
```
pmcb55, line 202:
```suggestion
      "The key bound to the access token for session [mySession] must be provided to refresh it."
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+    );
+  });
+
+  it("Refreshes a dpop token properly", async () => {
```
pmcb55, line 206:
```suggestion
  it("refreshes a DPoP token properly", async () => {
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+
+    expect(refreshedTokens.access_token).toEqual(mockDpopTokens().access_token);
+  });
 
-    const storageMock = mockStorageUtility(
-      {
-        [`solidClientAuthenticationUser:${key}`]: {
-          refreshToken: refreshTokenValue,
-        },
+  it("Refreshes a bearer token properly", async () => {
```
pmcb55, line 223:
```suggestion
  it("refreshes a bearer token properly", async () => {
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+
+    const mockedStorage = mockRefresherDefaultStorageUtility();
+
+    const refresher = getTokenRefresher({
+      storageUtility: mockedStorage,
+    });
+
+    await expect(
+      refresher.refresh("mySession", "some old refresh token", mockJwk())
+    ).rejects.toThrow("The IdP did not return the expected tokens.");
```
pmcb55, line 284: Can we name the IdP (and are we consistently using "IdP" in our messages? I'd prefer "Identity Provider"), and can we list the 'expected tokens' that did not get returned?
```suggestion
    ).rejects.toThrow(`The Identity Provider [{$mockRefresherDefaultStorageUtility.getFromSession("mockRefresherDefaultStorageUtility").issuer}] did not return the expected tokens, expected [${expectedTokenList}].");
```

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -24,44 +24,279 @@ import {
...
+  });
+
+  it("throws if the IdP does not return an id token", async () => {
```
pmcb55, line 287:
```suggestion
  it("throws if the IdP does not return an 'id_token'", async () => {
```

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -42,21 +50,65 @@ export interface ITokenRefresher {
...
+    const clientInfo: IClient = await this.clientRegistrar.getClient(
+      { sessionId },
+      oidcContext.issuerConfig
+    );
+    const client = new issuer.Client({
+      client_id: clientInfo.clientId,
+      client_secret: clientInfo.clientSecret,
+    });
+
+    if (refreshToken === undefined) {
```
pmcb55, line 79: This conditional seems weirdly placed here - i.e. its seems to be testing the value of the input param, so should it just be the first line of the function? And why is the input param optional if we immediately throw if it's `undefined`?

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -42,21 +50,65 @@ export interface ITokenRefresher {
...
+      // TODO: in a next PR, look up storage for a refresh token
+      throw new Error(
+        `Missing a refresh token to refresh the access token associated to the session [${sessionId}]`
+      );
+    }
+
+    if (oidcContext.dpop && dpopKey === undefined) {
+      throw new Error(
+        `The key bound to the access token associated to the session [${sessionId}] must be provided to refresh it.`
```
pmcb55, line 88: I'm not sure exactly the best wording here, but the current message text makes no reference to DPoP at all, while the conditional statement is explicitly checking for it, so maybe something like:
```suggestion
        `For session [${sessionId}], the key bound to the access token (i.e. the DPoP key) must be provided to allow us refresh our access token.`
```

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -42,21 +50,65 @@ export interface ITokenRefresher {
...
+      DPoP: dpopKey?.toJWK(true),
     });
-    /* eslint-enable camelcase */
+
+    if (
+      tokenSet.access_token === undefined ||
+      tokenSet.id_token === undefined
+    ) {
+      // The error message is left minimal on purpose not to leak the tokens.
+      throw new Error("The IdP did not return the expected tokens.");
```
pmcb55, line 101: From earlier comment, you do have the info to make this message much more helpful, I would add the missing info:
```suggestion
      const bothMissing: boolean = (tokenSet.access_token === undefined && tokenSet.id_token === undefined);
      const missingList = `${bothMissing ? "'access_token' and 'id_token'" : ${tokenSet.access_token === undefined ? 'access_token' : 'id_token'}`;
      throw new Error("The Identity Provider [${issuer}] did not return the expected token[${bothMissing ? 's' : ''} [${missingList}].");
```

pmcb55 (approved):
LGTM - just minor comments on messaging really.

packages/core/src/storage/StorageUtility.spec.ts
NSeydoux: Resolved in 0ba475eb12e5452ea5a8427d6f1417b422d7fc4f

packages/core/src/storage/StorageUtility.spec.ts
NSeydoux: Resolved in 0ba475eb12e5452ea5a8427d6f1417b422d7fc4f

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: I changed `IdP` to `Identity Provider` in our error messages, and specified the expected token list. on second thought, in the refresh flow, only an access token should be expected (since the ID token should not expire).

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: Originally, the logic was that the refresh token could be available in storage, and not necessarily provided as a parameter: that's why it is optional. You are right that the test should be at the top of the function either way (if the token cannot be retrieved from storage and it isn't provided) though, it is now fixed.

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: A key-bound access token is a DPoP token. I'll rephrase.

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: The test is removed in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee, since actually only the access_token should be refreshed.

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: Resolved in 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: 8818797c031a7e51ec3c56e1bf2e195d54ff6dee

PR #660, opened 2020-11-25
Bearer-authenticated fetch supports the refresh token flow

NSeydoux:
This ensures that the authenticated fetch built after having received a Bearer token supports the refresh token flow. On auth failure, the fetch function tries to update its current access token, and retries the request. The refresh token is also rotated if a new token is sent by the issuer.

The DPoP fetch will incorporate a similar logic in an upcoming PR.

### DIFF
--- a/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
+++ b/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
@@ -109,6 +109,23 @@ describe("buildBearerFetch", () => {
     );
   });
 
+  it("returns a fetch preserving the optional headers even after refresh", async () => {
+    const fetch = jest.requireMock("cross-fetch");
+    fetch.mockResolvedValueOnce(mockStatusResponse(401));
+    const myFetch = buildBearerFetch("myToken", {
+      refreshToken: "some refresh token",
+      sessionId: "mySession",
+      tokenRefresher: mockDefaultTokenRefresher(),
+    });
+    await myFetch("someUrl", { headers: { someHeader: "SomeValue" } });
+
+    expect(fetch.mock.calls[0][1].headers.Authorization).toEqual(
+      "Bearer myToken"
+    );
+
+    expect(fetch.mock.calls[0][1].headers.someHeader).toEqual("SomeValue");
+  });
+
   it("rotates the refresh tokens if a new one is issued", async () => {
     const fetch = jest.requireMock("cross-fetch");
     fetch.mockResolvedValue(mockStatusResponse(401));
@@ -147,11 +164,7 @@ describe("buildBearerFetch", () => {
 
   it("returns the initial response when the refresh flow fails", async () => {
     const fetch = jest.requireMock("cross-fetch");
-    // If the fetch function is called a second time, the response status will be 418.
-    // The expected behaviour is to return 401, the original response, on refresh failure.
-    fetch
-      .mockResolvedValueOnce(mockStatusResponse(401))
-      .mockResolvedValueOnce(mockStatusResponse(418));
+    fetch.mockResolvedValueOnce(mockStatusResponse(401));
     const myFetch = buildBearerFetch("myToken", {
       refreshToken: "some refresh token",
       sessionId: "mySession",
--- a/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
+++ b/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
@@ -116,7 +116,7 @@ describe("buildBearerFetch", () => {
     expect(fetch.mock.calls[0][1].headers.someHeader).toEqual("SomeValue");
   });
 
-  it("rotates the refresh tokens if a new one is issued", async () => {
+  it("rotates the refresh token if a new one is issued", async () => {
     const fetch = jest.requireMock("cross-fetch");
     fetch.mockResolvedValue({ status: 401 });
     const tokenSet = mockDefaultTokenSet();
--- a/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
+++ b/packages/node/src/authenticatedFetch/fetchFactory.spec.ts
@@ -27,6 +27,11 @@ import {
   buildDpopFetch,
   DpopHeaderPayload,
 } from "./fetchFactory";
+import {
+  mockDefaultTokenRefresher,
+  mockDefaultTokenSet,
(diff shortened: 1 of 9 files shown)


packages/node/src/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -42,6 +47,16 @@ const mockNotRedirectedResponse = (): MockedRedirectResponse => {
   };
 };
 
+type MockedStatusResponse = {
```
pmcb55, line 50: Is this type really required, or the function `mockStatusResponse()` below it? i.e. rather than usage like:
```
fetch.mockResolvedValueOnce(mockStatusResponse(401));
```
...it's actually less code, and easier to read, to just write:
```
fetch.mockResolvedValueOnce({ status: 401 });
```
...but I'm guessing this might be to stop TypeScript complaining? If so, personally I'd still just name this type `MockStatusResponse` (it's only used here twice), and maybe just add a comment to explain why this more convoluted code.

packages/node/src/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -77,6 +92,93 @@ describe("buildBearerFetch", () => {
...
+    await myFetch("someUrl", { headers: { someHeader: "SomeValue" } });
+
+    expect(fetch.mock.calls[0][1].headers.Authorization).toEqual(
+      "Bearer myToken"
+    );
+
+    expect(fetch.mock.calls[0][1].headers.someHeader).toEqual("SomeValue");
+  });
+
+  it("rotates the refresh tokens if a new one is issued", async () => {
```
pmcb55, line 129:
```suggestion
  it("rotates the refresh token if a new one is issued", async () => {
```

packages/node/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -22,33 +22,89 @@
 import { JWK, JWT } from "jose";
 import { fetch } from "cross-fetch";
 import { v4 } from "uuid";
+import { ITokenRefresher } from "../login/oidc/refresh/TokenRefresher";
 
 // node-fetch fetch has an additional property (isRedirect) which prevents using
```
pmcb55, line 27: This comment is a bit confusing - I'd suggest this instead, and explain why we want to use `typeof fetch`:
```suggestion
// We don't import the node-fetch fetch, as it has an additional property (isRedirect) that would prevent us from using `typeof fetch`, which we want to use because...????
```

packages/node/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -22,33 +22,89 @@
...
-  authToken: string,
-  // TODO: We need to push this refresh token into a wrapper around the fetch,
-  //  so dependent on that wrapper existing first!
-  _refreshToken: string | undefined
+  accessToken: string,
+  refreshOptions?: RefreshOptions
 ): fetchType {
-  return (init: RequestInfo, options?: RequestInit): Promise<Response> => {
-    return fetch(init, {
+  // currentAccessToken and currentRefreshOptions are initialized with the provided
```
pmcb55, line 56: I'm always wary of explicitly naming things like variables or function/method names in comments, suggest:
```suggestion
  // These local variables are initialized with the provided input
```

packages/node/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -22,33 +22,89 @@
...
+    }
+    try {
+      const tokenSet = await currentRefreshOptions.tokenRefresher.refresh(
+        currentRefreshOptions.sessionId,
+        currentRefreshOptions.refreshToken
+      );
+      currentAccessToken = tokenSet.access_token;
+      if (tokenSet.refresh_token) {
+        // If the refresh token is rotated, update it in the closure.
+        // TODO: Plug this in an external storage if one is provided.
```
pmcb55, line 89:
```suggestion
        // TODO: Plug this into external storage, if one is provided.
```

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -36,11 +36,17 @@ import { Issuer, TokenSet } from "openid-client";
 import { JWK } from "jose";
 import { configToIssuerMetadata } from "../IssuerConfigFetcher";
 
+// Some identifiers are in camelcase on purpose.
```
pmcb55, line 39: I think this suggested change is more correct...?
```suggestion
// Some identifiers are *not* in camelCase on purpose, as they are named using the official names from the OIDC/OAuth2 protocols.
```

pmcb55 (approved):
LGTM

packages/node/src/authenticatedFetch/fetchFactory.spec.ts
NSeydoux: No you're right, the intermediary mock function isn't necessary, I anticipated that more properties would need to be passed, and that passing them manually could be cumbersome, but that's not the case. Resolved in bbba942939219f8c7351bee9bfa17ed20ee74dfe

packages/node/src/authenticatedFetch/fetchFactory.ts
NSeydoux: Actually, the comment is outdated, and `fetchType` is no longer necessary since I switched from `node-fetch` to `cross-fetch`, the latter providing type definitions aligned with the ones we expect. Resolved in c122bd2c2b0c73842f1c35ddc853f16d12b67201

packages/node/src/authenticatedFetch/fetchFactory.ts
NSeydoux: Resolved in bbba942939219f8c7351bee9bfa17ed20ee74dfe

packages/node/src/authenticatedFetch/fetchFactory.ts
NSeydoux: Resolved in bbba942939219f8c7351bee9bfa17ed20ee74dfe

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: Oops ^^. Resolved in bbba942939219f8c7351bee9bfa17ed20ee74dfe

PR #667, opened 2020-11-26
Read data from a JSON file

NSeydoux:
This introduces a new `IStorage` implementation, `FileSystemStorage`, which uses a JSON file on the filesystem as storage. The purpose is to be able to test easily the support for externally-stored tokens (i.e. this isn't intended to be used in a production system). It will also be a good exemple of extension if one wants to support their own storage solution for user credentials.

This only implements the read feature, write and delete will be implemented in separate PRs.


pmcb55 (approved):
LGTM

PR #675, opened 2020-11-27
Cleaned up dependencies

NSeydoux:
Some dependencies had accumulated and were no longer used. This removes them fromt the relevant `package.json` files.


pmcb55 (approved):
LGTM

PR #677, opened 2020-11-27
Only use one storage in NodeJS

NSeydoux:
Using two storages is a requirement in the browser, because we want to expose as little information as possible in the local storage.
In NodeJS, this constraint does not apply, and it's more convenient to store everything in one storage, since the problem of persistent/non-persistent
storage that forces to use the local storage in the browser does not exist on the server side.


pmcb55 (approved):
LGTM

PR #694, opened 2020-12-01
tidied up UI now that login working

pmcb55:
Chore - just tidying up UI to report resource details in a scrollable text area, and fixed up status messages.

### DIFF
--- a/packages/browser/examples/demoClientApp/src/DemoClientApp.js
+++ b/packages/browser/examples/demoClientApp/src/DemoClientApp.js
@@ -212,9 +212,22 @@ class DemoClientApp extends Component {
     // this.lookupIdentityProviderConfig(this.state.loginIssuer);
   }
 
-  highlightPartOfWebId(webId, part) {
-    // Example: https://ldp.demo-ess.inrupt.com/110592712913443799002/profile/card#me
+  /**
+   * Very hacky attempt at splitting a WebID into 'components', such that:
+   *  Component 0 is up to, but not including, the username part of the WebID.
+   *  Component 1 is just the username part of the WebID.
+   *  Component 2 is everything after the username part of the WebID.
+   *
+   * Example: https://ldp.demo-ess.inrupt.com/110592712913443799002/profile/card#me
+   *  Component 0 is "https://ldp.demo-ess.inrupt.com/".
+   *  Component 1 is "110592712913443799002".
+   *  Component 2 is "/profile/card#me".
+   */
+  extractComponentOfWebId(webId, part) {
     let result;
+
+
+
     switch (part) {
       case 0:
         result = webId.substring(0, webId.indexOf(".com/") + 5);
@@ -339,15 +352,20 @@ class DemoClientApp extends Component {
     return (
       <div style={style}>
         <div>
-          <div className="tooltip">
+          <div class="tooltip">
             <div>
               <span>Login with your Identity Provider:</span>
-              &nbsp;<i className="fa fa-info-circle"></i>
+              &nbsp;<i class="fa fa-info-circle"></i>
               &nbsp;
             </div>
 
-            <span className="tooltiptext">
+            <span class="tooltiptext">
               Your Identity Provider is who you trust to manage your identity.
+              <p></p>
+              Hover your mouse over the editbox and click the 'x' on the
+              right-hand side of the field to clear it, and then use the
+              'down arrow' icon to select from various public Identity
+              Provider options.
             </span>
           </div>
           <input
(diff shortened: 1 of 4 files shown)


packages/browser/examples/demoClientApp/src/DemoClientApp.js
```diff
@@ -212,9 +212,22 @@ class DemoClientApp extends Component {
...
+   *  Component 0 is up to, but not including, the username part of the WebID.
+   *  Component 1 is just the username part of the WebID.
+   *  Component 2 is everything after the username part of the WebID.
+   *
+   * Example: https://ldp.demo-ess.inrupt.com/110592712913443799002/profile/card#me
+   *  Component 0 is "https://ldp.demo-ess.inrupt.com/".
+   *  Component 1 is "110592712913443799002".
+   *  Component 2 is "/profile/card#me".
+   */
+  extractComponentOfWebId(webId, part) {
```
NSeydoux, line 226: Now that I think about it, don't we want to discourage from doing this kind of processing in the IRI at all, and consider it opaque ? I know that foaf isn't great, but shouldn't we look up the user profile for their `foaf:nick` or `foaf/vcard:name` instead ?

packages/node/example/demoClientApp/src/index.js
```diff
@@ -112,8 +114,17 @@ app.get("/redirect", async (req, res) => {
...
         } else if (info.isLoggedIn) {
-          webIdToRead = info.webId;
+
+          // This is deliberately making a big assumption about the structure of
+          // the logged-in WebID, but if the WebID does have a different
+          // structure, which is perfectly valid of course, then this code
+          // shouldn't alter it, and the user is free to enter whatever resource
+          // URI they actually want anyway, so it's just a convenience in the
+          // 'common' case.
+          resourceToRead = info.webId.replace("/profile/card#me", "/private/");
```
NSeydoux, line 124: Same here, I get the intent, but is it something that we want to feature in a demo app that we advertize for if it's a pattern we discourage ?

packages/node/example/demoClientApp/src/index.js
```diff
@@ -123,52 +134,49 @@ app.get("/redirect", async (req, res) => {
...
+        resourceValueRetrieved = `Failed to fetch from resource [${resourceToFetch}]: ${error}`;
+      }
     } catch (error) {
-      resourceValueRetrieved = `Failed to fetch from resource [${resourceToFetch}]: ${error}`;
+      resourceValueRetrieved = `Resource to fetch must be a valid URL - got an error parsing [${resourceToFetch}]: ${error}`;
     }
-  } catch (error) {
-    resourceValueRetrieved = `Resource to fetch must be a valid URL - got an error parsing [${resourceToFetch}]: ${error}`;
+
+    log.info(`Fetch response: [${resourceValueRetrieved}]`);
```
NSeydoux, line 167: Should this be a `log.error` ?

NSeydoux (approved):

packages/browser/examples/demoClientApp/src/DemoClientApp.js
pmcb55: Yeah, for sure (which is why the comment says this is 'very hacky'). It's an interesting question though, which I'd basically articulate as '**_WebID's are the core identifier, but they're also kinda terrible because they're known to be terrible identifiers for humans to remember (based on Yahoo research from many years ago), and so can we (or should we even try to) easily render the 'important', human-friendly portion of a WebID without having to dereference and lookup the profile document?_**'.
Here, I specifically **_don't_** want to display the `foaf:nick`, or the user's `rdfs:label` or `rdfs:comment`, or something - I specifically want to highlight the differences between different WebID's, if possible. We can do this currently with out-of-the-box ESS and NSS generated WebID's, but that's only 'cos they follow particular naming conventions (e.g. Sarven's WebID doesn't follow either: https://csarven.ca#i).
But yeah, I think I'll probably remove this whole thing, as we'll never be able to reliably determine anything from WebIDs. It was useful when I was first demo'ing this app, 'cos I was using two WebID's from the the same ESS server, which made it really difficult for a human to visually distinguish between them.
But perhaps I could replace this with a simpler, far more generic 'visually highlight the differences between two WebID's', as I think that could be implemented with simple string comparison (i.e. if the two strings only differ by a single substring, then just highlight that substring in both, or something like that...). I'll create a little note-to-self backlog ticket:

packages/node/example/demoClientApp/src/index.js
pmcb55: Yeah, again this was just a very convenient, hacky way to quickly demo accessing a private resource - it's definitely not ideal (hence the comment). We could consider using the previous-spec-mandated Private Type Registry instead, but I'm not sure what the current spec says about that (if anything at all). But yeah, if there is a spec-mandated way to safely determine a known private resource in all Pods, then this code should be updated to use that instead.

packages/node/example/demoClientApp/src/index.js
pmcb55: Yep.

PR #709, opened 2020-12-03
tidied up InruptError HTTP support

pmcb55:
# New feature description
Tidied up the HTTP response support in the InruptError class, and now just depend on RDF/JS types, using N3 only as a Dev Dependency now (and will look to replace that with Rdf-Ext if that is a lighter implementation than N3).

# Checklist


packages/core/src/errors/InruptError.ts
```diff
@@ -21,60 +21,107 @@
...
- *  Then we can either import that implementation in this code, or just import
- *  the types when needed here, e.g.:
- import type { NamedNode } from "rdf-js";
- */
-import { NamedNode } from "n3";
-
-import { VocabTerm } from "@inrupt/solid-common-vocab";
+import { VocabTerm, NamedNode } from "@inrupt/solid-common-vocab";
 
 export default class InruptError extends Error {
```
jholleran, line 50: The error type looks to have a lot of HTTP specific details in it. It might be worth considering renaming this to HttpError. InruptError seems very general.

packages/core/src/errors/InruptError.ts
```diff
@@ -21,60 +21,107 @@
...
+    return this.httpErrorResponse !== undefined;
+  }
+
+  getHttpResponse(): (Response & { ok: false }) | undefined {
+    return this.httpErrorResponse;
+  }
+
+  getHttpStatusCode(): number {
+    if (!this.hasHttpResponse()) {
+      throw new InruptError(
```
jholleran, line 109: If this really is an error if there is no HTTP Status code it should have been left in the constructor (or else use a builder pattern to create the Error). Any client code that is catching this error will have to _always_ have to try can catch these if it calles this function (which can get clunky).

packages/core/src/errors/InruptError.ts
pmcb55: Yeah, it's still early days for this class (i.e. no nested error support yet, no runtime dynamic error IRI lookup/retrieval/parsing, etc.), so maybe it looks like there's a lot of HTTP-specific 'stuff' in here. But the intent is just to have this generic InruptError provide **_support for_** HTTP errors responses, that's all - i.e. it's just **_one of_** the capabilities or features that the generic error class provides (e.g. along with nesting errors, looking up IRI's, doing positional parameter replacement, etc.). Does that make sense, or would you still feel that a completely separate HttpError class would be better...?

jholleran: My prefered option would be to keep all the HTTP specific stuff out of the Error entirely. This InruptError would be more useful if it just makes it easy to gather as much context information as possible. So wrapping nested errors, an IRI for the type or Error, plus holding other context information all makes sence here. Handling (or knowing anything about) HTTP or doing lookups (to me) sounds like the wrong place for this functionality. That functionality should be handled in the Error Handling layer.

NSeydoux: I agree that it feels a little too specific that the generic Inrupt error class has specific HTTP properties. What about creating some types that could be combined to the class, a bit like we do in `solid-client`, something like:
```
const type WithHttpError = {
  httpErrorResponse?: Response & { ok: false };
}

function buildHttpError(error: InruptError, httpError: Response & { ok: false }): InruptError & WithHttpError {
  return ...;
}
```
so that you could do 
```
throw buildHttpError(myInruptError, myHttpError)
```
 ?

I think this would go in the direction of a decorator pattern, but a builder approach is also something we could look at. This approach I think would make it relatively easy to wrap errors too, because all the wrapping logic could be in the `InruptError` class, and the context-specific information would go in the decorated types.

packages/core/src/errors/InruptError.ts
```diff
@@ -21,60 +21,107 @@
...
-          : InruptError.appendErrorIri(
-              InruptError.lookupErrorIri(messageOrIri, messageParams),
-              messageOrIri as NamedNode,
-              appendErrorIri
-            ),
-        httpResponse,
-        appendHttpDetails
-      )
+      typeof messageOrIri === "string"
+        ? InruptError.substituteParams(messageOrIri, messageParams)
```
NSeydoux, line 64: I'm not sure where `substituteParams` comes from, but I don't see any implementation... Should we consider using a proper templating library ?

packages/core/src/errors/InruptError.ts
pmcb55: Yeah, I agree with Jarlath - adding in the HTTP stuff was really just to support the current Error use-cases in this `solid-client-authn-js` library, and it doesn't really belong here. I totally agree with the generic InruptError basically just being a dumping ground for 'as much context information as possible' (which of course could include HTTP information), but I agree we should use wrappers/decorators/builders to dump that information (and since we can use an RDF dataset as the generic 'information dump', we can keep this class completely agnostic to the data it gathers together). We could then define HTTP-specific, or SOAP-specific, or Apache-Camel-specific classes to transform their domain-specific data into generic RDF to be dumped in here.

packages/core/src/errors/InruptError.ts
pmcb55: Yeah, I was uncomfortable having any method in this class throw an error, but just wanted to start off super-defensive. But yeah, factoring out the HTTP-awareness from this class completely would tidy this up too.

e2e/browser/package.json
```diff
@@ -11,7 +11,9 @@
   },
   "author": "Inrupt, Inc.",
   "license": "MIT",
-  "dependencies": {},
+  "dependencies": {
+    "@inrupt/solid-common-vocab": "^0.5.2"
+  },
```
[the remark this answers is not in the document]
pmcb55: I've no idea - definitely looks like a mistake. I removed it here: 0bd7593

packages/core/src/errors/InruptError.ts
pmcb55: The function `substituteParams()` is just a static function at the bottom of the file. Yeah, we could use a proper templating library if it brought something we needed, but at the moment this function is just 17 lines of code (which includes some nice error checking), so for now I'd certainly rather those 17 lines of code over bringing in another dependency.

packages/core/src/errors/InruptError.ts
pmcb55: And yeah Nic, we could create types like you suggest too, that would certainly be one option.

packages/core/src/errors/InruptError.test.ts
```diff
@@ -75,11 +76,15 @@ describe("InruptError", () => {
...
-      const error = new InruptError(message, undefined, failedResponse, false);
+      const error = new InruptError(message).httpResponse(
+        failedResponse,
+        false
+      );
 
       expect(error.name).toEqual("Error");
-      expect(error.statusCode).toEqual(404);
-      expect(error.statusText).toEqual("Not Found");
+      expect(error.hasHttpResponse()).toBeTruthy();
```
NSeydoux, line 85: Do we want to test for truthiness, or for equality with `true` ?

packages/core/src/errors/InruptError.test.ts
```diff
@@ -90,14 +95,46 @@ describe("InruptError", () => {
...
       expect(error.toString()).toContain("404");
     });
+
+    it("throws if no HTTP response", () => {
+      const message = "Normal error string...";
+      const error = new InruptError(message);
+
+      expect(error.hasHttpResponse()).toBeFalsy();
+      expect(() => error.getHttpStatusCode()).toThrow(
+        "can't get it's HTTP Status Code!"
```
NSeydoux, line 114:
```suggestion
        "can't get its HTTP Status Code!"
```

packages/core/src/errors/InruptError.test.ts
```diff
@@ -90,14 +95,46 @@ describe("InruptError", () => {
...
+      );
+      expect(() => error.getHttpStatusText()).toThrow(
+        "can't get it's HTTP Status Text!"
```
NSeydoux, line 117:
```suggestion
        "can't get its HTTP Status Text!"
```

NSeydoux (approved):

packages/core/src/errors/InruptError.test.ts
pmcb55: Yep: 2d63c87

packages/core/src/errors/InruptError.test.ts
pmcb55: Yep: 2d63c87

packages/core/src/errors/InruptError.test.ts
pmcb55: Yep: 2d63c87

PR #710, opened 2020-12-03
Correct phrasing of purpose in README

RubenVerborgh:
If I'm not mistaken, the README had things mixed up:
- Solid identity servers handle authentication themselves
- This is really a client of resource serves (leveraging a token obtained from an identity server)


NSeydoux:
That's a good point. However, I'd still mention the Identity Provider, because obtaining the token from it in the first place is also a feature of this library, even though "authentication" per se is handled by the identity provider. What about `solid-client-authn is a suite of libraries designed to interact with Solid Identity Providers in order to enable authenticated requests to Solid resource servers` ? Or even changing `resource server` into Pod ?

README.md
```diff
@@ -1,6 +1,6 @@
 # Solid JavaScript Authentication - solid-client-authn
 
-`solid-client-authn` is a suite of libraries designed to authenticate with Solid identity servers.
+`solid-client-authn` is a suite of libraries designed to authenticate against Solid resource servers.
```
pmcb55, line 3: Hi @RubenVerborgh  - thanks for pointing this out. I don't like the original either, but I don't think the latter quite fits or 'works' either. I chatted with @acoburn about it, and we came up with this suggestion instead:
```suggestion
`solid-client-authn` is a suite of libraries designed to manage the authentication required to access protected resources hosted by Solid servers.
```
@NSeydoux - have you any suggested improvements...?

RubenVerborgh: LGTM; perhaps `s/designed //` and `s/hosted by /on/` because it's quite verb-heavy.

NSeydoux: That works for me too

pmcb55: Sorry - not sure how I didn't see these comments until now, but yeah, those suggestions are great, so updating to:
```suggestion
`solid-client-authn` is a suite of libraries to manage the authentication required to access protected resources on Solid servers.
```

pmcb55 (approved):

PR #717, opened 2020-12-04
Add "browser" tag to package.json

### LINKED ISSUE #684: Export ES6 and browser builds
It seems we are building (and deploying to npm) a `browserDist` folder, but it is not exported through a `browser` field or similar, and hence not being picked up. Same for an ES6 build.

NSeydoux:
Resolves #684 (in part)

We use webpack to output a browser bundle, so it should be advertized in the package.json.

Note that this does not add the ES module build. For consistency, I'd like to switch the whole build system to rollup, so I'll make sure the an ES module build is generated and exported after the migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+## Bugfix
+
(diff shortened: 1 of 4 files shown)


RubenVerborgh (approved):
LGTM; you might want to test the browser dist at some point (even if just loading it up in Node or so).

PR #766, opened 2020-12-09
Bootstrap : add menu of IDPs; emit token as JSON

jeff-zucker:
This replaces the command-line arguments in the bootstrap script with a menu/prompt interface and a menu of IDPs.  It presents the returned token as a JSON string which can be plugged directly into the login arguments of a script.  Eventually, the initial prompts might ask for a file path and the script could end by writing (encrypted?) JSON to that file for easy retrieval by scripts,

### DIFF
--- a/packages/node/example/bootstrappedApp/README.md
+++ b/packages/node/example/bootstrappedApp/README.md
@@ -33,3 +33,7 @@ request. It expects a number of arguments:
 ```
 node app.js --clientId <the client id> --clientSecret <the client secret>  --refreshToken <the refresh token> --oidcIssuer <the issuer that issued the token> --resource <the private resource you want to access>
 ```
+Within a script, you may simply paste the credentials obtained in the getting-a-token step, as arguments to login.  For example;
+```javascript
+    login( <JSON returned from getting a cookie goes here> )
+```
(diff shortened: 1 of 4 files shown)


NSeydoux (commented):
Thank you very much for your suggestion Jeff ! Returning the login info as JSON definitely makes sense. As for the IdP menu, I actually wanted to evolve this with an interactive prompt using [inquierer](https://www.npmjs.com/package/inquirer), because there are additional items that I'd like the user to be able to specify interactively and not necessarily in the CLI arguments.

However, in order to improve usability, we are moving this app from the internal demo of this repo into its own place: https://github.com/inrupt/generate-oidc-token. Do you mind re-applying the JSON part of your commit there (so that you are credited as a contributor), but maybe hold on the IdP menu part given I'm going to use a different library for this ? If you prefer, I can also re-apply your suggestions myself, but then you would'nt appear as contributor in the history, which is a shame considering the value of your work :)

jeff-zucker:
Sounds good.  What about the readline interface as opposed to yargs - even without the menu, we can prompt for clientName (that is username, isn't it?) and IDP.  Users don't have to remember the syntax, and it makes it easier to extend if you want to ask more questions later.

NSeydoux:
I absolutely agree that none of these should be required: basically, my thinking was that all CLI options passed through yargs are optional (but you can still pass them if you want to for some reason), and any element not passed in via yargs gets prompted.

NSeydoux:
This has been ported to https://github.com/inrupt/generate-oidc-token

PR #806, opened 2020-12-15
Reads WebID from ID token in refresh flow

NSeydoux:
Instead of manually refreshing the token upfront, a lazy evaluation approach was previously implemented: the fetch function was provided an empty access token when logging in a script, additionnally to the refresh token and client credentials. On the first fetch, the absence of credentials returned 401, which triggered a regular refresh flow. Now, instead but instead of pretending we have an access token initially that will be discarded on the first use in order to get a proper access token, we just go through the refresh flow when logging in, which gets us both the access token that we give to the fetch, but also the ID token that gets us the WebID. Note that the access token isn't used in the refresh flow. The empty string was perviously just used to send a fetch that looks authenticated but gets rejected, and from then trigger the refresh flow.

This change is completely transparent to the user: in any case, you'll just call session.login . The fact that the access token is obtained at this point or on the first session.fetch is an implementation detail
The limitation of the previous approach was that the refresh flow only ever happened in the fetch closure, which prevented us from getting the WebID and attaching it to the session

### DIFF
--- a/packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.spec.ts
+++ b/packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.spec.ts
@@ -272,11 +272,11 @@ describe("RefreshTokenOidcHandler", () => {
       })
     );
     await expect(result).rejects.toThrow(
-      "The Identity Provider did not return an ID token on refresh, which prevents from getting the user's WebID."
+      "The Identity Provider [https://example.com] did not return an ID token on refresh, which prevents from getting the user's WebID."
     );
   });
 
-  it("uses the rotated refresh token to build the authenticated fetch if applicable", async () => {
+  it("uses the rotated refresh token to build the DPoP-authenticated fetch if applicable", async () => {
     const tokenSet = mockDefaultTokenSet();
     tokenSet.refresh_token = "some rotated refresh token";
     const mockedTokenRefresher = mockTokenRefresher(tokenSet);
@@ -294,6 +294,43 @@ describe("RefreshTokenOidcHandler", () => {
           clientId: "some client id",
           clientSecret: "some client secret",
         },
+      })
+    );
+    expect(result?.webId).toEqual("https://my.webid/");
+
+    const mockedFetch = jest.requireMock("cross-fetch");
+    mockedFetch.mockResolvedValue({
+      status: 401,
+      url: "https://my.pod/resource",
+    });
+    if (result !== undefined) {
+      // ... and this should trigger the refresh flow.
+      await result.fetch("https://some.pod/resource");
+    }
+    expect(mockedRefreshFunction.mock.calls[1]).toContain(
+      "some rotated refresh token"
+    );
+  });
+
+  it("uses the rotated refresh token to build the Bearer-authenticated fetch if applicable", async () => {
+    const tokenSet = mockDefaultTokenSet();
+    tokenSet.refresh_token = "some rotated refresh token";
+    const mockedTokenRefresher = mockTokenRefresher(tokenSet);
+    const mockedRefreshFunction = jest.spyOn(mockedTokenRefresher, "refresh");
+
+    // This builds the fetch function holding the refresh token...
+    const refreshTokenOidcHandler = new RefreshTokenOidcHandler(
+      mockedTokenRefresher,
+      mockStorageUtility({})
+    );
+    const result = await refreshTokenOidcHandler.handle(
+      mockOidcOptions({
+        refreshToken: "some refresh token",
+        client: {
+          clientId: "some client id",
+          clientSecret: "some client secret",
+        },
+        dpop: false,
       })
     );
     expect(result?.webId).toEqual("https://my.webid/");
--- a/packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.spec.ts
+++ b/packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.spec.ts
@@ -26,12 +26,17 @@
 
 import "reflect-metadata";
 import { mockStorageUtility } from "@inrupt/solid-client-authn-core";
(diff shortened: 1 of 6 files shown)


packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
```diff
@@ -56,6 +58,49 @@ function validateOptions(
...
+  const rotatedRefreshOptions = {
+    ...refreshOptions,
+    refreshToken: tokens.refresh_token ?? refreshOptions.refreshToken,
+  };
+  if (dpop) {
+    authFetch = await buildDpopFetch(
+      tokens.access_token,
+      // dpopkey is assigned in the previous if block
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
Vinnl, line 92: Isn't it just easier to do `if (dpop && dpopKey)` above?

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
```diff
@@ -56,6 +58,49 @@ function validateOptions(
...
+      refreshOptions.refreshToken,
+      dpopKey
+    );
+  } else {
+    tokens = await refreshOptions.tokenRefresher.refresh(
+      refreshOptions.sessionId,
+      refreshOptions.refreshToken
+    );
+  }
+  // Rotate the refresh token if applicable
```
Vinnl, line 82: Could you help me understand what this means? I thought the `.refresh()` call above already fetched a new refresh token if applicable?

Vinnl (approved):
I really don't have the necessary background to just the contents of this PR, but code-wise it looks good I guess...

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
NSeydoux: Unfortunately no, it still complains that `Variable 'dpopKey' is used before being assigned.`

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
NSeydoux: Rotating the refresh token means that on each use of a refresh token to get an access token, the IdP also invalidates the consumed refresh token, and returns a new one alongside the refreshed access token. 
 
The `.refresh` indeed has a `refresh_token` in the returned `tokens: TokenSet` if the IdP returns one as part of the refresh flow. That means that the original refresh token `refreshOptions.refreshToken` is no longer usable, and this code snippet ensures that the new refresh token `tokens.refresh_token` is given to the `fetch` function so that it can refresh its access token when it expires. If no new refresh token is provided, we can only assume that the previous  `refreshOptions.refreshToken` is still valid.

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
Vinnl: Ah, I see. So I see the same `if...else` here twice - couldn't the contents of this `if` just be added to the previous one? The only resulting duplication would be lines 83-86, but I think that actually still saves more lines than it costs, and I think it's less likely that that causes mistakes in the future than e.g. the `if` clauses no longer being aligned.

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
Vinnl: Ah right, so the refresh token passed to us by the developer might be invalid now, in which case we replace it by the new refresh token provided by the IdP. Got it, thanks!

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
NSeydoux: Yeah true, I guss I'll switch back to that. I initially went this way to reduce (a bit) cyclic complexity, and avoid having to write two very similar tests, one checking for token rotation with DPoP, and the other for Bearer tokens, but you're right that it's liklier that an issue is introduced if the `if` statements get disaligned ^^

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.spec.ts
```diff
@@ -246,4 +251,64 @@ describe("RefreshTokenOidcHandler", () => {
...
+      mockOidcOptions({
+        refreshToken: "some refresh token",
+        client: {
+          clientId: "some client id",
+          clientSecret: "some client secret",
+        },
+      })
+    );
+    await expect(result).rejects.toThrow(
+      "The Identity Provider did not return an ID token on refresh, which prevents from getting the user's WebID."
```
pmcb55, line 275: Shouldn't the IdP be referenced in this error message, e.g.:
```suggestion
      "The Identity Provider [${identityProvider}] did not return an ID token on refresh, which prevents us from getting the user's WebID."
```
(NOTE: I also added the word 'us' in that message to make is more readable)

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
pmcb55: Hhmmmm... seems to warrant an explanation in the code itself though, as this is tricky behaviour...?

pmcb55: (...in particular I mean the implicit expectation that no new refresh_token from the IdP means we assume the previous one if still 'good').

pmcb55 (approved):
LGTM (just a couple of minor comments).

PR #830, opened 2020-12-17
Make dependency on ajv explicit

NSeydoux:
This adds `ajv` in the runtime dependencies of `solid-client-authn-core`, because it isn't present at the moment while it is a direct dependency. The issue was hidden so far somehow, but using the library in `generate-oidc-token` revealed the issue.


Vinnl (approved):
I suppose it was a dependency of a devDependency before and did not get loaded because of that?

In any case, looks like yet another issue that would've been caught with post-packaging import tests :)

PR #835, opened 2020-12-17
Stores static clients when logging in

NSeydoux:
Dynamically registering clients gets them stored in localStorage, ready
to be reused after redirection. It was not the case for statically
registered client credentials, which prevented the token exchange to
complete successfully. This resolves the issue, making it possible to
use statically registered clients.


Vinnl (approved):
Loads of GitHub Actions warnings in the code touched in this PR. Do those commented-out tests still make sense?

NSeydoux:
They do actually, I'll have to do a round of cleanup to put them in the correct place.

PR #836, opened 2020-12-17
Minor changes to the instructions

edwardsph:
Changed the param passed to bootstrap and the name of the app used in the demo.


NSeydoux (approved):
Thank you very much for this ! For convenience, we are moving this app out of the internals of this repo to its own place (https://github.com/inrupt/generate-oidc-token), so that you can `npx` it, if you want to get updates from there :)

packages/node/example/bootstrappedApp/README.md
```diff
@@ -16,7 +16,7 @@ the build of the library. Then, run `npm ci` again, this time in
 ## Getting a token
 
 To get all the credentials needed to authenticate your script, you can `cd src/`,
-and then run `node bootstrap.js --oidcProvider <your identity provider>`. 
+and then run `node bootstrap.js --oidcIssuer <your identity provider>`.
```
pmcb55, line 19: As per comment on Engineering channel today (I know you're moving this repo, but I'm just noticing this here now):
```suggestion
and then run `node bootstrap.js --solidIdentityProvider <your Solid Identity Provider>`. 
```

NSeydoux:
Actually, what I'll do is cut a release of https://github.com/inrupt/generate-oidc-token, and update the instructions to use that instead.

Vinnl:
@NSeydoux Can we just merge this for now? There have been three PRs with this fix now, so people are running into this while waiting for the separate project: https://github.com/inrupt/solid-client-authn-js/pull/851

NSeydoux:
I agree with your proposed change @pmcb55, but I think it should rather be applied to https://github.com/inrupt/generate-oidc-token: this example app should be deprecated soon not to be confused with the "maintained" version in the dedicated repo.

PR #864, opened 2020-12-21
Allow storing test credentials in a local file

Vinnl:
With this PR you can create a `.env.local` file in the same directory as the end-to-end tests, in which you can list simple key-values to use instead of environment variables. An example is in `.env.defaults`. When values are not listed in these files, they will be read from actual environment variables as they are today.

This builds on #865, so it's easiest to review this after that is merged.


pmcb55 (approved):

pmcb55 (commented):
Actually, I just noticed - I'd have named `packages/node/src/e2e/.env.defaults` without the plural, i.e. just `packages/node/src/e2e/.env.default` instead :) !

.gitignore
```diff
@@ -67,6 +67,7 @@ typings/
 
 # dotenv environment variables file
 .env
+.env*.local
```
matthieubosquet, line 70:
```suggestion
.env*
!.env.defaults
```
Ignore all `.env` files except for `.env.defaults` instead of ignore `.env` and all `.env*.local` files.

matthieubosquet: Irrelevant suggestion, I get it now, only the local files are supposed to hold values that should not be version controlled. As per "[files under version control](https://www.npmjs.com/package/dotenv-flow#files-under-version-control)".

matthieubosquet: Would it be fair to actually remove .env from `.gitignore`?

Per: ".env – for default (fallback) values, tracked by VCS" - See: [Files under version control](https://www.npmjs.com/package/dotenv-flow#files-under-version-control).

Vinnl: Done, and renamed `.env.defaults` to `.env` as requested by @ajacksified in Slack: b2e8ffea66e112ac337b469bd9312e9f44055376.

PR #865, opened 2020-12-21
Remove depdendency on existing data in e2e tests

Vinnl:
The end-to-end tests attempt to fetch both public and private
Resources, and depended on a Pod being pre-initialised with known
access permissions.

To make the tests work with any Pod for which the credentials are
provided, we now use the Pod's WebID document as a public resource
(as that is required by the Solid spec to be public), and the Pod
root as a private resource (as that will always exist and usually
not be initialised to be public).

With that, the tests now also run against the production instance of ESS. I've currently used the same credentials as used for the solid-client end-to-end tests, but those can be changed at any time in the environment variables.

I've also aligned the names of the environment variables with those in solid-client. I'll remove the old ones (`CLIENT_ID`, `CLIENT_SECRET` and `REFRESH_TOKEN`) when this is merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,11 +31,11 @@ jobs:
       - run: npm run test
       - run: npm run e2e-test
         env: 
-          REFRESH_TOKEN: ${{ secrets.REFRESH_TOKEN }}
-          CLIENT_ID: ${{ secrets.CLIENT_ID }}
-          CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
-          OIDC_ISSUER: ${{ secrets.OIDC_ISSUER }}
-          POD_ROOT: ${{ secrets.POD_ROOT }}
+          E2E_TEST_REFRESH_TOKEN: ${{ secrets.E2E_TEST_REFRESH_TOKEN }}
+          E2E_TEST_CLIENT_ID: ${{ secrets.E2E_TEST_CLIENT_ID }}
+          E2E_TEST_CLIENT_SECRET: ${{ secrets.E2E_TEST_CLIENT_SECRET }}
+          E2E_TEST_IDP_URL: ${{ secrets.E2E_TEST_IDP_URL }}
+          E2E_TEST_ESS_POD: ${{ secrets.E2E_TEST_ESS_POD }}
       - run: npx prettier --check "{packages/*/src,packages/*/__tests__}/**"
       - run: npm audit --audit-level=moderate
       - name: Archive code coverage results
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,8 @@ jobs:
           REFRESH_TOKEN: ${{ secrets.REFRESH_TOKEN }}
           CLIENT_ID: ${{ secrets.CLIENT_ID }}
           CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
+          OIDC_ISSUER: ${{ secrets.OIDC_ISSUER }}
+          POD_ROOT: ${{ secrets.POD_ROOT }}
       - run: npx prettier --check "{packages/*/src,packages/*/__tests__}/**"
       - run: npm audit --audit-level=moderate
       - name: Archive code coverage results
--- a/packages/node/src/e2e/e2e-test.spec.ts
+++ b/packages/node/src/e2e/e2e-test.spec.ts
@@ -22,15 +22,14 @@
 import { it, describe } from "@jest/globals";
 import { Session } from "../Session";
 
-const OIDC_ISSUER = "https://broker.demo-ess.inrupt.com/";
-
 // This first test just saves the trouble of looking for a library failure when
 // the environment wasn't properly set.
 describe("Environment", () => {
   it("contains the expected environment variables", () => {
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -31,9 +31,11 @@ jobs:
       - run: npm run test
       - run: npm run e2e-test
         env: 
-          REFRESH_TOKEN: ${{ secrets.REFRESH_TOKEN }}
-          CLIENT_ID: ${{ secrets.CLIENT_ID }}
-          CLIENT_SECRET: ${{ secrets.CLIENT_SECRET }}
+          E2E_TEST_REFRESH_TOKEN: ${{ secrets.E2E_TEST_REFRESH_TOKEN }}
+          E2E_TEST_CLIENT_ID: ${{ secrets.E2E_TEST_CLIENT_ID }}
+          E2E_TEST_CLIENT_SECRET: ${{ secrets.E2E_TEST_CLIENT_SECRET }}
+          E2E_TEST_IDP_URL: ${{ secrets.E2E_TEST_IDP_URL }}
```
pmcb55, line 37: We shouldn't be using IdP as a name anywhere anymore (as per my comment in the Inrupt Engineering channel from Dec 18th (I don't want to paste a direct link in here!)). I know we'll have to update the GitHub secret name too, i.e.:
```suggestion
          E2E_TEST_SOLID_IDENTITY_PROVIDER_URL: ${{ secrets.E2E_TEST_SOLID_IDENTITY_PROVIDER_URL }}
```

.github/workflows/ci.yml
```diff
@@ -31,9 +31,11 @@ jobs:
...
+          E2E_TEST_ESS_POD: ${{ secrets.E2E_TEST_ESS_POD }}
```
pmcb55, line 38: We can discuss this if you like, but the idea here is that this refers to a specific instance of an ESS deployment, and 'POD' is not really an instance (although it is kinda the name of the instance in the URL itself). i.e. we could also refer to Dev and Demo and Staging instances, and we'd name them similarly, i.e. DEV, DEMO, STAGING:
```suggestion
          E2E_TEST_ESS_PROD: ${{ secrets.E2E_TEST_ESS_PROD }}
```

.github/workflows/ci.yml
Vinnl: Feel free to make the change, but it was a bit too much effort for me after having already spent quite a bit of time on this - keeping the names the same made it easier to re-use the solid-client credentials. If you do go over these names, could you do the same for solid-client?

pmcb55 (commented):
Just to note, although it's fine for now, the WebID profile document is not necessarily hosted in the user's Pod (i.e. it could be 'anywhere' on the web, e.g. it could be hosted by an IdP), so it's not really the ideal candidate to be used as a '_publicly available resource in the user's Pod_'. I don't think we have any spec-mandated public resources on Pods, so the profile document will have to do for now.

.github/workflows/ci.yml
Vinnl: Yeah, I wasn't quite sure yet what the best approach is. In solid-client I actually have the GitHub secrets use `PROD` in their name (so it's `E2E_TEST_ESS_PROD_POD`), which are then mapped to `E2E_TEST_ESS_POD` by the CI script. That way, we can store credentials for different environments in the secrets, and the CI can just run the e2e tests multiple times with different environment variables if it wants to test against different environments. Additionally, this would allow e.g. the ESS team to run the tests for one specific environment only after a deploy of that environment.

However, I'd also be fine with the e2e test script just listing all different environments themselves. As above, I didn't do that for this PR to limit the scope.

(That said, I'd be wary to test against dev environments of ESS - if every breakage there slows us down, that might be a bit too much.)

Vinnl:
@pmcb55 Agreed, it's not perfect, but less bad than what we had :)

.github/workflows/ci.yml
pmcb55: Ok - I'll approve this PR then, and just create a new ticket to sweep through the codebases and update the names.

pmcb55 (approved):
Approving this PR, but I'll create a new ticket to update the names (i.e. IDP => SOLID_IDENTITY_PROVIDER, and POD => PROD).

.github/workflows/ci.yml
pmcb55: Yeah, good point. I've noted this aspect of the issue in the new 'update naming' ticket I just created. Having all these hard-coded, and manually updated, GitHub secrets seems really unwieldy. Instead I'd much rather see these e-2-e tests register an entirely new Pod user, and use that user for their tests, tearing it down afterwards - I'll just create a linked ticket for that!

NSeydoux:
That works well for ESS, we just might want to follow up on the approach CSS takes, but for instance on NSS (which is not in scope here because it does not support refresh tokens) the Pod root is public, so using it as a private resource might not show some auth bugs. However, I assume CSS will have a private Pod root as well, in which case the assumptions made here work fine :)

Vinnl:
@NSeydoux Oh crap, you're right - even though the actual directory listing is private, I'd forgotten that NSS also has an HTML file there. Hmm, I suppose when SCAN also works against NSS we can just have it point to `/private/`.

PR #874, opened 2020-12-22
Do not auto-updated individual Lerna packages

Vinnl:
According to [1], Dependabot will automatically recognise Lerna
projects and update all its dependencies in one go. This should
significantly cut down on the noise of four different pull requests
when a shared dependency releases a new version. And since issues
caused by a dependency upgrade are probably similar across the
different packages, it probably makes sense to fix them together as
well.

Question: what do you think about also removing the examples from Dependabot? Since we're not running automated tests against them, we're not verifying whether the updated dependencies actually work, so the automated updates are more performative than anything. Alternatives would be just not updating dependencies (to make sure they keep working), or removing the lock file (to make it clear that up-to-date dependencies are untested).

[1] https://dependabot.com/blog/lerna-support/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,26 +14,6 @@ updates:
       interval: daily
     open-pull-requests-limit: 10
   - package-ecosystem: npm
-    directory: "/packages/browser"
-    schedule:
-      interval: daily
-    open-pull-requests-limit: 10
-  - package-ecosystem: npm
-    directory: "/packages/node"
-    schedule:
-      interval: daily
-    open-pull-requests-limit: 10
-  - package-ecosystem: npm
-    directory: "/packages/core"
-    schedule:
-      interval: daily
-    open-pull-requests-limit: 10
-  - package-ecosystem: npm
-    directory: "/packages/oidc"
-    schedule:
-      interval: daily
-    open-pull-requests-limit: 10
-  - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


pmcb55 (approved):
I'd suggest leaving Dependabot on for the examples/demos. It's kinda an interesting question though - i.e. I **_generally_** like writing demo or example code as a unit test anyway, simply so it's easy to execute (e.g. the Artifact Generator has been doing that from the start to test the generation of common vocab bundles).
But perhaps this is a justification to look into making all our example/demo code executable as simple unit tests too (so we could add their execution to our CI to ensure they continue to 'work', and it would also ensure they still work after Dependabot updates)...? (Possibly as simple as just adding 'public static void main(String[] args) {}'...?)
And BTW, I don't think simply removing the `package-lock.json` "**_makes it clear that up-to-date dependencies are untested_**" - at least not to the average JS developer (like me!) anyways :) !

Vinnl:
Agreed that unit tests would be best to demo (in which case dependabot still makes sense). Until we've got that it might make sense to disable Dependabot for them still, but let's first see how much this decreases the load.

NSeydoux:
I agree on the principle, but here most demo apps are web apps with a UI, in which case unit tests are more tedious, and I'm not sure the value outwheights the costs. I'd be in favor of keeping the lockfile, updating them only when Snyk complains about security issues, or manually when we release a major version of our libraries.

PR #902, opened 2021-01-04
Happy new year !

NSeydoux:


Vinnl (commented):
Happy new year! That said, I don't believe you have to update the copyright every year. (The notice is technically not needed anyway, since you automatically get copyright over your works, but even so it generally refers to the creation date. Alternatively, we could do `2020 and later` or something like that?)

NSeydoux:
That's an excellent question. I'll bring it up to Aaron, who was updating the ESS code this way, and see if it is necessary.

acoburn:
My understanding of copyright statements is that there are basically two approaches, but neither allows "2020 and later".

One approach:
Say you have 10 files, each of which has a "Copyright 2020 ..." statement. If, at any point in 2021 you update some of those files, those files will need the "Copyright 2021 ..." statement. This will lead to different files having different copyright years. IMO, that becomes a pain to manage -- it's also harder to enforce with tooling, since you can't say "require all files to contain a copyright statement with 'Copyright 2021 ...' or the like".

The other approach:
Again, 10 files, and say you want to enforce having a copyright statement with tooling, it just ends up being easier to manage if you change everything in one fell swoop.

A third approach is to just omit the copyright portion, but I'm assuming that's a non-starter here.

NSeydoux:
I think including the copyright notice in each source file is a legal recommendation, even though it may be more critical in the proprietary code than in the open code, as long as the overall project clearly shares a license (even though worst case scenario, the default license that applies is plain copyright, which is restrictive).

In any case, since this header is checked by the linter, indeed keeping everything aligned is much easier :)

Vinnl (approved):
Well, I don't really care, just thought that this looks like exactly the type of manual work that won't be done reliably, but it's not a big issue either way - as far as I know, a copyright notice is a courtesy, not a requirement, and hence there's no clearly defined format either. Although it's especially unclear to me what "2021" means in this case for files that were wholly created in 2020. Like, does copyright not apply if the year mentioned is not the current year? :)

PR #908, opened 2021-01-06
Added package entry in issue template

NSeydoux:
Since the repository hosts multiple packages, and the issue tracker is
at the repo level, issues would be more helpful if they specified which
package exactly is impacted by a bug.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -15,6 +15,16 @@ Please fill in as much of the template below as you're able. Feel free to delete
 ### Search terms you've used
 <!-- What search terms have you used to check whether this bug was already reported? -->
 
+### Impacted package
+
+What package(s) of the monorepo is impacted by the bug ?
+
+- [ ] solid-client-authn-browser
+- [ ] solid-client-authn-node
+- [ ] solid-client-authn-core
+- [ ] oidc-client-ext
+- [ ] Other (please specify): ...
+
 ### Bug description
 <!-- A short description of what the problem is. -->
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,7 +17,7 @@ Please fill in as much of the template below as you're able. Feel free to delete
 
 ### Impacted package
 
-What package(s) of the monorepo is impacted by the bug ?
+Which packages do you think might be impacted by the bug ?
 
 - [ ] solid-client-authn-browser
 - [ ] solid-client-authn-node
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -15,6 +15,16 @@ Please fill in as much of the template below as you're able. Feel free to delete
 ### Search terms you've used
 <!-- What search terms have you used to check whether this bug was already reported? -->
 
+### Impacted package
+
+What package(s) of the monorepo is impacted by the bug ?
```
Vinnl, line 20: This might be a new term for people and doesn't really matter to the answer:

```suggestion
What package(s) is impacted by the bug ?
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -15,6 +15,16 @@ Please fill in as much of the template below as you're able. Feel free to delete
...
+
+- [ ] solid-client-authn-browser
+- [ ] solid-client-authn-node
+- [ ] solid-client-authn-core
+- [ ] oidc-client-ext
+- [ ] Other (please specify): ...
```
Vinnl, line 26: Sounds like this might be useful :) (Though maybe just "is your app running in Node or in the browser" would be easier to answer and provide most of the relevant info.)

```suggestion
- [ ] Other (please specify): ...
- [ ] I don't know
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -15,6 +15,16 @@ Please fill in as much of the template below as you're able. Feel free to delete
...
 
+### Impacted package
+
+What package(s) of the monorepo would implement the feature ?
+
+- [ ] solid-client-authn-browser
+- [ ] solid-client-authn-node
+- [ ] solid-client-authn-core
+- [ ] oidc-client-ext
+- [ ] Other (please specify): ...
```
Vinnl, line 26: Ditto.

```suggestion
What package(s) would implement the feature ?

- [ ] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...
- [ ] I don't know
```

Vinnl (approved):
Just some suggestions, feel free to ignore.

.github/ISSUE_TEMPLATE/bug_report.md
pmcb55, line 20: Further to Vinnie's comment, maybe this wording is better (and removes the need for '...package(s) is(are)...' :) !)
```suggestion
Which packages do you think might be impacted by the bug ?
```

pmcb55 (approved):

.github/ISSUE_TEMPLATE/bug_report.md
NSeydoux: That's a good point, actually I'll change to that, in 90% of the cases that's going to be what we need to know, and I expect the remaining 10% to be bugs we report ourselves ^^

Issue #894, opened 2020-12-31
await session.logout() has no effect on session object


gibsonf1:
await session.logout() has no apparent effect on the session object itself but it does remove the ability to use the token.

As a work around, we are currently doing this:  session.info.isLoggedIn = false

Ideally, the logout would also hit the /.well-known/solid/logout  api on the app as the old solid-auth-client used to do, or is that option deprecated?

Thanks!

NSeydoux:
Hi @gibsonf1 , and thanks for reporting this ! #909 should resolve the issue. 

Unfortunately, `/.well-known/solid/logout` isn't spec-compliant with OpenID connect, which is why `solid-client-authn-browser` does not have this feature. But basically (and correct me if I'm wrong), this endpoint ends the session of the current user against the Solid server (both Pod server and Identity Provider), regardless of the app you are using, right ? To make a parallel, it is as if logging out of Gitter also logged you out of Github, or am I mistaking ?

Anyways, handling a post-logout redirect is something that we consider. It would be very helpful if you described your logout use case :).

NoelDeMartin:
I've also been wondering about this, because I noticed that `solid-auth-client` seems to log out from the idp as well, but this library doesn't (I'm not sure if `solid-auth-client` is calling `/.well-known/solid/logout` or what).

The use-case I've been thinking about is when I'm using someone else's computer, if I log into a Solid app and I logout, I wouldn't want that computer to be able to access that app again without introducing the password. And right now I'd have to visit the idp site manually (which, as user, I may not be familiar with because I was redirected to it during log in).

I agree with the simil with Github/Gitter, the difference I see is that Github is a standalone service and in the case of Solid apps it's possible that Solid apps is the only thing that users log into, and they may be unaware of the distinction. But yeah, I see the issues and I'm not sure what an authentication library should do by default.

In any case, it'd be nice to have the possibility to log out from the idp using this library as well, even if it's opt-in and not the default behaviour. That way, we could add something in the application's UI to configure it. For example, a check box saying "log out from idp as well" that is unchecked by default.

erikhofer:
@NoelDeMartin thanks for fixing this! But what @NoelDeMartin says is still valid. Maybe open a new ticket for further discussion?

NoelDeMartin:
@erikhofer I think you wanted to ping @NSeydoux on that first mention :)

[...]

PR #909, opened 2021-01-06
Bugfix: logout did not update session info

NSeydoux:
On logout, the `info` property on the sessio object wasn't updated, so
after logout, `session.info.isLoggedIn` still returned true. This is now
fixed.

This PR fixes bug #894.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ which will instantiate a new Session implicitly behind the scenes. If you do
 need access to this Session, you can do so using the new function
 `getDefaultSession`.
 
+### Bugfix
+
+- The `session.info.isLoggedIn` property is now set to false on logout. 
+
 The following sections document changes that have been released already:
 
 ## 1.2.6 - 2020-12-23
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -114,6 +114,14 @@ describe("Session", () => {
       await objectWithLogout.logout();
       expect(clientAuthnLogout).toHaveBeenCalled();
     });
+
+    it("updates the session's info", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const mySession = new Session({ clientAuthentication });
+      mySession.info.isLoggedIn = true;
+      await mySession.logout();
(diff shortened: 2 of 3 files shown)


Vinnl (commented):
Hmm, I'd done some shallow looking into this issue, but I hadn't expected this to be the fix. Tracing my steps (through what I still strongly feel are way too many layers of indirection :stuck_out_tongue: ):

- `Session.logout` calls
- `ClientAuthentication.logout` which calls
- `handle` on an `ILogoutHandler` (which I'm assuming will instantiate to `GeneralLogoutHandler`?), which then calls
- `sessionInfoManager.clear`, which then calls
- `clear` on an `IStorageUtility`.

Now, if I understand it correctly, the SessionInfoManager is the one responsible for the data in `session.info`, and looks into the StorageUtility for that. So how come `isLoggedIn` is still `true`?

NSeydoux:
To be exact, `SessionInfoManager` is responsible for the session information that's persisted in storage, but it's the `Session` object that maintains `session.info`. Actually, without the token being stored in local storage anymore, I'm not sure keeping any session information in the local storage is interesting, or at least what is stored could evolve quite  a bit.

When we call `sessionInfoManager.clear()`, what we really clear is all the additional information that's associated to the session (and that is both actually required to persist through redirection, and to be cleaned up on logout), e.g. the user's identity provider, and the PKCE token (code verifier). But the status of the session does not need to be persisted (and even, it's incorrect to persist it, since the session gets logged out on refresh, which doesn't update the status in storage). Which is why `session.info.isLoggedIn` was still unchanged after `sessionInfomanager.clean()`. Is it a bit clearer ?

Vinnl (approved):
Yeah I guess, although I do see a lot of references to `isLoggedIn` in `SessionManager.ts`.

And so there's no need to clear other data from `Session.info`, e.g. the WebID?

NSeydoux:
i was a bit hesitant on this: does it make sense to have a WebID set at all on a session that isn't logged in ? My current position is that if you log out, it may be helpful to keep the WebID on the `session.info` to display some information to the user who just logged out, but I also see how it would make sense to clear the WebID on logout, so not a strong opinion either ways.

PR #920, opened 2021-01-07
Feat/session management/rs cookie write

NSeydoux:
# New feature description

This write information about sessions on resource servers into local storage. When logging in an IdP, the library checks if the resource server that hosts the user's WebID has a `/session` endpoint. If so, information is stored in local storage following this schema:
```
"tmp-resource-server-session-info": {
  webId: <the user's WebID>,
  sessions: {
    <some-resource-server-iri>: {
      expires: <some timestamp>
    }
  }
}
```

Initially, only the resource server hosting the WebID will be persisted, but the model allows additional servers. 

The purpose of this information is to be able (in an upcoming PR) to be able to be aware that one has an active cookie for a given resource server, in order to initialize the session object appropriately. This approach is based on strong assumptions (the WebID is hosted on the same Pod as the data the user is going to access for instance), and it has known shortcomins. It is only a stopgap solution until a more permanent solution is implemented, which wouldn't require a cookie on the resource server side.

# Checklist

- [X] All acceptance criteria are met.
- [ ] The changelog has been updated, if applicable.
- [ ] New functions/types have been exported in `index.ts`, if applicable.
- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #921, opened 2021-01-07
Read the temporary session-workaround-cookie

Vinnl:
# New feature description

Unfortunately, regular sessions are lost when the user refreshes
the page or opens a new tab.
While we're figuring out the API for a longer-term solution, as a
temporary workaround some *resource* servers set a cookie that
keeps the user logged in after authenticated requests, and expose
the fact that they set it on a special endpoint.
After login, we store that fact in LocalStorage. This means that we
can now look for that data, and if present, indicate that the user
is already logged in.
Note that there are a lot of edge cases that won't work well with
this approach, so it will be removed in due time.

The type also still needs to be aligned with the write-side, specifically [this one](https://github.com/inrupt/solid-client-authn-js/pull/920/files#diff-659ac87dfd3711f4cfcea3c7bf6970980f4740fd59df45f04c7977bffaa23e98R118) and [this one](https://github.com/inrupt/solid-client-authn-js/pull/921/files#diff-4f6320deb9e43b5216428fb7c2dd230f4b0c5325c6f147f6a73b15659c355fb9R176-R181). I'm just already pushing it because I'm not working tomorrow, so Nicolas can finish it up.

# Checklist

- [ ] All acceptance criteria are met. -> This changed slightly; it's now being done in `handleIncomingRedirect` rather than in the constructor, as this will almost certainly align with the real implementation.
- [ ] Relevant documentation, if any, has been written/updated. N/A, I think?
- [ ] The changelog has been updated, if applicable. N/A, I think?

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -355,7 +355,7 @@ describe("Session", () => {
         });
       });
 
-      it("logs you in even if your Identity Provider is not your Resource server", async () => {
+      it("logs you in even if your Solid Identity Provider is not your Resource server", async () => {
         // Mock localStorage's `getItem` method:
         const existingLocalStorage = window.localStorage;
         const localStorageMock = {
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -303,20 +303,22 @@ describe("Session", () => {
           value: localStorageMock,
           writable: true,
         });
-        const clientAuthentication = mockClientAuthentication();
-        clientAuthentication.handleIncomingRedirect = jest.fn();
-        const mySession = new Session({ clientAuthentication });
-        await mySession.handleIncomingRedirect("https://some.url");
-        expect(mySession.info.isLoggedIn).toEqual(true);
-        expect(mySession.info.webId).toEqual("https://my.pod/profile#me");
-        expect(
-          clientAuthentication.handleIncomingRedirect
-        ).toHaveBeenCalledTimes(0);
-
-        // Remove the mocked method:
-        Object.defineProperty(window, "localStorage", {
-          value: existingLocalStorage,
-        });
+        try {
+          const clientAuthentication = mockClientAuthentication();
+          clientAuthentication.handleIncomingRedirect = jest.fn();
+          const mySession = new Session({ clientAuthentication });
+          await mySession.handleIncomingRedirect("https://some.url");
+          expect(mySession.info.isLoggedIn).toEqual(true);
+          expect(mySession.info.webId).toEqual("https://my.pod/profile#me");
+          expect(
+            clientAuthentication.handleIncomingRedirect
+          ).toHaveBeenCalledTimes(0);
+        } finally {
+          // Remove the mocked method:
+          Object.defineProperty(window, "localStorage", {
+            value: existingLocalStorage,
+          });
+        }
       });
 
       it("does not mark an almost-expired session as logged in", async () => {
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -267,6 +267,158 @@ describe("Session", () => {
       );
       expect(clientAuthnHandleIncomingRedirect).toHaveBeenCalled();
     });
+
+    // This workaround will be removed after we've settled on an API that allows us to silently
+    // re-activate the user's session after a refresh:
+    describe("(using the temporary workaround for losing a session after refresh)", () => {
+      it("does not report the user to be logged in when the workaround cookie is not present, and just executes the redirect", async () => {
+        const clientAuthentication = mockClientAuthentication();
+        clientAuthentication.handleIncomingRedirect = jest.fn(
+          async (_url: string) => undefined
+        );
+        const mySession = new Session({ clientAuthentication });
+        await mySession.handleIncomingRedirect("https://some.url");
+        expect(mySession.info.isLoggedIn).toEqual(false);
+        expect(mySession.info.webId).toBeUndefined();
+        expect(
+          clientAuthentication.handleIncomingRedirect
+        ).toHaveBeenCalledTimes(1);
+      });
+
+      it("re-initialises the user's WebID without redirection if the proper cookie is set", async () => {
+        // Mock localStorage's `getItem` method:
+        const existingLocalStorage = window.localStorage;
+        const localStorageMock = {
+          getItem: jest.fn(() =>
+            JSON.stringify({
+              webId: "https://my.pod/profile#me",
+              sessions: {
+                "https://my.pod/": { expiration: 9000000000000 },
+              },
+            })
+          ),
+        };
+
+        Object.defineProperty(window, "localStorage", {
+          value: localStorageMock,
+          writable: true,
+        });
+        const clientAuthentication = mockClientAuthentication();
+        clientAuthentication.handleIncomingRedirect = jest.fn();
+        const mySession = new Session({ clientAuthentication });
+        await mySession.handleIncomingRedirect("https://some.url");
+        expect(mySession.info.isLoggedIn).toEqual(true);
+        expect(mySession.info.webId).toEqual("https://my.pod/profile#me");
+        expect(
+          clientAuthentication.handleIncomingRedirect
+        ).toHaveBeenCalledTimes(0);
+
+        // Remove the mocked method:
+        Object.defineProperty(window, "localStorage", {
+          value: existingLocalStorage,
+        });
+      });
+
+      it("does not mark an almost-expired session as logged in", async () => {
+        // Mock localStorage's `getItem` method:
+        const existingLocalStorage = window.localStorage;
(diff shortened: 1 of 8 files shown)


NSeydoux:
(Note for after the weekend) There is still a small bug: the cookie is set correctly, and the session is initialized accordingly, but the fetch fails.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -267,6 +267,158 @@ describe("Session", () => {
...
+        clientAuthentication.handleIncomingRedirect = jest.fn();
+        const mySession = new Session({ clientAuthentication });
+        await mySession.handleIncomingRedirect("https://some.url");
+        expect(mySession.info.isLoggedIn).toEqual(true);
+        expect(mySession.info.webId).toEqual("https://my.pod/profile#me");
+        expect(
+          clientAuthentication.handleIncomingRedirect
+        ).toHaveBeenCalledTimes(0);
+
+        // Remove the mocked method:
```
pmcb55, line 316: Is this safe - i.e. if this one test fails mid-way, will it leave bad state? If so, this could be in a `try { } finally { <code here> }` block, or a Jest `afterTest()` cleanup function...?

packages/browser/__tests__/Session.spec.ts
```diff
@@ -267,6 +267,158 @@ describe("Session", () => {
...
+          clientAuthentication.handleIncomingRedirect
+        ).toHaveBeenCalledTimes(1);
+
+        // Remove the mocked method:
+        Object.defineProperty(window, "localStorage", {
+          value: existingLocalStorage,
+        });
+      });
+
+      it("logs you in even if your Identity Provider is not your Resource server", async () => {
```
pmcb55, line 356: This is probably going to be a pain to remember and implement, but I think we agreed to always try to refer to 'Solid Identity Providers', i.e:
```suggestion
      it("logs you in even if your Solid Identity Provider is not your Resource server", async () => {
```

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -87,6 +87,10 @@ async function setupResourceServerSession(
 ): Promise<void> {
   const webIdAsUrl = new URL(webId);
   const resourceServerIri = webIdAsUrl.origin;
+  // Querying the /session endpoint does not set the cookie, but issuing an
+  // authenticated request to any actual resource (even public ones) does,
+  // so we fetch the user's WebID before checking the /session endpoint.
+  await authenticatedFetch(webId);
```
pmcb55, line 93: In the code just below this, after we check the RS gives us back a 200, shouldn't we be checking that the value of the WebID in the response body also matches what we think our WebID is (i.e. the value of the passed-in `webId` parameter)...?

pmcb55 (approved):

packages/browser/__tests__/Session.spec.ts
NSeydoux: It had some side-effect, and an error thrown in the middle of this test failed at least one other test in a non-trivial manner. I added a `try/finally` here: 17d07888e28a48e8dca085b94e9571d9927d4d21

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Technically we could, but I'm not sure how much security it would add: the cookie would still be set regardless, so the user would have access to the resource server based on the WebID of the cookie, even in the (unlikely) case where it doesn't match the WebID attached to the ID token and to the Access token. I only say it's unlikely because only ESS will ever implement this cookie-based session management, and given it's a temporary workaround, I don't think it creates much value to consider edge cases, but rather we should move quick to a cleaner solution. What do you think ?

NSeydoux (approved):

packages/browser/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -38,6 +38,7 @@ export function buildBearerFetch(
   return (init, options): Promise<Response> => {
     return fetch(init, {
       ...options,
+      credentials: "include",
```
Vinnl, line 41: For my education: what's the difference between the `fetch` created by this factory, and the `fetchWithCookies` created in ClientAuthentication?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
Vinnl: Agreed that it would be better to do that if this were a long-term solution, but given the temporary nature of this and the likely usages, I also agree that there's no need to implement this now. (And to avoid the risk of there being nothing as permanent as a temporary solution, it might be _better_ not to treat it as something permanent.)

Vinnl (commented):
Identified one potential error, the rest are just minor suggestions or questions.

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -145,8 +145,8 @@ function mockClientRegistrar(client: IClient): IClientRegistrar {
   };
 }
 
-const mockFetchOnce = (response: Response): void => {
-  window.fetch = jest.fn().mockReturnValueOnce(
+const mockFetch = (response: Response): void => {
```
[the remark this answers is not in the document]
NSeydoux: Since the cookie isn't set when we hit the `/session` endpoint, but when we request one of the user's resources, we need to mock two requests instead of one.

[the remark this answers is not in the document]
Vinnl: Ah got it, thanks.

packages/browser/src/authenticatedFetch/fetchFactory.ts
NSeydoux: `fetchFactory` creates an authenticated fetch from an access token, by setting the `Authorization` header. In this case, the `credentials: "include"` entry enables (if I understood it correctly) to set the cookies based on the response's `Set-Cookie` headers, which seem to be ignored otherwise. This authenticated fetch is the one that will allow the first access to a user resource and set the Resource Server cookie on the browser.

On the other hand, `fetchWithCookie` just enhances a "plain" fetch (so a non-authenticated one, without any `Authorization` header) so that if cookies exist on the browser, they get sent along the request. Note that the cookies need to have the `SameOrigin=None` attribute set, so that they can be sent on cross-orogin requests. Unlike the authenticated fetch, `fetchWithCookie` would never set a cookie if it doesn't exist: the request would simply return 401.

packages/browser/src/ClientAuthentication.ts
```diff
@@ -84,14 +94,14 @@ export default class ClientAuthentication {
...
-  fetch = window.fetch;
+  fetch = fetchWithCookies;
 
   logout = async (sessionId: string): Promise<void> => {
     await this.logoutHandler.handle(sessionId);
 
     // Restore our fetch() function back to the environment fetch(), effectively
     // leaving us with un-authenticated fetches from now on.
-    this.fetch = window.fetch;
+    this.fetch = fetchWithCookies;
```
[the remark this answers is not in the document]
NSeydoux: Both very good points: 81281258f3ac7c28fe4834eb6e7f6045b04028b0 and 4c4455ea084ef31f86a04ed991a5119ae2db8d68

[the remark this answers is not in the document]
Vinnl: :+1: Any reason it couldn't be set to `window.fetch` like before?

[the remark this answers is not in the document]
NSeydoux: If set to `windows.fetch`, fetching after logging out threw the error `Calling 'fetch' on an object not implementing the Window interface`, which was also an issue documented in https://github.com/inrupt/solid-client-authn-js/pull/921/commits/520a19bdf9ca2d4e6e7e377a3dcf24053cf4ba70#diff-4f6320deb9e43b5216428fb7c2dd230f4b0c5325c6f147f6a73b15659c355fb9L132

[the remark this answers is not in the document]
Vinnl: I think that should be solvable by setting it to `window.fetch.bind(window)`, or possibly by just referring to the `fetch` global, rather than wrapping it in a wrapper function with identical signature. Not a biggie, but would be future-proof against updates of the Fetch API.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
pmcb55: Sure, I get the short-term situation thing. As a minimal-work compromise then, I'd just add a simple comment to the code that links to the first comment in this wee thread (I'm just a tad worried that this short-term solution becomes a longer-than-expected-term solution, so a comment in the code might at least catch someone's attention later if it does linger...!).

PR #928, opened 2021-01-08
Rephrased e2e test description with debug hints

NSeydoux:
Replay of d02a42a3e937b774287df00a6ba47c34da15ac3b without the
package-lock updates.

### DIFF
--- a/e2e/browser/README.md
+++ b/e2e/browser/README.md
@@ -1,25 +1,37 @@
-# solid-client-authn-testcafe
+# End-2-End Browser Tests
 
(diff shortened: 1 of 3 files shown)


pmcb55 (approved):
Cool. Yeah, I noticed the `optional: true` changes to the `package-lock.json` too, and have no idea what's going on there. I did run `npm ci`, but that shouldn't change it (I thought), but I noticed before VInnie had similar looking issues, and didn't know what the issue was (if I remember correctly). Curiosier and curiosier...

PR #941, opened 2021-01-12
Refactor get, implement set and delete

NSeydoux:
The get method previously parsed the file on each read. This changes it
to only parse the file once on load, and to read it from memory on
subsequent reads.

The set method writes a new value in the in-memory cache, and persists
it on the disk. This way, the persisted value is always the latest. As write, 
delete also updates both the in-memory cache and the file on
disk.

- [X] new classes exported from `index.ts`
- [x] CHANGELOG updated


packages/node/src/storage/FileSystemStorage.spec.ts
```diff
@@ -20,88 +20,188 @@
...
+      );
+    });
+
+    it("throws in case of write error", async () => {
+      const mockedFs = jest.requireMock("fs");
+      mockedFs.existsSync = jest.fn().mockReturnValue(true);
+      mockedFs.readFileSync = jest.fn().mockReturnValue("{}");
+      mockedFs.promises = {
+        writeFile: () => {
+          throw new Error("Some write error");
```
Vinnl, line 139: Given that it's an async function, I think this is a more accurate simulation of its behaviour:

```suggestion
          return Promise.reject(new Error("Some write error"));
```

Vinnl (approved):
Code-wise this seems fine, but I'm having some doubts about whether this is the approach we want to take. It seems kind-of limited and with a number of assumptions (e.g. how to deal with race conditions), so I'm wondering whether we don't just want to allow the developer to pass in their own object with a `get`, `set` and `delete` function and implement it in whatever way works for them? Possibly providing a simple in-memory storage for simple use cases.

Unless I'm missing what this is for?

NSeydoux:
You are absolutely right that this is a very simplistic approach, and it is mostly intended to:
- enable testing with a persistent storage in NodeJS, to deal with refresh tokens in a cleaner manner
- provide an example to developers who want to bind to their own storage system.

Currently, when you do `const session = new Session()`, an in-memory storage is used, which disappears if the script stops. The intention of this piece of code is to make it possible to do 
```
const storage = new FileSystemStorage("../session-data.json")
  const session = new Session({
    secureStorage: storage,
    insecureStorage: storage
  });
```
so that refresh tokens aren't lost when the server restarts. One could also do domething like
`const storage = new DatabaseStorage(<database connection information>)`, or whatever backend storage they want to use for their NodeJS app.

Vinnl:
Hmm OK, but if it's just an example do we want to export it as part of our public API? Same goes for the `insecureStorage`/`secureStorage` API - why would I ever want to supply an insecure storage? Especially if we're passing in the same value for both?

NSeydoux:
I intended to use it in the `packages/node/example/bootstrappedApp` script, so I could very well just move both files here, and not include `FileSystemStorage` in `solid-client-authn-node` codebase, and I think that's a good idea, I'll change this. 

Regarding the name of the two storage, once again it comes from the isomorphic API, so we can rename them on the Node side, because the issue doesn't apply, but in the `browser`, the `insecureStorage` is the `localStorage`, which is the only persistent storage we have available,  but which is also available to other scripts on the same origin. On the NodeJS side, these could be called `persistentStorage` and `volatileStorage`, or even just exposing a single `storage` parameter, since there's no particular reason to have the split we have on the browser.

Vinnl:
Yeah sounds good. For the browser library, I'd even question whether we want to expose custom storage layers in the first place, or at least whether we really need to expose an option for _both_ persistent storage (and whether we require that to be insecure, as the API currently implies) and volatile storage. But that's a discussion for another day :)

pmcb55 (approved):

Issue #944, opened 2021-01-13
Unauthenticated fetch causes TypeError: 'fetch' called on an object that does not implement interface Window


RubenVerborgh:
### Impacted package

Which packages do you think might be impacted by the bug ?

- [x] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description
When upgrading https://github.com/RubenVerborgh/solid-patch-tool from v1.1.1 to 1.4.0, the same code now crashes with
> TypeError: 'fetch' called on an object that does not implement interface Window

### To Reproduce
1. Go to https://rubenverborgh.github.io/solid-patch-tool/
2. Click "Apply patch" (without first logging in)
3. 
4. 

### Expected result
It does a normal fetch

### Actual result
It crashes

### Environment
Please run

  npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers

in your project folder and paste the output here:

```
  System:
    OS: macOS Mojave 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 363.47 MB / 16.00 GB
    Shell: 5.0.18 - /usr/local/bin/bash
  Binaries:
    Node: 14.4.0 - ~/.nvm/versions/node/v14.4.0/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.9 - ~/.nvm/versions/node/v14.4.0/bin/npm
  Browsers:
    Chrome: 87.0.4280.141
    Firefox: 84.0.2
    Safari: 14.0.2
  npmPackages:
    @inrupt/solid-client-authn-browser: ^1.4.0 => 1.4.0
    webpack: ^4.44.2 => 4.44.2
    webpack-cli: ^4.2.0 => 4.2.0
```

NSeydoux:
Thanks for reporting this ! I'll investigate first thing tomorrow morning.

NSeydoux:
Release 1.4.1 should include the fix for this issue.

matthieu-fesselier:
I currently face exactly the same issue:
```
  System:
    OS: macOS 11.3
    CPU: (4) x64 Intel(R) Core(TM) i5-6267U CPU @ 2.90GHz
    Memory: 2.05 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
[...]

PR #946, opened 2021-01-14
Fix unauthenticated fetch bug

NSeydoux:
This PR fixes bug #944 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ deprecated, and replaced by `storage`.
 - When providing a storage that already contains session information in the NodeJS
 context (as demonstrated with the `FileSystemStorage` in `packages/node/example/bootstrappedApp`),
 the refresh token is now picked up instead of the request failing.
+- Calling `Session::fetch` before logging the Session in threw an error.
 
 ## 1.4.0 - 2020-01-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,15 @@ deprecated, and replaced by `storage`.
 
 ### Bugfix
 
+
+#### node
+
 - When providing a storage that already contains session information in the NodeJS
 context (as demonstrated with the `FileSystemStorage` in `packages/node/example/bootstrappedApp`),
 the refresh token is now picked up instead of the request failing.
+
+#### browser
+
 - Calling `Session::fetch` before logging the Session in threw an error.
 
 ## 1.4.0 - 2020-01-11
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -125,11 +125,12 @@ describe("Session", () => {
   });
 
   describe("fetch", () => {
-    it("wraps up ClientAuthentication fetch", async () => {
+    it("wraps up ClientAuthentication fetch if logged in", async () => {
       window.fetch = jest.fn();
       const clientAuthentication = mockClientAuthentication();
       const clientAuthnFetch = jest.spyOn(clientAuthentication, "fetch");
       const mySession = new Session({ clientAuthentication });
+      mySession.info.isLoggedIn = true;
       await mySession.fetch("https://some.url");
       expect(clientAuthnFetch).toHaveBeenCalled();
     });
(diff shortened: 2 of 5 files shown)


packages/browser/src/Session.ts
```diff
@@ -129,6 +129,12 @@ export class Session extends EventEmitter {
    * @param init Optional parameters customizing the request, by specifying an HTTP method, headers, a body, etc. Follows the [WHATWG Fetch Standard](https://fetch.spec.whatwg.org/).
    */
   fetch = async (url: RequestInfo, init?: RequestInit): Promise<Response> => {
+    if (!this.info.isLoggedIn) {
+      // TODO: why does this.clientAuthentication.fetch return throws
+      // ""'fetch' called on an object that does not implement interface Window"
+      // when unauthenticated ?
```
RubenVerborgh, line 135: Let me guess, do we somewhere have `clientAuthentication.fetch = window.fetch`? That would explain it. It should be `window.fetch.bind(window)`.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -139,12 +140,26 @@ describe("Session", () => {
...
+    // ""'fetch' called on an object that does not implement interface Window"
+    // when unauthenticated.
+    it("does not call authenticated fetch if logged out", async () => {
+      window.fetch = jest.fn();
+      const clientAuthentication = mockClientAuthentication();
+      const clientAuthnFetch = jest.spyOn(clientAuthentication, "fetch");
+      const mySession = new Session({ clientAuthentication });
+      await mySession.fetch("https://some.url/");
+      expect(clientAuthnFetch).not.toHaveBeenCalled();
+      expect(window.fetch).toHaveBeenCalled();
```
RubenVerborgh, line 161: You explicitly want to check here whether `window.fetch` has been called with `window` as `this` pointer.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -139,12 +140,26 @@ describe("Session", () => {
...
       };
       await objectWithFetch.fetch("https://some.url");
       expect(clientAuthnFetch).toHaveBeenCalled();
     });
+
+    // Prevents the bug when clientAuthentication.fetch throws
+    // ""'fetch' called on an object that does not implement interface Window"
+    // when unauthenticated.
+    it("does not call authenticated fetch if logged out", async () => {
+      window.fetch = jest.fn();
```
RubenVerborgh, line 155:
```suggestion
      window.fetch = jest.fn(function () {
        expect(this).toBe(window);
      });
```

RubenVerborgh (changes requested):
The actual bug is here: https://github.com/inrupt/solid-client-authn-js/blob/1cfabd03eab6a3d2a3859144e69ae1a7b880d1ef/packages/browser/src/sessionInfo/SessionInfoManager.ts#L43

packages/browser/src/Session.ts
NSeydoux: Well there is, but it's only on the `logout` code path. When logging in, the used fetch function is
```
const fetchWithCookies = (
  info: RequestInfo,
  init?: RequestInit
): ReturnType<typeof fetch> => {
  return window.fetch(info, {
    ...init,
    credentials: "include",
  });
};
```, 
because (for now) we rely on a RS cookie not to get logged out on refresh. So we do override a parameter, but `window.fetch` isn't rebound... Would you have a pointer to a read that would explain what your solution solves, so that I can see how it could apply here ?

**Comment by RubenVerborgh** [comment]
Suggestion: do _not_ allow global usage of `fetch` as you do now; disallow it in the linter. Instead, have one file (where you ignore that linter rule) export a version of `fetch` bound to `global`, and use that one everywhere.

File: packages/browser/__tests__/Session.spec.ts (hunk 2 above)
Comment (NSeydoux) [reply packages/browser/__tests__/Session.spec.ts:158]: Excellent, thanks ! Done in e646adbe43aa0575c16b0f1c54f81830a59ae993

File: packages/browser/__tests__/Session.spec.ts (hunk 3 above)
Comment (NSeydoux) [reply packages/browser/__tests__/Session.spec.ts:154]: e646adbe43aa0575c16b0f1c54f81830a59ae993

File: packages/browser/src/sessionInfo/SessionInfoManager.ts [hunk 4]
```diff
@@ -33,14 +33,15 @@ import {
...
+import { fetchWithCookies } from "../ClientAuthentication";
 
 export function getUnauthenticatedSession(): ISessionInfo & {
   fetch: typeof fetch;
 } {
   return {
     isLoggedIn: false,
     sessionId: v4(),
-    fetch,
+    fetch: fetchWithCookies,
```
RubenVerborgh, line 44: Just checking if `fetchWithCookies` is what you want here; the original one was a regular fetch.

NSeydoux:
Thanks so much for that, I don't think I was going to resolve this properly before a while ^^. I think I'll merge the quick fix right now so that I can release a patched version, and then implement the linter rule you suggest to prevent regression in a future commit.

RubenVerborgh (approved):
Looks good; also appreciate the follow-up on linter.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ deprecated, and replaced by `storage`.
 - When providing a storage that already contains session information in the NodeJS
 context (as demonstrated with the `FileSystemStorage` in `packages/node/example/bootstrappedApp`),
 the refresh token is now picked up instead of the request failing.
+- Calling `Session::fetch` before logging the Session in threw an error.
```
Vinnl, line 23: Note: this changelog is shared between all published packages, so probably a good idea to split the notes by affected package too.

Vinnl (approved):
+1 on the question about `fetchWithCookies` for the unauthenticated fetcher, otherwise LGTM.

packages/browser/src/sessionInfo/SessionInfoManager.ts
NSeydoux: The previous workaround that prevented the bug from manifesting (wrongfully removed in https://github.com/inrupt/solid-client-authn-js/commit/520a19bdf9ca2d4e6e7e377a3dcf24053cf4ba70#diff-4f6320deb9e43b5216428fb7c2dd230f4b0c5325c6f147f6a73b15659c355fb9L132, I must have messed up testing then) did include the cookie, so that you are still able to access private resources on refresh. This is a stopgap solution that only works against ESS, and has known limitations, which is why we will replace it as soon as possible with a proper fix to cleanly resolve the refresh issue in a way that's independent f the resource server, and inline with the OIDC philosophy.

PR #948, opened 2021-01-14
Node coverage almost 100% - only SessionManager left

pmcb55:
Tech debt cleanup - improving code coverage for Node package.

The only file left is SessionManager (which is up to about 60%), but we're looking at re-working that, so finishing here.


NSeydoux:
Since this commit doesn't change anything about the dependencies, it should not change the `package-lock.json`. Could you revert it to its previous state ? And in the future, can you check that a `package-lock` isn't commited if you're not updating the dependencies ?

pmcb55:
Ok - package-locks removed.

NSeydoux (approved):

PR #949, opened 2021-01-14
Move reference to `window` into a getter

ajacksified:
This prevents the session constructor from throwing errors, which is especially useful in test frameworks which do not stub window, and in SSR contexts such as Next.js which hand control to a single page app.

This specifically fixes the use-case in which the Session context provider is used from the React SDK, which calls a Session constructor; that constructor throws errors in the initial Node rendering of the page before the single-page-app takes over.


NSeydoux (approved):

NSeydoux:
Haha you rebased before I had the chance to :). Should this figure in the CHANGELOG as a bugfix ?

PR #961, opened 2021-01-18
Only emit login event when session info is set

### LINKED ISSUE #955: session.info.webId not available in onLogin callback
When I call `session.info.webId` in the onLogin (https://github.com/inrupt/solid-client-authn-js/blob/28412f4fdc74500b0612cd4ec8780c20e5ccc33f/packages/browser/src/Session.ts#L240) callback, I expect to get webId of the logged in user. However I get 'undefined':

```
readonly session = new Session();
webId: string;

async handleRedirectAfterLogin() {
  this.session.onLogin(() => {
    this.webId = this.session.info.webId;
    this._snackBar.open('Successfully logged in.' + this.webId, 'Dismiss');
  });
  await this.inruptService.session.handleIncomingRedirect(window.location.href);
}
```

I am using angular with the solid-client-authn-browser library:
```
    "@inrupt/solid-client": "^1.3.0",
    "@inrupt/solid-client-authn-browser": "^1.4.1",
    "@inrupt/vocab-common-rdf": "^0.7.2",
```

Am I doing something wrong? How else to get webId reliably after login?

NSeydoux:
Resolves #955 

The login even was emitted *before* setting the session info, which
could result in a race where the onLogin callback couldn't read session
information, surch as the WebId.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,7 @@ The following sections document changes that have been released already:
 
 #### browser and node
 
-- The `login` even was emitted before setting the session info, which
-could result in a race where the onLogin callback couldn't read session
-information, surch as the WebID.
+- The `onLogin` callback couldn't read session information, such as the WebID.
 
 ## 1.4.1 - 2020-01-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following sections document changes that have been released already:
 
-N/A
+### Bugfix
+
+#### browser and node
+
+- The `login` even was emitted before setting the session info, which
+could result in a race where the onLogin callback couldn't read session
+information, surch as the WebID.
 
 ## 1.4.1 - 2020-01-14
 
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -438,15 +438,8 @@ describe("Session", () => {
   });
 
   describe("onLogin", () => {
-    // The `done` callback is used in order to make sure the callback passed to
-    // onLogin is called.
-    // eslint-disable-next-line jest/no-done-callback
-    it("calls the registered callback on login", async (done) => {
-      const myCallback = jest.fn((): void => {
-        if (done) {
-          done();
-        }
-      });
+    it("calls the registered callback on login", async () => {
+      const myCallback = jest.fn();
       const clientAuthentication = mockClientAuthentication();
       clientAuthentication.handleIncomingRedirect = jest.fn(
         async (_url: string) => {
@@ -504,6 +497,8 @@ describe("Session", () => {
       mySession.onLogin(myCallback);
       await mySession.handleIncomingRedirect("https://some.url");
       expect(myCallback).toHaveBeenCalled();
+      // Verify that the conditional assertion has been called
+      expect.assertions(2);
     });
   });
 
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -442,13 +442,11 @@ describe("Session", () => {
     // onLogin is called.
     // eslint-disable-next-line jest/no-done-callback
     it("calls the registered callback on login", async (done) => {
-      let hasBeenCalled = false;
-      const myCallback = (): void => {
-        hasBeenCalled = true;
+      const myCallback = jest.fn((): void => {
         if (done) {
           done();
         }
-      };
+      });
       const clientAuthentication = mockClientAuthentication();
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following sections document changes that have been released already:
 
-N/A
+### Bugfix
+
+#### browser and node
+
+- The `login` even was emitted before setting the session info, which
```
Vinnl, line 17:
```suggestion
- The `login` event was emitted before setting the session info, which
```

Though maybe there's not a need to go in so much detail - just a "session information sometimes wasn't available when `onLogin` was triggered". But either way :)

packages/browser/__tests__/Session.spec.ts
```diff
@@ -487,6 +487,28 @@ describe("Session", () => {
...
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      const myCallback = (): void => {
+        hasBeenCalled = true;
+        expect(mySession.info.webId).toBe("https://some.webid#them");
+      };
+      mySession.onLogin(myCallback);
+      await mySession.handleIncomingRedirect("https://some.url");
+      expect(hasBeenCalled).toEqual(true);
```
Vinnl, line 510: If you make this a mock function you don't need to manually track calls, I think. Also, be sure to add an `expect.assertions(2)` to make sure the conditional `expect` is actually called.

```suggestion
      const myCallback = jest.fn((): void => {
        expect(mySession.info.webId).toBe("https://some.webid#them");
      });
      mySession.onLogin(myCallback);
      await mySession.handleIncomingRedirect("https://some.url");
      expect(myCallback).toHaveBeenCalledTimes(1);
```

packages/node/src/Session.spec.ts
```diff
@@ -316,6 +316,31 @@ describe("Session", () => {
         mySession.handleIncomingRedirect("https://some.url")
       ).resolves.not.toThrow();
     });
+
+    it("sets the appropriate information before calling the callback", async (done) => {
```
Vinnl, line 320: I'm not sure if this check is just triggered by the name "callback" in the description, but best to deal with it before merging in.

image

CHANGELOG.md
```diff
@@ -10,7 +10,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+could result in a race where the onLogin callback couldn't read session
+information, surch as the WebID.
```
pmcb55, line 19:
```suggestion
- The `login` event was emitted before setting the session info, which
could result in a race condition where the `onLogin` callback couldn't read
the session information, such as the WebID.
```

packages/node/src/Session.spec.ts
```diff
@@ -316,6 +316,31 @@ describe("Session", () => {
...
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      const myCallback = (): void => {
+        hasBeenCalled = true;
+        expect(mySession.info.webId).toBe("https://some.webid#them");
+        if (done) {
```
Vinnl, line 336: Likewise about the mock function and `expect.assertions` - and is this `done()` call really necessary here?

Vinnl (approved):

pmcb55:
There appears to be a scary degree of duplication here between the Browser and Node implementations - is they any chance this code (and test code) could be refactored into the `core` package instead...?

packages/node/src/Session.spec.ts
NSeydoux: That's a good point. I initially assumed (when implementing the tests back on `solid-auth-fetcher`) that it would be necessary to do so to prevent trhe test from completing before the callback is called, but it appears that it is not the case. Fixed in 80a6e1d11ae1bff9b4437654f60b9193209820e0

packages/node/src/Session.spec.ts
NSeydoux: 80a6e1d11ae1bff9b4437654f60b9193209820e0

packages/browser/__tests__/Session.spec.ts
NSeydoux: `expect.assertions()` is added in 059619d999004cd0bae8abeb5ac91d0fba343a80

CHANGELOG.md
NSeydoux: Good point, I copy-pasted the commit message, but it actually makes sense that the changelog entry only summarizes the issue from a user perspective, and not details the internals of the bug.

NSeydoux:
Unfortunately, these two parts of the code are dependant on both code from the `core` package, but also of code from the environment-specific packages. Refactoring it should be possible without introducing cyclic dependencies using some inversion of control, but that would still be a non-trivial task.

PR #962, opened 2021-01-18
Feat/1644 multi session/get session from storage

NSeydoux:
getStoredSession is a function that will return a Session based on its
session ID if the provided ID exists in storage. It is mostly useful to
handle multiple sessions on the server side, where the session object
will not always be available in the scope of the function that processes
some data.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The following sections document changes that have been released already:
 
 #### node
 
-- Add a `getStoredSession` function to retrieve a session from storage based on
+- `getStoredSession`: a function to retrieve a session from storage based on
 its session ID (for multi-session management).
 
 ## 1.4.1 - 2020-01-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,13 @@ The following sections document changes that have been released already:
 
 - The `onLogin` callback couldn't read session information, such as the WebID.
 
+### New features
+
+#### node
+
+- Add a `getStoredSession` function to retrieve a session from storage based on
+its session ID (for multi-session management).
+
 ## 1.4.1 - 2020-01-14
 
 ### Backward-compatible API changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following sections document changes that have been released already:
 
+### New features
+
+#### node
+
+- `getStoredSession`: a function to retrieve a session from storage based on
+its session ID (for multi-session management).
+
 ## 1.4.2 - 2020-01-19
 
 ### Bugfix
@@ -17,13 +24,6 @@ The following sections document changes that have been released already:
 #### browser and node
 
 - The `onLogin` callback couldn't read session information, such as the WebID.
-
-### New features
-
-#### node
-
-- `getStoredSession`: a function to retrieve a session from storage based on
-its session ID (for multi-session management).
 
 ## 1.4.1 - 2020-01-14
 
--- a/packages/core/src/sessionInfo/ISessionInfo.ts
+++ b/packages/core/src/sessionInfo/ISessionInfo.ts
@@ -20,7 +20,8 @@
  */
 
 /**
- * Defines the data that should be persisted for each session.
+ * Defines the data that should be persisted for each session. This interface
+ * is exposed as part of our public API.
  */
 export interface ISessionInfo {
   /**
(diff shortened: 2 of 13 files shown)


packages/core/src/sessionInfo/ISessionInfoManager.ts
```diff
@@ -39,7 +39,9 @@ export interface ISessionInfoManagerOptions {
  */
 export interface ISessionInfoManager {
   update(sessionId: string, options: ISessionInfoManagerOptions): Promise<void>;
-  get(sessionId: string): Promise<ISessionInfo | undefined>;
-  getAll(): Promise<ISessionInfo[]>;
+  get(
```
Vinnl, line 42: I'm not 100% on how/if this is exposed to the developer, but does it not potentially make more sense to have separate getters for `ISessionInternalInfo`?

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@ The following sections document changes that have been released already:
 
 - The `onLogin` callback couldn't read session information, such as the WebID.
 
+### New features
+
+#### node
+
+- Add a `getStoredSession` function to retrieve a session from storage based on
```
Vinnl, line 23:
```suggestion
- `getStoredSession`: a function to retrieve a session from storage based on
```

packages/node/src/Session.spec.ts
```diff
@@ -350,3 +363,87 @@ describe("Session", () => {
...
+      .fn()
+      .mockReturnValue(clientAuthentication);
+    const mySession = await getStoredSession("mySession", mockStorage({}));
+    expect(mySession?.info).toEqual(
+      expect.objectContaining({
+        webId: "https://my.webid",
+        isLoggedIn: true,
+        sessionId: "mySession",
+      })
+    );
```
Vinnl, line 393: Are any other properties on this object valid? Or might it be better to check for the exact object, to ensure that e.g. the refresh token and issuer are not added to it?

```suggestion
    expect(mySession?.info).toStrictEqual(
      {
        webId: "https://my.webid",
        isLoggedIn: true,
        sessionId: "mySession",
      }
    );
```

(And likewise for the other instances of `expect.objectContaining`.)

Vinnl (approved):
LGTM, but just wondering if maybe the name would be better as `getSessionFromStorage`, since you have to pass it the storage yourself?

packages/core/src/sessionInfo/ISessionInfoManager.ts
NSeydoux: It isn't exposed to developers at all, this is part of the purely internal API. Basically, the only difference between the two functions would be the set of keys they access in memory, but I don't think I would add a separate function for that. One change that could be possible is to have the type `ISessionInternalInfo` include the type `ISessionInfo`, and only use the latter as part of the public API. We could then have a `getPublic` function to get `ISessionInfo`, and `getinternal` to get all the values in memory. What do you think ?

Vinnl: Effectively we already only use `ISessionInfo` as part of the public API, right (and the union type only in internal API's). So in that case I think no changes are necessary.

packages/node/src/Session.spec.ts
NSeydoux: You are right, here we not only check that the values we want are included, but also that the ones we don't aren't present. Updated in ee3c9b9b9c2c65d263f05f07c6dabfc03a45875a.

packages/core/src/sessionInfo/ISessionInfo.ts
```diff
@@ -39,3 +39,15 @@ export default interface ISessionInfo {
    */
   sessionId: string;
 }
+
+export interface ISessionInternalInfo {
```
pmcb55, line 43: This interface could maybe use a comment to explain why it's needed, and why it's explicitly for internal use only...?

packages/node/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -111,26 +112,35 @@ export class SessionInfoManager implements ISessionInfoManager {
...
     const webId = await this.storageUtility.getForUser(sessionId, "webId");
     const isLoggedIn = await this.storageUtility.getForUser(
       sessionId,
       "isLoggedIn"
     );
+    const refreshToken = await this.storageUtility.getForUser(
+      sessionId,
+      "refreshToken"
+    );
+    const issuer = await this.storageUtility.getForUser(sessionId, "issuer");
```
pmcb55, line 127: Just a thought (for later), but would it make sense to use vocab terms (i.e., IRIs) for the 'keys' that appear here (and for all the localStorage keys we use throughout our code), i.e., `webId`, `refreshToken` and `issuer`? It would make the keys in storage a few bytes longer, but would make them completely globally unique, and more importantly, self-descriptive (i.e., dereference the IRI to see the `rdfs:comments` (in multiple languages!), and any other meta-data related to that term)? Would also mean we'd get rid of all these nasty magic strings...
```suggestion
    const webId = await this.storageUtility.getForUser(sessionId, "webId");
    const isLoggedIn = await this.storageUtility.getForUser(
      sessionId,
      "isLoggedIn"
    );
    const refreshToken = await this.storageUtility.getForUser(
      sessionId,
      "refreshToken"
    );
    const issuer = await this.storageUtility.getForUser(sessionId, "issuer");
```

pmcb55 (approved):

PR #968, opened 2021-01-19
add current URL as default

pmcb55:
# New feature description
On the handleIncomingRedirect method of the Session class, set
the current browser URL as the default value, just as a
convenience for the developer.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+- Just as a convenience for the developer, on the `handleIncomingRedirect`
+  method of the `Session` class in the browser package, set the current browser
+  URL as the default value.
+
 The following sections document changes that have been released already:
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
-- Just as a convenience for the developer, on the `handleIncomingRedirect`
-  method of the `Session` class in the browser package, set the current browser
-  URL as the default value.
+
+### New features
+
+#### browser
+
+- `handleIncomingRedirect` uses the current browser URL as a default value.
 
 The following sections document changes that have been released already:
 
--- a/packages/browser/__tests__/ClientAuthentication.spec.ts
+++ b/packages/browser/__tests__/ClientAuthentication.spec.ts
@@ -146,7 +146,7 @@ describe("ClientAuthentication", () => {
         "https://coolapp.com/redirect?state=userId&id_token=idToken&access_token=accessToken";
       await clientAuthn.handleIncomingRedirect(url);
 
-      // Calling handleredirect should give us an authenticated fetch.
+      // Calling handle redirect should give us an authenticated fetch.
       expect(clientAuthn.fetch).not.toBe(unauthFetch);
 
       await clientAuthn.logout("mySession");
@@ -198,7 +198,7 @@ describe("ClientAuthentication", () => {
       });
       expect(defaultMocks.redirectHandler.handle).toHaveBeenCalledWith(url);
 
-      // Calling handleredirect should have updated the fetch.
+      // Calling handle redirect should have updated the fetch.
       expect(clientAuthn.fetch).not.toBe(unauthFetch);
     });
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+- Just as a convenience for the developer, on the `handleIncomingRedirect`
+  method of the `Session` class in the browser package, set the current browser
+  URL as the default value.
```
NSeydoux, line 13:
```suggestion

### New features

#### browser

- `handleIncomingRedirect` uses the current browser URL as a default value.
```

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -146,7 +146,7 @@ describe("ClientAuthentication", () => {
         "https://coolapp.com/redirect?state=userId&id_token=idToken&access_token=accessToken";
       await clientAuthn.handleIncomingRedirect(url);
 
-      // Calling handleredirect should give us an authenticated fetch.
+      // Calling handle redirect should give us an authenticated fetch.
```
NSeydoux, line 149:
```suggestion
      // Calling `handleRedirect` should give us an authenticated fetch.
```

packages/browser/__tests__/Session.spec.ts
```diff
@@ -162,17 +162,62 @@ describe("Session", () => {
...
+        window.history.replaceState = replaceState;
+      }
+    });
+
     it("wraps up ClientAuthentication handleIncomingRedirect", async () => {
       // eslint-disable-next-line no-restricted-globals
       history.replaceState = jest.fn();
       const clientAuthentication = mockClientAuthentication();
-      const clientAuthnHandle = jest.spyOn(
+      const clientAuthnHandler = jest.spyOn(
```
NSeydoux, line 214: This was named `clientAuthnHandle` because it spies on the `handle` function, and isn't a "handler" per se. Why rename it ?

packages/browser/__tests__/Session.spec.ts
```diff
@@ -268,7 +313,7 @@ describe("Session", () => {
 
     it("preserves a binding to its Session instance", async () => {
       const clientAuthentication = mockClientAuthentication();
-      const clientAuthnHandleIncomingRedirect = jest.spyOn(
+      const clientAuthnHandlerIncomingRedirect = jest.spyOn(
```
NSeydoux, line 316: ditto

NSeydoux (approved):
It all looks goo, just a few minor comments on some names.

packages/browser/__tests__/ClientAuthentication.spec.ts
pmcb55: No, I'm trying to fix this. Basically it _seems like_ the original comment was trying to refer to the explicit function name (I think, it wasn't in quotes or anything, and it also wasn't camel-cased, so either way it looked like a typo!), but then the function name seems to have changed (i.e., from `handleRedirect()` to `handleIncomingRedirect()`), at which point the comment got out-of-sync with the code, which is always confusing, and therefore a big reason to try to never refer to things explicitly in comments. So I changed that to just be 'handle redirect' to try and reflect in English the behaviour of the code, but without naming any functions explicitly. Does that make sense now, i.e., we should leave the original change I made)...?

packages/browser/__tests__/Session.spec.ts
pmcb55: Yeah, I was in two minds here - I don't really like the original name, but I didn't want to change it too much. I would say any `handleXXXX()` function is by definition as 'handler', and so this code is spying on a handler (and not a handle). Plus a 'handle' has other connotations. But overall maybe I'll just rename this `incomingRedirectHandler`, i.e., is the `clientAuthn` prefix really relevant here...?

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: I understand the resoning, and it makes sense, but then the phrase "Calling handle redirect" is a bit weird, isn't it ? Couldn't it just be "Handling the redirect" or something ? Anyway, if the sentence is correct in English, I'm happy with your original suggestion.

pmcb55: Yeah, I was just trying to tweak the original as little as possible, but I think "Calling the redirect handler..." is better again - I'll change it to that: 9165db5

packages/browser/__tests__/Session.spec.ts
pmcb55: Yeah, I went with `incomingRedirectHandler`: 7f8d4af

packages/browser/__tests__/Session.spec.ts
pmcb55: Changed in: 7f8d4af

PR #970, opened 2021-01-19
Feat/1644 multi session/get session from storage all

NSeydoux:
This implements the different functions required to list all the known session IDs for a given storage, in order to be able to iterate through them is required. Note that `getSessionFromStorage` returns a Session object, potentially logged in, while `getSessionIdFromStorageAll` only returns the session ids, and it is up to the caller to the manually retrieve individual sessions. This is mostly for performance concerns: logging a session in requires an HTTP interaction, so if multiple sessions need to be logged in at once, that could represent a costly operation (that the server is still able to perform with the proposed API anyway). 

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,9 @@ The following sections document changes that have been released already:
 
 - `getSessionFromStorage`: a function to retrieve a session from storage based on
 its session ID (for multi-session management).
-- `getStoredSessionIdAll`: a function to retrieve the session IDs for all stored
+- `getSessionIdFromStorageAll`: a function to retrieve the session IDs for all stored
 sessions.
-- `clearSessionAll`: a function to clear all information about all sessions in
+- `clearSessionFromStorageAll`: a function to clear all information about all sessions in
 storage.
 
 ## 1.4.2 - 2020-01-19
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ The following sections document changes that have been released already:
 
 - `getSessionFromStorage`: a function to retrieve a session from storage based on
 its session ID (for multi-session management).
+- `getStoredSessionIdAll`: a function to retrieve the session IDs for all stored
+sessions.
+- `clearSessionAll`: a function to clear all information about all sessions in
+storage.
 
 ## 1.4.2 - 2020-01-19
 
--- a/packages/browser/src/sessionInfo/SessionInfoManager.ts
+++ b/packages/browser/src/sessionInfo/SessionInfoManager.ts
@@ -170,4 +170,26 @@ export class SessionInfoManager implements ISessionInfoManager {
   async clear(sessionId: string): Promise<void> {
     return clear(sessionId, this.storageUtility);
   }
+
+  /**
+   * Registers a new session, so that its ID can be retrieved.
+   * @param sessionId
+   */
+  async register(_sessionId: string): Promise<void> {
+    throw new Error("Unimplemented");
+  }
+  /**
+   * Returns all the registered session IDs. Differs from getAll, which also
+   * returns additional session information.
+   */
+  async getRegisteredSessionIdAll(): Promise<string[]> {
+    throw new Error("Unimplemented");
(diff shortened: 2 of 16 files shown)


packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -170,4 +170,27 @@ export class SessionInfoManager implements ISessionInfoManager {
   async clear(sessionId: string): Promise<void> {
     return clear(sessionId, this.storageUtility);
   }
+
+  /**
+   * Registers a new session, so that its ID can be retrieved.
+   * @param sessionId
+   */
+  async register(_sessionId: string): Promise<void> {
+    throw new Error("Unimplemented");
```
Vinnl, line 179: Is this because of a shared interface for something that we don't want/need to implement in the browser?

If so: ugh... Please can we just split the two projects? :P

packages/node/src/ClientAuthentication.ts
```diff
@@ -59,6 +59,8 @@ export default class ClientAuthentication {
     // could return to a previous redirect URL that contains OIDC params that
```
Vinnl, line 59: Not part of this PR, but there's a comment here mentioning the browser even though this is the Node package :)

packages/node/src/__mocks__/ClientAuthentication.ts
```diff
@@ -32,3 +45,21 @@ export const mockClientAuthentication = (): ClientAuthentication =>
...
+  logoutHandler: ILogoutHandler;
+};
+
+export const mockCustomClientAuthentication = (
+  mocks: Partial<CustomMocks>
+): ClientAuthentication =>
+  new ClientAuthentication(
+    mocks.loginHandler ?? LoginHandlerMock,
+    mocks.redirectHandler ?? RedirectHandlerMock,
+    mocks.logoutHandler ?? mockLogoutHandler(mocks.storage!),
```
Vinnl, line 63: A bunch of warnings here about the non-null assertion, so either explicitly initialise it or add an ESLint ignore comment. (Likewise below.)

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ The following sections document changes that have been released already:
 
 - `getSessionFromStorage`: a function to retrieve a session from storage based on
 its session ID (for multi-session management).
+- `getStoredSessionIdAll`: a function to retrieve the session IDs for all stored
+sessions.
+- `clearSessionAll`: a function to clear all information about all sessions in
```
Vinnl, line 28:
```suggestion
- `getSessionIdFromStorageAll`: a function to retrieve the session IDs for all stored
sessions.
- `clearSessionFromStorageAll`: a function to clear all information about all sessions in
```

Vinnl (approved):

packages/browser/src/sessionInfo/SessionInfoManager.ts
NSeydoux: I'm afraid so ^^.

packages/node/src/__mocks__/ClientAuthentication.ts
NSeydoux: Weird, ESlint did not complain (that I realised), only the GH static code analysis tool takes issue with this. I'll see how to fix this.

Vinnl: I think ESLint is still only set to warnings, so it's probably somewhere in your console's logs :)

NSeydoux: 4f6ba7c2594e2dccbf474288cb6e68734c2a6d7a

NSeydoux: Oh yeah looking for them (in the large output of all the tests ran by lerna ^^). Should it be an error ?

Vinnl: I think it might be good to turn warnings into an error, but I think there are so many of them that that's probably best done separately.

PR #976, opened 2021-01-20
Feat/1644 multi session/demo app

NSeydoux:
This PR does two things:
1. implement a sample app demonstrating multi-session management
2. Fix a bug regarding the way the default storage was used in Node: When not specifying a storage for
getClientAuthenticationWithDependencies, a default storage is
initialized. However, if we want users to be able to retrieve a session
from the default storage in node, this default storage should be a
singleton shared between all instances, and not a storage that gets
reinitialized each time a session is searched for.

# Checklist


CHANGELOG.md
```diff
@@ -15,10 +15,6 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - `handleIncomingRedirect` uses the current browser URL as a default value.
 
-The following sections document changes that have been released already:
```
Vinnl, line 18: Heh, how did this happen? :smile:

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+    sessionIds.map(async (sessionId) => {
+      return await getSessionFromStorage(sessionId);
+    })
+  );
+  const htmlSessions =
+    sessions.reduce((sessionList, session) => {
+      if (session?.info.isLoggedIn) {
+        return sessionList + `<li>${session?.info.webId}</li>`;
+      }
+      return "Anonymous";
```
Vinnl, line 60:
```suggestion
      return sessionList + "<li>Anonymous</li>";
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+    }, "<ul>") + "</ul>";
+  res
+    .writeHead(200, { "Content-Type": "text/html" })
+    .write(
+      `<p>There are currently ${sessionIds.length} visitors: ${htmlSessions}</p>`
+    );
+  res.end();
```
Vinnl, line 67: I don't think all this ceremony is necessary: https://expressjs.com/en/4x/api.html#res.send

> When the parameter is a String, the method sets the Content-Type to “text/html”:

Likewise for the other `res.write` instances that send HTML.

```suggestion
  res.send(`<p>There are currently ${sessionIds.length} visitors: ${htmlSessions}</p>`);
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+  req.session!.sessionId = session.info.sessionId;
+  await session.login({
+    redirectUrl: "http://localhost:3001/redirect",
+    oidcIssuer: DEFAULT_OIDC_ISSUER,
+    clientName: clientApplicationName,
+    handleRedirect: (data) => {
+      res.writeHead(302, {
+        location: data,
+      });
+      res.end();
```
Vinnl, line 81: This too can be simplified I think: https://expressjs.com/en/4x/api.html#res.redirect

```suggestion
      res.redirect(data);
```

(And I suppose `redirectUrl` would be a more descriptive name than `data`, or are we expecting other data there as well?)

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+      .write("<p>Expected a 'resource' query param</p>");
+  } else if (session) {
+    res
+      .writeHead(200, { "Content-Type": "text/html" })
+      .write(
+        "<pre>" +
+          (
+            await (await session.fetch(req.query["resource"] as string)).text()
+          ).replace(/</g, "&lt;") +
+          "</pre>"
```
Vinnl, line 127: Maybe it'd be better to just pipe the return value of `fetch` directly to the user, rather than injecting it into an HTML page?

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+  res.end();
+});
+
+app.get("/fetch", async (req, res, next) => {
+  const session = await getSessionFromStorage(req.session!.sessionId);
+  if (!req.query["resource"]) {
+    res
+      .writeHead(400, { "Content-Type": "text/html" })
+      .write("<p>Expected a 'resource' query param</p>");
+  } else if (session) {
```
Vinnl, line 119: Feels like everything after the `if` can be compressed with a `const fetch = (session ?? new Session()).fetch`.

Vinnl (approved):
Sweet, I didn't realise that the storage was optional for `get*FromStorage()`, but that looks pretty ergonomic.

packages/node/example/multiSession/README.md
```diff
@@ -0,0 +1,27 @@
...
+To install this demo app, first run `npm ci` at the root of the repository, which
+will trigger the build of the libraries our app depends on. Then, run `npm ci` 
+again, this time in `packages/node/example/multiSession/`, and then `npm run build`.
+You should be all set. If you change the code, make sure to re-run `npm run build`
+before you re-run it.
+
+## Running the app
+
+Running `node dist/serverSideApp.js` starts the server. Four endpoints are then available:
+- `/login`, to initiate the login process against a solid Identity Provider
```
pmcb55, line 22:
```suggestion
- `/login`, to initiate the login process against a Solid Identity Provider 
```

packages/node/example/multiSession/README.md
```diff
@@ -0,0 +1,27 @@
...
+before you re-run it.
+
+## Running the app
+
+Running `node dist/serverSideApp.js` starts the server. Four endpoints are then available:
+- `/login`, to initiate the login process against a solid Identity Provider 
+(by default, `https://broker.pod.inrupt.com`)
+- `/redirect`, where the Solid Identity Provider will redirect users after login
+- `/fetch`, where you have to pass a `resource=<some resource>` query param in 
+order to fetch a resource
```
pmcb55, line 26:
```suggestion
order to fetch a resource (protected resources will require you be logged in and to have been granted READ access to that resource, whereas public resources can be fetched whether you're logged in or not)
```

CHANGELOG.md
NSeydoux: I think there was an unfortunate conflict resolution ^^

packages/node/example/multiSession/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "author": "Inrupt, Inc.",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/inrupt/solid-client-authn-js/issues"
+  },
+  "homepage": "https://github.com/inrupt/solid-client-authn-js#readme",
+  "dependencies": {
+    "cookie-session": "^1.4.0",
+    "express": "^4.17.1"
+  },
```
pmcb55, line 22: Just curious - but is this really the intent here - i.e., to not depend on SCAN at all, but instead import directly from the 'dist' code using `import x from '../../dist/y';`...? I guess that makes it easy to test code changes locally, but slightly awkward to lift this demo code and re-use it for my own NodeJS server...
Would it warrant a mention in the README - i.e., "This demo imports code from SCAN directly rather than depending on a deployed npm module. This allows you to..., but means..." ...?

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// The only import we need from the Node AuthN library is the Session class.
```
pmcb55, line 22: Yikes! Comment mentioning a class name explicitly, and going out-of-sync with the code, already :) !
```suggestion
// We only need the following imports from the Node AuthN library.
```

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: That does look better a302bfe10edcc3676f3ec0e67e332a6082f4addf

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,164 @@
...
+const app = express();
+const PORT = 3001;
+
+const DEFAULT_OIDC_ISSUER = "https://broker.pod.inrupt.com/";
+
+app.use(
+  cookieSession({
+    name: "session",
+    // These keys are required by cookie-session to sign the cookies.
+    keys: ["key1", "key2"],
```
pmcb55, line 43: So are the values of these keys relevant now? If not, I'd use their values to explain that they are required, but not used in this demo, and so their values are irrelevant
```suggestion
    // These keys are required by cookie-session to sign the cookies, but we don't use that signing feature in this demo.
    keys: ["Required, but value not relevant - key1", "Required, but value not relevant - key2"],
```

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: Indeed 6168832e4ece408a454a56b0c83eae428491240d

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: Because of the `<url>` that are interpreted as HTML, Turtle looks quite bad if it isn't processed a bit.

Vinnl: But why have it interpreted as HTML? Can't we just pass on the Content-Type as well?

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: Express does sign the cookies though, but it's all handled by `cookie-session`

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+  const htmlSessions =
+    sessions.reduce((sessionList, session) => {
+      if (session?.info.isLoggedIn) {
+        return sessionList + `<li>${session?.info.webId}</li>`;
+      }
+      return sessionList + "<li>Anonymous</li>";
+    }, "<ul>") + "</ul>";
+  res.send(
+    `<p>There are currently ${sessionIds.length} visitors: ${htmlSessions}</p>`
```
pmcb55, line 60:
```suggestion
    `<p>There are currently [${sessionIds.length}] visitors: ${htmlSessions}</p>`
```

packages/node/example/multiSession/package.json
NSeydoux: Yeah, directly importing the code from the `dist` folder makes it much easier to use with local changes, without needing to publsh the package on verdaccio or NPM.But you are right, that means one cannot copy-paste and run this code in an environment where `solid-client-authn-node` isn't available. i'll add a comment directly in the code in the imports section to specify this.

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// We only need the following imports from the Node AuthN library.
```
pmcb55, line 19: Missing end of comment here (only spotted this when I tried to clone and build the code - 'cos the compiler complains with:
```
src/serverSideApp.ts:129:1 - error TS1010: '*/' expected.
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+    res.send(`<p>Already logged in with WebID: [${session.info.webId}]</p>`);
+  }
+});
+
+app.get("/redirect", async (req, res) => {
+  const session = await getSessionFromStorage(req.session!.sessionId);
+  if (session === undefined) {
+    res
+      .status(400)
+      .send(`<p>No session stored for ID ${req.session!.sessionId}</p>`);
```
pmcb55, line 83:
```suggestion
      .send(`<p>No session stored for ID [${req.session!.sessionId}]</p>`);
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+  res.send(
+    `<p>There are currently ${sessionIds.length} visitors: ${htmlSessions}</p>`
+  );
+});
+
+app.get("/login", async (req, res, next) => {
+  const session = new Session();
+  req.session!.sessionId = session.info.sessionId;
+  await session.login({
+    redirectUrl: "http://localhost:3001/redirect",
```
pmcb55, line 68: I'd move this to a constant at the top of the file, and then add a comment explaining it, something like:
```suggestion
// This is the endpoint our NodeJS demo app listens on to receive incoming login redirect requests (i.e., as part of the standard OIDC login flow).
const DEMO_APP_LOGIN_REDIRECT_ENDPOINT = "http://localhost:3001/redirect";

    redirectUrl: DEMO_APP_LOGIN_REDIRECT_ENDPOINT,
```

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: Oh yeah I see what you mean, I missed your point initially. Unfortunately, for `Content-Type: text/turtle`, my browser doesn't display it, it opens a popup to open it with an application. i guess it could be configured otherwise, but for simplicity think wrapping the content in `<pre>` tags is still a good option.

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+      res.status(400).send(`<p>Not logged in after redirect</p>`);
+    }
+  }
+  res.end();
+});
+
+app.get("/fetch", async (req, res, next) => {
+  const session = await getSessionFromStorage(req.session!.sessionId);
+  if (!req.query["resource"]) {
+    res.status(400).send("<p>Expected a 'resource' query param</p>");
```
pmcb55, line 98: Would be nice to explain this a little, and offer an example (as anyone trying this would, I suspect, first try with no query param at all):
```suggestion
    res.status(400).send("<p>Expected a 'resource' query param, for example <strong>http://localhost:3001/fetch?resource=https://pod.inrupt.com/MY_USERNAME/</strong> to fetch the resource at the root of your Pod (which, by default, only <strong>you</strong> will have access to).</p>");
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+app.get("/redirect", async (req, res) => {
+  const session = await getSessionFromStorage(req.session!.sessionId);
+  if (session === undefined) {
+    res
+      .status(400)
+      .send(`<p>No session stored for ID ${req.session!.sessionId}</p>`);
+  } else {
+    await session.handleIncomingRedirect(getRequestFullUrl(req));
+    if (session.info.isLoggedIn) {
+      res.send(`<p>Logged in as [${session.info.webId}] after redirect</p>`);
```
pmcb55, line 87:
```suggestion
      res.send(`<p>Logged in as [<strong>${session.info.webId}</strong>] after redirect</p>`);
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+  const sessionIds = await getSessionIdFromStorageAll();
+  const sessions = await Promise.all(
+    sessionIds.map(async (sessionId) => {
+      return await getSessionFromStorage(sessionId);
+    })
+  );
+  const htmlSessions =
+    sessions.reduce((sessionList, session) => {
+      if (session?.info.isLoggedIn) {
+        return sessionList + `<li>${session?.info.webId}</li>`;
```
pmcb55, line 55:
```suggestion
        return sessionList + `<li><strong>${session?.info.webId}</strong></li>`;
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+      }
+      return sessionList + "<li>Anonymous</li>";
```
pmcb55, line 57:
```suggestion
      return sessionList + "<li><strong>Anonymous</strong></li>";
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+        "</pre>"
+    );
+  }
+});
+
+app.get("/logout", async (req, res, next) => {
+  const session = await getSessionFromStorage(req.session!.sessionId);
+  if (session) {
+    session.logout();
+    res.send(`<p>Logged out</p>`);
```
pmcb55, line 116: Again, just more info on logout (I assume the same WebID could login multiple times and have multiple sessions, so just to explain that so the developer doesn't think logging out of one of their session logs them out of all their sessions):
```suggestion
    const sessionInfo = session.info;
    session.logout();
    res.send(`<p>Logged out WebID [<strong>${sessionInfo.webId}</strong>] (which had session ID [<strong>${sessionInfo.sessionId}</strong>] (the same WebID could be logged into multiple sessions!))</p>`);
```

packages/node/example/multiSession/src/serverSideApp.ts
```diff
@@ -0,0 +1,128 @@
...
+  } else {
+    res.status(400).send(`<p>No active session to log out</p>`);
+  }
+});
+
+app.listen(PORT, async () => {
+  console.log(`Listening on ${PORT}`);
```
pmcb55, line 123:
```suggestion
  console.log(`Listening on [${PORT}]...`);
```

pmcb55 (approved):
All these comments are just trivial tweaks - really great job!

packages/node/example/multiSession/src/serverSideApp.ts
pmcb55: Yeah, certainly for demo purposes, wrapping in HTML makes the most sense. But I think maybe Vinnie is referring to this script being **_actually useful_** - i.e. where you want the script to fetch a private resource for you, and return that resource as pure RDF (or binary if it's a Non-RDFSource). That's a really good point (if that's what Vinnie meant, otherwise it might just be a silly point that I'm making :) !) - so I'd support that ability with a separate `/fetchRaw` endpoint that does just stream back whatever the server replies with. (But not for now - in a subsequent PR)

packages/node/example/multiSession/src/serverSideApp.ts
pmcb55: Ok fine, so then I'd just change the values to reflect that:
```suggestion
    // These keys are required by cookie-session to sign the cookies.
    keys: ["Required, but value not relevant for this demo - key1", "Required, but value not relevant for this demo - key2"],
```

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: 837acf456a0fb34a401b067aca4e7d5206e6c340

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: 837acf456a0fb34a401b067aca4e7d5206e6c340

packages/node/example/multiSession/src/serverSideApp.ts
NSeydoux: 837acf456a0fb34a401b067aca4e7d5206e6c340

packages/node/example/multiSession/package.json
NSeydoux: 9a6dec5215e71f56d4b543a238051cb677b0a823

PR #990, opened 2021-01-26
Also list default session functions in API docs

Vinnl:
The API docs did not list the functions using the default session, because they were not listed in the toctree. I think this should fix that, but I'll mark this PR as a draft until I can verify for sure that this fixes it in the preview deployment.

### DIFF
--- a/packages/browser/docs/api/source/index.rst
+++ b/packages/browser/docs/api/source/index.rst
@@ -10,3 +10,4 @@ solid-client-authn-browser API
 
    /classes/*
    /interfaces/*
+   /modules/*
--- a/packages/browser/tsconfig.json
+++ b/packages/browser/tsconfig.json
@@ -8,7 +8,7 @@
 
   "typedocOptions": {
     "out": "website/docs/api/browser",
-    "entryPoints": ["./src/index.ts"],
+    "entryPoints": ["./src/index.ts", "./src/defaultSession.ts"],
     "entryDocument": "index.md"
   },
 


Vinnl:
Hi @kay-kim, I was wondering if you can help here. After generating the API docs, I have the following files in /packages/browser/docs/api/source:

```bash
$ ls --recursive
.:
api  index.rst

./api:
classes  index.md  interfaces

./api/classes:
configurationerror.md  inmemorystorage.md  notimplementederror.md  sessionmanager.md  session.md

./api/interfaces:
ilogininputoptions.md  isessioninfo.md  isessionmanageroptions.md  isessionoptions.md  istorage.md
```

Now `/api/index.md` contains some API documentation that I want to be visible. However, if I add it to `/index.rst` as such:

```
:orphan:

==============================
solid-client-authn-browser API
==============================

.. toctree::
   :glob:
   :titlesonly:

   /classes/*
   /interfaces/*
   /index.md
```

[I get the following error](https://vercel.com/inrupt/solid-client-authn-js/7mutqleya):

> /vercel/workpath0/packages/browser/docs/api/build/source/index.rst:self referenced toctree found. Ignored.

(I'm assuming that's related to it.) Do you know what the correct way to have `index.md` included in the built Sphinx site is?

kay-kim:
Hey @Vinnl  -- so ... in Sphinx, we don't actually specify the file extension - we just include the file name -- 
so, when including index.md in index.rst, Sphinx thinks it's including itself (hence the self-reference error).
 
Also, the index.md is just the typedoc's README.md -- and typedoc outputs to source/api/index.md and not source/index.md. 
I think we can just output the README.md to some other file name, include that in the index.rst, and update the Makefile when we copy over the files.  I can add a follow up commit to this pull request since I just wrote a jumble of words :D

Vinnl:
@kay-kim In this case, it's not actually the README - it includes a bunch of functions that I want documented. Although now that you mention it, it _is_ weird that it's in `index.md`.

The reason I added `/index.md` rather than `/api/index.md` is because it also outputs thing to `/api/classes/*`, which is also just listed as `/classes/*`. But let me see what happens if I remove the file extension and update the path; if that works, maybe that helps in debugging.

kay-kim:
Hey @Vinnl  -- 
the directory thing -- that's because of the Makefile:

> migrate: clean
> 	mkdir -p $(SOURCECOPYDIR)
> 
>    # Copying to SOURCECOPYDIR instead of copying source dir to BUILDDIR
>    # in case someone forgets to backslash after build/
> 
>    # First remove redundant index.md (auto-generated) so that
> 	# we can use the index.rst file (which has the toctree)
> 	if [ -f $(SOURCEDIR)/api/index.md ]; then rm $(SOURCEDIR)/api/index.md; fi

NSeydoux (approved):

Vinnl:
@kay-kim @NSeydoux So the default session functions are now displayed: https://solid-client-authn-js-o6fezhdvg.vercel.app/browser/modules/defaultsession.html

However, somehow adding that also triggered the showing of the index module, which is redundant: https://solid-client-authn-js-o6fezhdvg.vercel.app/browser/modules/index.html

I tried to remove it but didn't manage without breaking other stuff. I'm inclined to give up and leave it this way, but let me know if you have other insights or thoughts.

kay-kim:
Let me take a look.

kay-kim:
So ... not sure why the index.ts isn't sufficient to act as  single entry point when it exports all the things, including defaultSession.ts 

The issue now when we include defaultSession.ts as an entry point is that all the other files are relative to the modules/index.md - which is why all class and interface files now are prefixed with "index." - which we definitely don't want.
See: https://solid-client-authn-js-o6fezhdvg.vercel.app/browser/classes/index.configurationerror.html

Will have to look a bit more into typedoc and see if we need to do something to defaultSession.ts

NSeydoux:
Note that there are functions exported from `Session.ts`, but not part of the `Session` class, which are missing from the current API doc but included in https://solid-client-authn-js-o6fezhdvg.vercel.app/browser/modules/index.html. We also want these functions to be part of the final doc, and since these also are "standalone" functions not attached to a class, whatever issue prevents the default session functions to be exposed should also apply for these.

kay-kim:
Our entry point `src/index.ts` does the following: `export * from "./defaultSession";` 

- Since these functions aren't part of a class or interface or whatever,  typedoc  seems to includes these functions in the `index.md` (our entryDocument) rather than treat those functions as belonging to a separate module.

Possible Solution:  If we modify src/index.ts  to `export * as defaultSession from "./defaultSession";` , then

- typedoc will create a `module/defaultsession.md` but have as its title Namespace: defaultSession 

- Luckily, we can easily modify the makefile in Sphinx to just change the title inside the file to Module: defaultSession so that it looks like: https://kay-stage-docs2-git-authn-defaultsession.kay-kim.vercel.app/modules/defaultsession.html#module-defaultsession

But, am not sure if specifying a namespace during export is sanctioned or verboten.

Vinnl:
@kay-kim Unfortunately that would affect actual usage of those functions: rather than being able to `import { fetch } from "@inrupt/solid-client-authn-browser";` you will then have to use `import { defaultSession } from "@inrupt/solid-client-authn-browser";` and use `defaultSession.fetch` instead, which both kind of defeats the purpose of these API's, and would break tree-shaking (i.e. importing `fetch` also results in sending the rest of the default session API over the write). I don't think we want to let the our docs build system influence the use of our library :)

NSeydoux:
@kay-kim , to be able to cut a release, I merged this in in its current state, but we can revisit it later to make it cleaner.

PR #997, opened 2021-01-28
List session management functions in the API doc

NSeydoux:
Functions not exposed by a class, such as `getSessionFromStorage`, were
not listed as part of the public API. This introduces the same issue
that was identified in https://github.com/inrupt/solid-client-authn-js/pull/990

### DIFF
--- a/packages/node/src/Session.ts
+++ b/packages/node/src/Session.ts
@@ -73,7 +73,7 @@ export interface ISessionOptions {
 /**
  * If no external storage is provided, this storage gets used.
  */
-const defaultStorage = new InMemoryStorage();
+export const defaultStorage = new InMemoryStorage();
 
 /**
  * A {@link Session} object represents a user's session on an application. The session holds state, as it stores information enabling acces to private resources after login for instance.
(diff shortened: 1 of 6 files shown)


packages/node/src/Session.ts
```diff
@@ -73,7 +73,7 @@ export interface ISessionOptions {
 /**
  * If no external storage is provided, this storage gets used.
  */
-const defaultStorage = new InMemoryStorage();
+export const defaultStorage = new InMemoryStorage();
```
Vinnl, line 76: Maybe it's time we add an exports map to SCAN as well, so that at least in Node 12 + (i.e. soon all supported releases) it's not actually possible for developers to reach into the package and grab this.

Vinnl (approved):

packages/node/src/multiSession.ts
```diff
@@ -0,0 +1,84 @@
...
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { IStorage } from "@inrupt/solid-client-authn-core";
+import ClientAuthentication from "./ClientAuthentication";
+import { getClientAuthenticationWithDependencies } from "./dependencies";
+import { defaultStorage, Session } from "./Session";
+
+export async function getSessionFromStorage(
```
Vinnl, line 27: Btw, it's nice that these are now listed in the docs, but maybe we should add some actual docs to these as well? :)

NSeydoux: :scream:

PR #999, opened 2021-01-28
Swallow exception when looking up custom endpoint

### LINKED ISSUE #996: Login fails on CORS error with /session endpoint
After updating the library this morning, we were no longer able to login to TrinPod because of a CORS error on the new /session endpoint of the login sequence.  We fixed the issue by adding a /session enpoint to TrinPod that now returns a CORS response.

```
Access to fetch at 'http://frederick.localhost:8333/session' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

NSeydoux:
Resolves #996 

This prevents a CORS error thrown by an internal fetch to bubble up an
discrupt the login flow. Note that this is part of a temporary code
snippet that will be removed once a longer-term solution for auth after
refresh is implemented.

### DIFF
--- a/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -146,11 +146,10 @@ function mockClientRegistrar(client: IClient): IClientRegistrar {
 }
 
 const mockFetch = (response: Response): void => {
-  window.fetch = jest.fn().mockReturnValue(
-    new Promise((resolve) => {
-      resolve(response);
-    })
-  ) as jest.Mock<ReturnType<typeof window.fetch>, [RequestInfo, RequestInit?]>;
+  window.fetch = jest.fn().mockResolvedValue(response) as jest.Mock<
+    ReturnType<typeof window.fetch>,
+    [RequestInfo, RequestInit?]
+  >;
 };
 
 const defaultMocks = {
--- a/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -424,6 +424,42 @@ describe("AuthCodeRedirectHandler", () => {
     ).toEqual({});
   });
 
+  it("swallows the exception if the fetch to the session endpoint fails", async () => {
+    window.fetch = jest
+      .fn()
+      .mockReturnValueOnce(
+        new Promise((resolve) => {
+          resolve(
+            new Response("https://my.webid", {
+              status: 200,
+            })
+          );
+        })
+      )
+      .mockImplementation(() => {
+        throw new Error("Some error");
+      });
+
+    const mockedStorage = mockStorageUtility({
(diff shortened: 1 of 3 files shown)


packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -424,6 +424,42 @@ describe("AuthCodeRedirectHandler", () => {
...
+          resolve(
+            new Response("https://my.webid", {
+              status: 200,
+            })
+          );
+        })
+      )
+      .mockImplementation(() => {
+        throw new Error("Some error");
+      });
```
Vinnl, line 441: This can be simplified with `mockResolvedValue`. Additionally, it's a rejected Promise (since it's an async function), not an exception, so I suppose that would more accurately use `mockRejectedValue`?

```suggestion
      .mockResolvedValueOnce(
        new Response("https://my.webid", {
          status: 200,
        })
      )
      .mockRejectedValueOnce(new Error("Some error"));
```

Vinnl (approved):

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
NSeydoux: Right, that looks much better: c2b1f5bf971b5b781bc45d64550b5326336de6c4

NSeydoux: Actually let's fix the root cause ^^ c0b777bf192af90006e1acb3a1397df28b7b2e5c

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -91,27 +91,38 @@ async function setupResourceServerSession(
...
-    await storageUtility.storeResourceServerSessionInfo(
-      webId,
-      resourceServerIri,
-      // Note that here, if the lifespan of the cookie was returned by the server,
-      // we'd expect a relative value (the remaining time of validity) rather than
-      // an absolute one (the moment when the cookie expires).
-      Date.now() + DEFAULT_LIFESPAN
+  try {
+    const resourceServerResponse = await authenticatedFetch(
+      `${resourceServerIri}/session`
```
pmcb55, line 96: Jeez - seeing these hard-coded magic string values buried deep inside code like this makes the Java-developer in me squirm like crazy. I really think we should get these out - what's the issue with just moving these to a 'constant.ts' file or something (ideally into a vocab, but a local 'constant.ts' file seems like an easy first baby step)...?

NSeydoux: This is intentionally self-contained so that it is easy to remove, because it is meant to be a temporary solution.

pmcb55: Ah Ok - well, then a big dirty `TODO:` comment to that effect is fine instead I suppose...

pmcb55 (approved):

PR #1007, opened 2021-02-01
Feat/store idtoken on login

pmcb55:
#### browser

This branch was originally created to save the OAuth2 ID Token in session storage, but in fact we were already doing that, so just took the opportunity to clean up some of the surrounding code instead, by removing some of unsupported/unused code.

- removed Implicit Flow code.
- removed unused TokenSaver code.


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Internal refactor
+
+#### browser
+
+- removed Implicit Flow code.
+- removed unused TokenSaver code.
+
```
NSeydoux, line 17: An internal refactoring shouldn't me mentioned in the CHANGELOG, which is meant for the consumers of the library: the developers should rather read the commit history. In this case though, it is indeed a change that should be reflected to library users, as feature change: the implicit flow is no longer supported. Whether this is really breaking or not is up for debate, because while it technically is a breaking change, I don't think any OIDC issuer only supports the implicit flow and not the auth code flow, and no user-facing controls enable chosing one's flow, so there is no reason why this should be breaking, and it could merely be a patch release.

NSeydoux (approved):

PR #1018, opened 2021-02-02
Store issuer claim in localStorage

pmcb55:
# New feature description
- store the user's issuer claim, specifically to 'localStorage' to allow retrieval on tab refresh.

# Checklist


packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -38,6 +38,35 @@ import {
 } from "../../../../src/login/oidc/redirectHandler/AuthCodeRedirectHandler";
 import { RedirectorMock } from "../../../../src/login/oidc/__mocks__/Redirector";
 import { SessionInfoManagerMock } from "../../../../src/sessionInfo/__mocks__/SessionInfoManager";
+import { KEY_CURRENT_ISSUER } from "../../../../dist/constant";
+
+// Not sure this is strictly necessary, but perhaps best to mock out
+// 'localStorage' in case of parallelized tests.
```
NSeydoux, line 44: I think it's safe to remove this comment, mocking out localStorage does make sense.

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -281,37 +310,56 @@ describe("AuthCodeRedirectHandler", () => {
     // We use ts-ignore comments here only to access mock call arguments
     /* eslint-disable @typescript-eslint/ban-ts-comment */
     it("returns an authenticated bearer fetch by default", async () => {
+      /* tslint:disable-next-line:no-any */
+      (global as any).localStorage = new LocalStorageMock();
```
NSeydoux, line 314: Does it work if this is replaced by `window.localStorage = ...` ?

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -281,37 +310,56 @@ describe("AuthCodeRedirectHandler", () => {
...
       expect(
         // @ts-ignore
         header
       ).toMatch(/^Bearer some token$/);
+
+      // Also check that our issuer was stored correctly (specifically in
+      // 'localStorage').
+      expect(window.localStorage.getItem(KEY_CURRENT_ISSUER)).toEqual(
+        testIssuer
+      );
```
NSeydoux, line 356: Shouldn't the `handle` method always store the issuer, regardless of the type of token ? In this case, I think an independent test would be best.

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
pmcb55: Sure - 9d54085

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
pmcb55: Yeah, that's better (although still need `(window as any)` to avoid TS error with overwriting a readonly property!) - 39ae8d7

packages/browser/__tests__/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
pmcb55: Yeah, I did consider it - and I'm kinda 50-50 on it. It was just 1-point ticket, and so I lumped it into existing tests. My justification was just that there's no logic involved now (at least not now that the 'bearer token' tests have been updated to include `issuer` extraction, which is actually an unrelated improvement!), so the fact we're testing for this behaviour on existing happy-path tests is kinda Ok I think.

NSeydoux (approved):

PR #1022, opened 2021-02-03
Release

NSeydoux:
This PR bumps the version to 1.5.1.

Note that even though this technically adds a "new feature" (saving the ID token), I listed this as a patch release, because the feature in question doesn't change anything on the user experience yet, and no additional API functionnality is added. I also somewhat shrinked the CHANGELOG note about the implicit flow support to make it more compact and list it under the deprecation section.

# Checklist

- [X] I used `lerna version <major|minor|patch> --no-push` to update the version number, first inspecting the CHANGELOG to determine if the release was major, minor or patch.
- [X] The CHANGELOG has been updated to show version and release date - https://keepachangelog.com/en/1.0.0/.
- [X] The **only** commits in this PR are:
  - the CHANGELOG update.
  - the version update.
- [ ] I will make sure **not** to squash these commits, but **rebase** instead.
- [ ] Once this PR is merged, I will push the tag created by `lerna version ...` (e.g. `git push origin vX.X.X`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,31 +8,18 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+The following sections document changes that have been released already:
+
+## 1.5.1 - 2020-02-03
+
 ### Deprecation
 
 #### browser
 
 - Deprecated SessionManager
-
-### Internal refactor
-
-These changes are being reflected here even though normally internal refactoring 
-shouldn't be mentioned in the CHANGELOG (since it's meant for the consumers of
-the library, whereas library developers would rather read the commit history).
-In this case though, it is indeed a change that should be mentioned to
-library users, as it's really a feature change: i.e., the implicit flow is no
-longer supported.
-
-Whether this is really breaking or not is up for debate, because while it
-technically is a breaking change, we don't think any OIDC issuers only support
-the implicit flow and not the auth code flow, and no user-facing controls enable
-chosing one's flow, so there's no reason why this should be breaking, and it
-could be merely a patch release.
-
-#### browser
-
-- removed Implicit Flow code.
-- removed unused TokenSaver code.
+- The implicit flow is no longer supported. However, no known Solid Identity issuer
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -8,45 +8,30 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
-These changes are being reflected here even though normally internal refactoring 
-shouldn't be mentioned in the CHANGELOG (since it's meant for the consumers of
-the library, whereas library developers would rather read the commit history).
-In this case though, it is indeed a change that should be mentioned to
-library users, as it's really a feature change: i.e., the implicit flow is no
-longer supported.
-
-Whether this is really breaking or not is up for debate, because while it
-technically is a breaking change, we don't think any OIDC issuers only support
+- The implicit flow is no longer supported. However, no OIDC issuer only supports
```
Vinnl, line 20: I'm not sure, but is "OIDC issuer" too broad?

```suggestion
- The implicit flow is no longer supported. However, no known Solid server only supports
```

CHANGELOG.md
```diff
@@ -8,45 +8,30 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 the implicit flow and not the auth code flow, and no user-facing controls enable
-chosing one's flow, so there's no reason why this should be breaking, and it
-could be merely a patch release.
-
-#### browser
-
-- removed Implicit Flow code.
-- removed unused TokenSaver code.
+chosing one's flow, so there's no reason why this should be breaking.
```
Vinnl, line 22: "chosing" is a typo, and I think we can be a little more confident here.

```suggestion
choosing one's flow, so this has no user impact.
```

Vinnl (approved):
If there's nothing a developer or user can do with the new version that they weren't able to do before, it's not a new feature, so looks good to me :)

CHANGELOG.md
NSeydoux: That's a good point, it's more precise if we focus it to "Solid Identity Issuers" (Solid servers being a little unclear, since it may apply to both IdPs and Pod servers)

PR #1047, opened 2021-02-11
Log the user out when the session expires

Vinnl:
When the session expired, `session.info.isLoggedIn` would still be `true`, and there was no way for the developer to know that the session was no longer valid.

With this change, we check the expiration time as sent by the server, add it to a timestamp we generated _before_ sending a request to the server (to ensure we're not running behind), and then log the session out when it expires (this also generates a `logout` event).

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -260,6 +260,38 @@ describe("Session", () => {
       );
     });
 
+    it("logs the session out when its tokens expire", async () => {
+      jest.useFakeTimers();
+      const MOCK_TIMESTAMP = 10000;
+      jest.spyOn(Date, "now").mockReturnValueOnce(MOCK_TIMESTAMP);
+      const clientAuthentication = mockClientAuthentication();
+      const incomingRedirectHandler = jest.spyOn(
+        clientAuthentication,
+        "handleIncomingRedirect"
+      );
+      incomingRedirectHandler.mockResolvedValueOnce({
+        isLoggedIn: true,
+        sessionId: "Arbitrary session ID",
+        expirationDate: MOCK_TIMESTAMP + 1337,
+      });
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.url");
+      expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), 1337);
+      expect(mySession.info.isLoggedIn).toBe(true);
+
+      // Usually, we'd be able to use `jest.runAllTimers()` here. However,
+      // logout happens asynchronously. While this is inconsequential in
+      // running apps (the timeout complets asynchronously anyway), here, we can
+      // take advantage that we return the Promise from the callback in
+      // `setTimeout`, so that we can `await` it in this test before checking
+      // whether logout was successful:
+      const expireTimeout = ((setTimeout as unknown) as jest.Mock<
+        typeof setTimeout
+      >).mock.calls[0][0];
+      await expireTimeout();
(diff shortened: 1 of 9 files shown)


packages/browser/__tests__/Session.spec.ts
```diff
@@ -271,37 +297,33 @@ describe("Session", () => {
       expect(mySession.info.sessionId).toEqual("mySession");
     });
 
-    function sleep(ms: number): Promise<void> {
-      return new Promise((resolve) => setTimeout(resolve, ms));
```
Vinnl, line 275: Since I had to mock `setTimeout` for the other test I couldn't use this improvised `sleep` function anymore, but I think I've rewritten it to the same effect. Specifically, rather than just retrying every `ms` and checking whether `secondRequestIssued` was `true` already, I just manually resolve the Promise from the first request _after_ I issued the second one.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -271,37 +297,33 @@ describe("Session", () => {
...
       const firstTokenRequest = mySession.handleIncomingRedirect(
         "https://my.app/?code=someCode&state=arizona"
       );
       const secondTokenRequest = mySession.handleIncomingRedirect(
         "https://my.app/?code=someCode&state=arizona"
       );
-      secondRequestIssued = true;
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      // We know that it has been set by the call to `blockingRequest`.
+      continueAfterSecondRequest!();
```
Vinnl, line 326: I'm not sure why there are still Check warnings here, but I think they're from an earlier CI run, noting the ESLint comment on line 324.

NSeydoux: I'm not sure GH Actions even consider ESLint comments, which is kind of annoying, in particular since the repeat FOR EACH OS OF THE RUN MATRIX

Vinnl: I mean, would you get the point if it was listed less than six times?

(But yeah, as far as I know the annotations arise from warnings thrown during the build commands that we specify, i.e. the specific ESLint command, so I think it's _supposed_ to consider them.)

packages/browser/__tests__/Session.spec.ts
```diff
@@ -260,6 +260,32 @@ describe("Session", () => {
...
+      });
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.url");
+      expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), 1337);
+      expect(mySession.info.isLoggedIn).toBe(true);
+
+      const expireTimeout = ((setTimeout as unknown) as jest.Mock<
+        typeof setTimeout
+      >).mock.calls[0][0];
+      await expireTimeout();
```
NSeydoux, line 285: I'm discovering this part of the jest API, so this is an educational question for me: why couldn't `jest.runAllTimers();` be used there ?

NSeydoux (approved):

packages/browser/__tests__/Session.spec.ts
Vinnl: It was new for me too :) But the reason I didn't is because the function we pass to `setTimeout` returns a Promise which, in regular code, will be ignored, but is useful in the test to tell us when logging out actually succeeded (because that's async as well). So now the `expect` on the next line will actually succeed, rather than being called before `logout` is done.

But this is a reasonable question; I'll add a comment to mention this as well.

Vinnl: 39dec622

PR #1064, opened 2021-02-16
record suggested refactor of oidc-client-ext

pmcb55:
Just added a suggested refactoring improvement to the mock for 'oidc-client-ext' as a comment (just so it doesn't get lost).


NSeydoux (approved):
I think I'd add a TODO/FIXME label somewhere, so that we can find this when searching for these keywords across the codebase.

pmcb55:
D'oh - yeah, good point - done: #f6fddab

PR #1065, opened 2021-02-16
Verify ID token when looking up current issuer 

NSeydoux:
This adds the method `getCurrentIssuer` to the ClientAuthentication class.
Based on the current session ID in storage, the associated ID token is
validated, before returning its issuer to the caller. This way, we ensure
that the ID token hasn't been tampered with, that its audience is the
client associated to the current session, and that the issuer hasn't
been changed in storage either.

# Checklist

### DIFF
--- a/packages/browser/__tests__/ClientAuthentication.spec.ts
+++ b/packages/browser/__tests__/ClientAuthentication.spec.ts
@@ -22,10 +22,15 @@
 // Required by TSyringe:
 import "reflect-metadata";
 import {
+  IIssuerConfig,
   mockStorageUtility,
   StorageUtility,
+  USER_SESSION_PREFIX,
 } from "@inrupt/solid-client-authn-core";
 import { mockStorage } from "@inrupt/solid-client-authn-core/dist/storage/__mocks__/StorageUtility";
+import { Response } from "cross-fetch";
+import { JSONWebKey } from "jose";
+import { signJwt } from "@inrupt/oidc-client-ext";
 import { LoginHandlerMock } from "../src/login/__mocks__/LoginHandler";
 import {
   RedirectHandlerMock,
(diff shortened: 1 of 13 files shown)


packages/browser/src/ClientAuthentication.ts
```diff
@@ -123,6 +126,47 @@ export default class ClientAuthentication {
...
 
+  getCurrentIssuer = async (): Promise<string | null> => {
+    const currentSessionId = window.localStorage.getItem(KEY_CURRENT_SESSION);
+    if (currentSessionId === null) {
+      return null;
+    }
+    const sessionInfo = await this.sessionInfoManager.get(currentSessionId);
+    // Several session information are required in order to validate that the ID
+    // token in storage hasn't been tampered with, and has actually been issued
+    // by the issuer present in storage.
```
Vinnl, line 137: "Information" is uncountable in English :)

```suggestion
    // Several types of session data are required in order to validate that the ID
    // token in storage hasn't been tampered with, and has actually been issued
    // by the issuer present in storage.
```

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -22,10 +22,15 @@
...
 import {
+  IIssuerConfig,
   mockStorageUtility,
   StorageUtility,
+  USER_SESSION_PREFIX,
 } from "@inrupt/solid-client-authn-core";
 import { mockStorage } from "@inrupt/solid-client-authn-core/dist/storage/__mocks__/StorageUtility";
+import { Response } from "cross-fetch";
+import { JSONWebKey } from "jose";
+import { signJwt } from "@inrupt/oidc-client-ext";
```
Vinnl, line 33: Just wondering... Shouldn't this be a relative import?

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -34,14 +39,92 @@ import {
...
+    iat: 1462266216,
+  };
+};
+
+// mockIdTokenPayload(
+//   "https://my.pod/profile#me",
+//   "https://some.issuer",
+//   "https://some.app/registration"
+// ),
+
```
Vinnl, line 97:
```suggestion
```

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -318,4 +402,243 @@ describe("ClientAuthentication", () => {
...
+  describe("getCurrentIssuer", () => {
+    it("returns null no current session is in storage", async () => {
+      const clientAuthn = getClientAuthentication({});
+
+      await expect(clientAuthn.getCurrentIssuer()).resolves.toBeNull();
+    });
+
+    // ts-ignore is used to mock out local storage.
+    /* eslint-disable @typescript-eslint/ban-ts-comment */
```
Vinnl, line 414: This appears to be a couple of lines too high? (Which I would assume would result in it not working?)

Also, I'd generally prefer type assertions over ts-ignore, to make sure we only catch expected errors. (So in these cases, something like `(new LocalStorageMock(/* ... */)) as typeof window.localStorage` or something.)

Vinnl: Ah wait, I suppose `window.localStorage` is readonly? In that case, what you could do is add a type assertion to `window` I think, e.g. `(window as any).localStorage`.

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -318,4 +402,243 @@ describe("ClientAuthentication", () => {
...
+
+      const clientAuthn = getClientAuthentication({
+        issuerConfigFetcher: mockedIssuerConfig,
+        sessionInfoManager: mockSessionInfoManager(mockedStorage),
+      });
+
+      await expect(clientAuthn.getCurrentIssuer()).resolves.toBeNull();
+    });
+
+    it("returns null if the current client ID doesn't match the ID token audience", async () => {
```
Vinnl, line 561: It's not entirely clear to me which two values are the ones that shouldn't match for this test - is it `https://some-other.app/registratoin` and `https://some.issuer/jwks`? Maybe some clarifying comment could help for future readers.

packages/browser/__tests__/ClientAuthentication.spec.ts
```diff
@@ -318,4 +402,243 @@ describe("ClientAuthentication", () => {
...
+
+      const clientAuthn = getClientAuthentication({
+        issuerConfigFetcher: mockedIssuerConfig,
+        sessionInfoManager: mockSessionInfoManager(mockedStorage),
+      });
+
+      await expect(clientAuthn.getCurrentIssuer()).resolves.toBeNull();
+    });
+
+    it("returns null if the ID token signature cannot be verified", async () => {
```
Vinnl, line 588: And here I'm having real trouble figuring out what the reason is that the ID token signature cannot be verified - though that might just be my ignorance of OIDC.

packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
```diff
@@ -20,7 +20,11 @@
  */
 
 import "reflect-metadata";
-import { mockStorageUtility } from "@inrupt/solid-client-authn-core";
+import {
+  mockStorageUtility,
+  StorageUtility,
+} from "@inrupt/solid-client-authn-core";
+import { mockStorage } from "@inrupt/solid-client-authn-core/dist/storage/__mocks__/StorageUtility";
```
Vinnl, line 27: If these shouldn't be relative imports, then it's still a bit weird that one imports from the package proper, and the other reaches into `/dist`.

Vinnl (approved):
I don't really know the context of this PR, but it seems to make sense and code-wise looks mostly OK.

packages/browser/src/ClientAuthentication.ts
NSeydoux: Ah right, I always forget ^^

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: I'm not 100% sure, but I think Lerna deals with that under the cover, that's how I understand `lerna bootstrap` at least. Each code base acts as if it was independent, and it's the monorepo tooling that stitches it together.

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: Yeah exactly it's because `localStorage` is readonly. I'll change it to your suggestion, which is also how Pat approached it in another file.

packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
NSeydoux: Oops yeah that looks like a good ol' auto import

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: 9489169716956c6376af4bf92fdd2cb3abafc4df

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: I think part of the issue is that some values come from the defaults st at the top of the file, which isn't really DAMP, and really isn't clear when reading the test. I think 26a099b4adc9a94574af9c7c6ad816c29cbb0ff5 will improve that.

packages/browser/__tests__/ClientAuthentication.spec.ts
NSeydoux: I think the test description wasn't great either. 26a099b4adc9a94574af9c7c6ad816c29cbb0ff5 should make it better

packages/browser/__tests__/sessionInfo/SessionInfoManager.spec.ts
NSeydoux: Actually no, the export was missing: 0c89e2c953720663712c5ab8f014a6c72a417ce7

packages/browser/__tests__/ClientAuthentication.spec.ts
Vinnl: Yep clearer, thanks!

PR #1069, opened 2021-02-17
Feat/restore url on redirect

pmcb55:
# New feature description
Added new `onSessionRestore` event to `Session` (and default session) to allow
  the developer to register an event callback that will be called whenever a
  session is restored (e.g., due to a browser page refresh). The callback is
  given a URL parameter, which represents the current URL of the browser
  *before* the session restoration (to allow the developer to restore their
  app's state if needed, e.g., if the app is a Single Page App (SPA) and the
  developer wishes to restore the users 'current page' to exactly where they
  were before the refresh).

# Checklist

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -331,8 +331,8 @@ describe("Session", () => {
       const secondTokenRequest = mySession.handleIncomingRedirect(
         "https://my.app/?code=someCode&state=arizona"
       );
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       // We know that it has been set by the call to `blockingRequest`.
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       continueAfterSecondRequest!();
       const tokenRequests = await Promise.all([
         firstTokenRequest,
@@ -553,7 +553,7 @@ describe("Session", () => {
 
   describe("onLogout", () => {
     // The `done` callback is used in order to make sure the callback passed to
-    // onLogin is called. If it is not, the test times out, which is why
+    // our event handler is called. If it is not, the test times out, which is why
     // no additional assertion is required.
     // eslint-disable-next-line jest/expect-expect, jest/no-done-callback
     it("calls the registered callback on logout", async (done) => {
@@ -565,8 +565,65 @@ describe("Session", () => {
       const mySession = new Session({
         clientAuthentication: mockClientAuthentication(),
       });
+
       mySession.onLogout(myCallback);
       await mySession.logout();
+    });
+  });
+
+  describe("onSessionRestore", () => {
+    // The `done` callback is used in order to make sure the callback passed to
+    // our event handler is called. If it is not, the test times out, which is why
+    // no additional assertion is required.
+    // eslint-disable-next-line jest/expect-expect, jest/no-done-callback
+    it("calls the registered callback on", async (done) => {
+      // Set our window's location to our test value.
+      const defaultLocation = "https://coolSite.com/myAppTestPage";
+
+      const myCallback = (urlBeforeRestore: string): void => {
+        expect(urlBeforeRestore).toEqual(defaultLocation);
+        if (done) {
+          done();
+        }
+      };
+
+      const {
+        location,
+        history: { replaceState },
+      } = window;
+
+      try {
+        // location and history aren't optional on window, which makes TS complain
+        // (rightfully) when we delete them. However, they are deleted on purpose
+        // here just for testing.
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        delete window.location;
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        delete window.history.replaceState;
+
+        window.location = {
+          href: defaultLocation,
+        } as Location;
+        window.history.replaceState = jest.fn();
+
+        const mySession = new Session({
+          clientAuthentication: mockClientAuthentication(),
+        });
+        mySession.onSessionRestore(myCallback);
+
+        // Callback should not be called on initial login...
+        await mySession.handleIncomingRedirect("https://not-relevant.com/");
+
+        // ...but should be called on a subsequent calls, i.e., on session
+        // restore after a silent login.
+        await mySession.handleIncomingRedirect("https://not-relevant.com/");
+      } finally {
+        // Restore our 'window' state.
+        window.location = location;
+        window.history.replaceState = replaceState;
+      }
     });
   });
 });
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -611,14 +611,21 @@ describe("Session", () => {
         const mySession = new Session({
           clientAuthentication: mockClientAuthentication(),
         });
+        const sessionRestoreTrigger = jest.spyOn(
+          mySession,
+          "fireSessionRestoreEvent"
+        );
+
         mySession.onSessionRestore(myCallback);
 
         // Callback should not be called on initial login...
         await mySession.handleIncomingRedirect("https://not-relevant.com/");
+        expect(sessionRestoreTrigger).not.toHaveBeenCalled();
 
         // ...but should be called on a subsequent calls, i.e., on session
         // restore after a silent login.
         await mySession.handleIncomingRedirect("https://not-relevant.com/");
+        expect(sessionRestoreTrigger).toHaveBeenCalled();
       } finally {
         // Restore our 'window' state.
         window.location = location;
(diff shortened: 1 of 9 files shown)


packages/browser/src/Session.ts
```diff
@@ -157,8 +163,10 @@ export class Session extends EventEmitter {
     url: string = window.location.href
   ): Promise<ISessionInfo | undefined> => {
     if (this.info.isLoggedIn) {
+      this.fireSessionRestoreEvent();
       return this.info;
     }
```
NSeydoux, line 168:
```suggestion
    if (this.info.isLoggedIn) {
      return this.info;
    }
```
In the browser context, if already logged in, it means no redirection happened the last login, which means there is no reason to restore the session here. This `if` test is here to prevent side-effects when calling `handleIncomingRedirect` not immediately after a redirect, with a logged in session, so I don't think we should call a callback in that context.

packages/browser/src/Session.ts
```diff
@@ -213,6 +221,8 @@ export class Session extends EventEmitter {
         ) {
           this.info.isLoggedIn = true;
           this.info.webId = reference.webId;
+
+          this.fireSessionRestoreEvent();
```
NSeydoux, line 225:
```suggestion
```
With this temporary workaround, we base authentication on the RS cookie, so no redirection happens either, which means there is no need to restore the previous URL.

NSeydoux (changes requested):
The `sessionRestore` event aims at solving the scenario where a browser app has client-side navigation, which means the registered redirect IRI may differ from the IRI of the page where the user currently is: this means that on refresh, when we do silent auth, the user would end up on the redirect page, instead of the page they just refreshed. To resolve that, before redirecting the user away, we'll store the current IRI, and restore it on redirect. This means the `restoreSession` event should be fired after an actual redirect-based login, so somewhere near 
```
this.tokenRequestInProgress = true;
    const sessionInfo = await this.clientAuthentication.handleIncomingRedirect(
      url
    );
    if (sessionInfo) {
      this.info.isLoggedIn = sessionInfo.isLoggedIn;
      this.info.webId = sessionInfo.webId;
      this.info.sessionId = sessionInfo.sessionId;
      if (sessionInfo.isLoggedIn) {
        // The login event can only be triggered **after** the user has been
        // redirected from the IdP with access and ID tokens.
        this.emit("login");
      }
```
My intuition is that it should happen in a different branch from the `login` event: either you login, or you restore a previous session, but not both. To choose one path or the other, you check if the current URL entry exists in storage or not.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -589,8 +589,72 @@ describe("Session", () => {
...
       await mySession.logout();
     });
   });
+
+  describe("onSessionRestore", () => {
+    // The `done` callback is used in order to make sure the callback passed to
+    // our event handler is called. If it is not, the test times out, which is why
+    // no additional assertion is required.
+    // eslint-disable-next-line jest/expect-expect, jest/no-done-callback
+    it("calls the registered callback on", async (done) => {
```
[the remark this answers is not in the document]
pmcb55: Yeah, I totally agree - I knew using `done` isn't the done thing anymore(!), but was just copying what was there. Yeah, the Jest way is much better.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -589,8 +589,72 @@ describe("Session", () => {
...
+        history: { replaceState },
+      } = window;
+
+      try {
+        // location and history aren't optional on window, which makes TS complain
+        // (rightfully) when we delete them. However, they are deleted on purpose
+        // here just for testing.
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        delete window.location;
```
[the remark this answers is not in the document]
pmcb55: Done - 96bb352

NSeydoux (approved):

packages/browser/__tests__/Session.spec.ts
```diff
@@ -565,8 +567,47 @@ describe("Session", () => {
...
     });
   });
+
+  describe("onSessionRestore", () => {
+    it("calls the registered callback on session restore", async () => {
+      // Set our window's location to our test value.
+      const defaultLocation = "https://coolSite.com/resource";
+      const currentLocation = "https://coolSite.com/redirect";
+
+      // This pretends we have previously triggerd silent authentication and stored
```
pmcb55, line 582:
```suggestion
      // This pretends we have previously triggered silent authentication and stored
```

packages/browser/__tests__/Session.spec.ts
```diff
@@ -565,8 +567,47 @@ describe("Session", () => {
...
+      // the location.
+      mockLocalStorage({
+        [KEY_CURRENT_URL]: defaultLocation,
+      });
+      // This acts as the URL the user has been redirected to.
+      mockLocation(currentLocation);
+      // This pretends the login is successful
```
pmcb55, line 589:
```suggestion
      // This pretends the login is successful.
```

packages/browser/src/Session.ts
```diff
@@ -57,22 +57,17 @@ async function silentlyAuthenticate(
...
-  if (
-    storedSessionInfo === undefined ||
-    storedSessionInfo.idToken === undefined
-  ) {
-    return;
-  }
-  window.localStorage.setItem(KEY_CURRENT_URL, window.location.href);
+  // currently logged in, and the user has refreshed their browser page. The ID
+  // token is validated, and its issuer claim is extracted to know where silent
+  // authentication is possible.
```
pmcb55, line 62: It's not immediately clear here that the ID token is actually being validated - so should we rename `getCurrentIssuer()` to be `validateCurrentIssuer()` or something, just to try and make it clearer what's going on?
Slight typo here:
```suggestion
  // currently logged in, and the user has refreshed their browser page. The ID
  // token is validated, and its issuer claim is extracted to know if silent
  // authentication is possible.
```

packages/browser/src/Session.ts
```diff
@@ -57,22 +57,17 @@ async function silentlyAuthenticate(
...
   const issuer = await clientAuthn.getCurrentIssuer();
-  if (issuer !== null) {
+  const storedSessionInfo = await clientAuthn.getSessionInfo(sessionId);
+  if (issuer !== null && storedSessionInfo !== undefined) {
```
pmcb55, line 65: What would it mean here if one of these conditions was true, but the other wasn't? Is that worth trying to explain in a comment here...? I'm also thinking in terms of error handling and debugging this in the field - i.e., if someone reports that silent authentication isn't working - could we easily tell why...?

packages/browser/src/Session.ts
```diff
@@ -204,9 +199,17 @@ export class Session extends EventEmitter {
...
-      // redirected from the IdP with access and ID tokens.
-      this.emit("login");
+      const currentUrl = window.localStorage.getItem(KEY_CURRENT_URL);
+      if (currentUrl === null) {
+        // The login event can only be triggered **after** the user has been
+        // redirected from the IdP with access and ID tokens.
+        this.emit("login");
+      } else {
+        // If an URL is stored in local storage, we are being logged in after a
+        // silent authentication.
```
pmcb55, line 209:
```suggestion
        // If an URL is stored in local storage, we are being logged in after a
        // silent authentication, so remove our currently stored URL location
        // to clean up our state now that we are completing the re-login process.
```

packages/browser/src/Session.ts
```diff
@@ -249,4 +252,16 @@ export class Session extends EventEmitter {
...
     this.on("logout", callback);
   }
+
+  /**
+   * Register a callback function to be called when a session is restored.
+   *
+   * Note: the callback will be called with the saved value of the 'current URL'
+   * at the time the session was restored.
+   *
+   * @param callback The function called when a user completes logout.
```
pmcb55, line 262:
```suggestion
   * @param callback The function called when a user's already logged-in session is restored, e.g., after a silent authentication is completed after a page refresh.
```

pmcb55 (commented):
Looks good - just a couple of comments.

packages/browser/src/Session.ts
NSeydoux: This isn't a type (although I guess the phrasing isn't great then ^^): the issuer claim indicates against which IdP silent authentication can be performed, hence the `where`

NSeydoux: 37d12534b06a1eac66de53e6948c3e306bebea26 renames the function

packages/browser/src/Session.ts
NSeydoux: Actually, thinking to answer this question lead me to refactor this portion a bit: it was redundant, because both `validateCurrentIssuer` and `getSessionInfo` looked up the same session information in storage. 26ad6902e4ab31ed2e972d7f47ecc61bafa4ece4 makes it clearer.

PR #1076, opened 2021-02-19
Store current IRI before silent authentication

NSeydoux:
This refactors code introduced in
1b4357f6b02327cf03ecf6d478392fdafc59e5ea. In the original design, the
current IRI was stored after successful login, while actually it makes
more sense to only store it during the silent authentication process,
right before silently redirecting the user away from the app, in case
client-side navigation happened and the user refreshes a page that is no
longer the one he logged in to.

For context, this is the first step to silent authentication: after a
refresh, the access token is no longer available. To log back in, the
use must be redirected to the IdP, and storing the current IRI before
redirecting makes it easier to restore it after the user is logged back
in.

# Checklist

### DIFF
--- a/packages/browser/src/Session.ts
+++ b/packages/browser/src/Session.ts
@@ -54,14 +54,9 @@ export interface ISessionOptions {
 }
 
 async function silentlyAuthenticate(
-  sessionId: string | null,
+  sessionId: string,
   clientAuthn: ClientAuthentication
 ) {
-  if (sessionId === null) {
-    // If there is no current session defined, there is no ID token, so silent
-    // authentication cannot happen.
-    return;
-  }
   // Check if we have an ID Token in storage - if we do then we may be
   // currently logged in, and the user has refreshed their browser page. In
   // this case, it can be really useful to save the user's current browser
@@ -74,7 +69,6 @@ async function silentlyAuthenticate(
     window.localStorage.setItem(KEY_CURRENT_URL, window.location.href);
   }
 }
-
 /**
  * A {@link Session} object represents a user's session on an application. The session holds state, as it stores information enabling acces to private resources after login for instance.
  */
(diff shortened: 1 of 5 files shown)


packages/browser/src/Session.ts
```diff
@@ -238,6 +263,15 @@ export class Session extends EventEmitter {
           await this.logout();
         }, sessionInfo.expirationDate - Date.now());
       }
+    } else {
+      // Silent authentication happens after a refresh, which means there are no
+      // OAuth parames in the current location IRI. It can only succeed if a session
```
pmcb55, line 268:
```suggestion
      // OAuth params in the current location IRI. It can only succeed if a session
```

packages/browser/src/Session.ts
```diff
@@ -52,6 +53,28 @@ export interface ISessionOptions {
...
+  // location, so that we can restore that location after we complete the
+  // entire re-login-the-user-silently-flow (e.g., if the user was on a
+  // specific 'page' of a Single Page App, then presumably they'll expect to
+  // be brought back to exactly that same 'page' after the full refresh).
+  const storedSessionInfo = await clientAuthn.getSessionInfo(sessionId);
+  if (storedSessionInfo?.idToken !== undefined) {
+    window.localStorage.setItem(KEY_CURRENT_URL, window.location.href);
+  }
+}
+
```
pmcb55, line 77: I think I'd just tweak this to always only work **_only_** with a sessionID, and put the sessionID===null check outside this function.
```suggestion
async function silentlyAuthenticate(
  sessionId: string,
  clientAuthn: ClientAuthentication
) {
  // Check if we have an ID Token in storage - if we do then we may be
  // currently logged in, and the user has refreshed their browser page. In
  // this case, it can be really useful to save the user's current browser
  // location, so that we can restore that location after we complete the
  // entire re-login-the-user-silently-flow (e.g., if the user was on a
  // specific 'page' of a Single Page App, then presumably they'll expect to
  // be brought back to exactly that same 'page' after the full refresh).
  const storedSessionInfo = await clientAuthn.getSessionInfo(sessionId);
  if (storedSessionInfo?.idToken !== undefined) {
    window.localStorage.setItem(KEY_CURRENT_URL, window.location.href);
  }
}
```

packages/browser/src/Session.ts
```diff
@@ -238,6 +263,15 @@ export class Session extends EventEmitter {
...
       }
+    } else {
+      // Silent authentication happens after a refresh, which means there are no
+      // OAuth parames in the current location IRI. It can only succeed if a session
+      // was previously logged in, in which case its ID will be present with a known
+      // identifier in local storage.
+      await silentlyAuthenticate(
+        window.localStorage.getItem(KEY_CURRENT_SESSION),
+        this.clientAuthentication
+      );
```
pmcb55, line 274: As a follow-up to the suggestion above, this would be become (just a bit more readable and cleaner I think):
```suggestion
      // Check if we have a locally stored session ID...
      const storedSessionID = window.localStorage.getItem(KEY_CURRENT_SESSION);
      // ...if not, then there is no ID token, and so silent authentication cannot happen, but
      // if we do have a stored session ID, attempt to re-authenticate now silently.
      if (storedSessionID !== null) {
        await silentlyAuthenticate(storedSessionID, this.clientAuthentication);
      }
```

pmcb55 (approved):
Looks great - really love the localStorage mocking refactor!

PR #1077, opened 2021-02-19
Add redirectUrl to stored session info

NSeydoux:
In order to do silent authentication, one must provide a redirect URL
(like for regular login). In order to do so, on successful login, the
redirect URL is stored with the session information, so that it can be
retrieved after refresh and be used to trigger silent authentication.

Note to reviewers: even if this is related to the same feature as #1076 , these two PRs are independant from each other and can be reviewed in any order.

# New feature description

# Checklist

### DIFF
--- a/packages/core/src/sessionInfo/ISessionInfo.ts
+++ b/packages/core/src/sessionInfo/ISessionInfo.ts
@@ -76,7 +76,7 @@ export interface ISessionInternalInfo {
   issuer?: string;
 
   /**
-   * The redirect IRI registered when initially logging the session in.
+   * The redirect URL registered when initially logging the session in.
    */
   redirectUrl?: string;
 }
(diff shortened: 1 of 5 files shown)


packages/core/src/sessionInfo/ISessionInfo.ts
```diff
@@ -74,4 +74,9 @@ export interface ISessionInternalInfo {
    * The OIDC issuer that issued the tokens authenticating the session.
    */
   issuer?: string;
+
+  /**
+   * The redirect IRI registered when initially logging the session in.
```
pmcb55, line 79: Pesky URI/URL/IRI naming, but in this case I think it's just URL:
```suggestion
   * The redirect URL registered when initially logging the session in.
```

pmcb55 (approved):

PR #1078, opened 2021-02-19
Add internal support for silent authentication

NSeydoux:
This enables to specify the "prompt" option from our own internal API
all the way down to the OIDC library we are using. The "prompt" option,
when set to 'none', prevents the Identity Provider to display any screen
or to have any sort of user interaction, which is what we want for
silent authentication.

Note that the "prompt" option was not added to
the external API, because there is no reason for the calling library to
decide it needs to go through silent authentication: we make this call
entierly based on internal state. The code to implement this logic of
whether we want or not to go through silent authentication is not part
of this PR.

# Checklist

### DIFF
--- a/packages/browser/__tests__/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.spec.ts
@@ -29,27 +29,29 @@ import {
   IRedirectorOptions,
   StorageUtilityMock,
 } from "@inrupt/solid-client-authn-core";
-import { SigninRequest } from "@inrupt/oidc-client-ext";
 import AuthorizationCodeWithPkceOidcHandler from "../../../../src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler";
 import canHandleTests from "./OidcHandlerCanHandleTests";
 import { SessionInfoManagerMock } from "../../../../src/sessionInfo/__mocks__/SessionInfoManager";
 import { standardOidcOptions } from "../../../../src/login/oidc/__mocks__/IOidcOptions";
 import { RedirectorMock } from "../../../../src/login/oidc/__mocks__/Redirector";
 
+jest.mock("@inrupt/oidc-client-ext");
+
 const expectedSigninRedirectUrl = "https://test";
-jest.mock("@inrupt/oidc-client-ext", () => {
-  return {
-    OidcClient: jest.fn().mockImplementation(() => {
-      return {
-        createSigninRequest: (): Promise<SigninRequest> =>
-          Promise.resolve({
-            url: expectedSigninRedirectUrl,
-            state: "test state",
-          }),
-      };
+
+const mockOidcModule = (
+  url: string = expectedSigninRedirectUrl,
+  state = "test state"
+) => {
+  const oidcModule = jest.requireMock("@inrupt/oidc-client-ext");
+  oidcModule.OidcClient = jest.fn().mockReturnValue({
+    createSigninRequest: jest.fn().mockResolvedValue({
+      url,
+      state,
     }),
-  };
-});
+  });
+  return oidcModule;
+};
 
 describe("AuthorizationCodeWithPkceOidcHandler", () => {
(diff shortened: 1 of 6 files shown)


packages/browser/__tests__/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.spec.ts
```diff
@@ -129,7 +133,27 @@ describe("AuthorizationCodeWithPkceOidcHandler", () => {
       );
     });
 
+    it("supports silent authentication", async () => {
```
pmcb55, line 136: Seems this test is more specifically about setting an option and asserting that the low-level library is called with it
```suggestion
    it("passes our the 'prompt' option down to our OIDC client library implementation", async () => {
```

packages/browser/src/ClientAuthentication.ts
```diff
@@ -70,7 +70,12 @@ export default class ClientAuthentication {
   // Isn't Javascript fun?
   login = async (
     sessionId: string,
-    options: ILoginInputOptions
+    options: ILoginInputOptions & {
+      // TODO: the 'prompt' parameter shouldn't be exposed as part of the public API.
+      // This classe's login should take the internal IOidcOptions type as an
```
pmcb55, line 75:
```suggestion
      // This login function should take the internal IOidcOptions type as an
```

packages/core/src/login/ILoginOptions.ts
```diff
@@ -38,7 +38,10 @@ export default interface ILoginOptions extends ILoginInputOptions {
   // 'bundling' into this options interface), but it wouldn't be a significant
   // improvement really...
   sessionId: string;
-
+  /**
+   * Specify wether the Solid Identity Provider may or may interact with the user.
```
pmcb55, line 42:
```suggestion
   * Specify whether the Solid Identity Provider may, or may not, interact with the user (for example,
   * the normal login process **_requires_** human interaction for them to enter their credentials,
   * but if a user simply refreshes the current page in their browser, we'll want to log them in again
   * automatically, i.e., without prompting them to manually provide their credentials again).
```

packages/core/src/login/oidc/IOidcOptions.ts
```diff
@@ -33,28 +33,31 @@ import { IClient } from "./IClient";
...
   sessionId: string;
   refreshToken?: string;
-}
-
-/**
- * @hidden
- */
-export interface IAccessTokenOidcOptions extends ICoreOidcOptions {
+  /**
+   * Specify wether the Solid Identity Provider may or may interact with the user.
```
pmcb55, line 49: Ditto to the comment above - same typos here too...

packages/core/src/login/oidc/IOidcOptions.ts
```diff
@@ -33,28 +33,31 @@ import { IClient } from "./IClient";
...
+   */
+  prompt?: string;
+  /**
+   * True if a dpop compatible auth_token should be requested.
```
pmcb55, line 53:
```suggestion
   * True if a DPoP compatible auth_token should be requested.
```

packages/core/src/login/oidc/IOidcOptions.ts
```diff
@@ -33,28 +33,31 @@ import { IClient } from "./IClient";
...
+   */
   dpop: boolean;
+  /**
+   * The URL to which the user should be redirected after authorizing.
```
pmcb55, line 57:
```suggestion
   * The URL to which the user should be redirected after logging in and 
   * authorizing our app to access their Pod.
```

pmcb55 (approved):

packages/core/src/login/oidc/IOidcOptions.ts
NSeydoux: 2f7f4b921cbb20661cc1adccba736312f7e587da

packages/core/src/login/oidc/IOidcOptions.ts
NSeydoux: I rephrased this slightly in 2f7f4b921cbb20661cc1adccba736312f7e587da, just to be extra clear that logging in refers in this context to the user logging in the IdP

PR #1079, opened 2021-02-19
Trigger silent authentication

NSeydoux:
This implements silent authentication e.g. after a refresh. To do so, it was necessary to retrieve client secrets with session information from storage, in order to provide them when calling `login` to trigger the silent flow. This includes a couple of fixes required for silent authentication to work, including validating ID tokens signed with RSA algorithms, and making sure that all relevant session information (in particular, the ID token) was stored in the permanent storage.

This simply redirects the user away from the page, and straight back from the Solid Identity Provider. The incoming redirect is identical to an incoming redirect for a regular login, which is why no dedicated code was added to that end.

# Checklist

### DIFF
--- a/packages/browser/src/sessionInfo/SessionInfoManager.ts
+++ b/packages/browser/src/sessionInfo/SessionInfoManager.ts
@@ -132,13 +132,11 @@ export class SessionInfoManager implements ISessionInfoManager {
         secure: true,
       }
     );
-    if (isLoggedIn === undefined) {
-      return undefined;
-    }
 
     const webId = await this.storageUtility.getForUser(sessionId, "webId", {
       secure: true,
     });
+
     const clientId = await this.storageUtility.getForUser(
       sessionId,
       "clientId",
@@ -174,9 +172,20 @@ export class SessionInfoManager implements ISessionInfoManager {
         secure: true,
       }
     );
+
     const issuer = await this.storageUtility.getForUser(sessionId, "issuer", {
-      secure: true,
+      secure: false,
     });
+
+    if (
+      clientId === undefined &&
+      idToken === undefined &&
+      isLoggedIn === undefined &&
+      webId === undefined &&
+      refreshToken === undefined
+    ) {
+      return undefined;
+    }
 
     return {
       sessionId,
--- a/packages/browser/src/sessionInfo/SessionInfoManager.ts
+++ b/packages/browser/src/sessionInfo/SessionInfoManager.ts
@@ -146,6 +146,15 @@ export class SessionInfoManager implements ISessionInfoManager {
         secure: false,
       }
     );
+
+    const clientSecret = await this.storageUtility.getForUser(
+      sessionId,
+      "clientSecret",
+      {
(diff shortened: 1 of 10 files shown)


packages/oidc/src/dpop/dpop.spec.ts
```diff
@@ -218,4 +218,25 @@ describe("validateIdToken", () => {
...
+  it("returns true if the ID token is valid for an RSA signature", async () => {
+    const payload = {
+      sub: "https://my.webid",
+      iss: "https://some.issuer",
+      aud: "https://my.app/registration",
+      iat: 2551294137,
+      exp: 2551301337,
+    };
+    const idToken = await signJwt(payload, mockJwk(), {
+      algorithm: "RS256",
```
acoburn, line 231: Are you planning to also support ES256 at some point?

NSeydoux: Currently, ES256 is the algorithm we use by default for DPoP tokens, and both ES256 and RS256 are supported for ID token validation. I was under the understanding that verifying EC signatures was faster than RSA, wich is why it was more convenient for DPoP tokens, where the signature is verified on each request. However, I don't remember at all on which source I'm basing this, so it may very well be an error from me, and I'm happy to talk about it to learn more.

acoburn: Sounds great. In my experience, ES256 signatures are faster to transfer, verify and compute, though not by much. The real difference is in key generation, where ECDSA key pairs are much, much faster to create than RSA key pairs.

packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -165,10 +172,21 @@ export class SessionInfoManager implements ISessionInfoManager {
...
-      secure: true,
+      secure: false,
     });
 
+    if (
+      clientId === undefined &&
+      idToken === undefined &&
+      isLoggedIn === undefined &&
+      webId === undefined &&
+      refreshToken === undefined
```
pmcb55, line 185: Is it true now that a `refreshToken` is **_always_** going to be stored...?

pmcb55 (approved):
Looks great (just a quick question on the `refreshToken`).

packages/browser/src/sessionInfo/SessionInfoManager.ts
NSeydoux: No, to the contrary: we just return `undefined` if none of these values are stored, but if at least one of them is defined, then we return whatever we have. In the browser code, it's unlikely that we have a refresh token at all, and this is going to be removed when we clean up the code to remove support for them, but right now it doesn't hurt to check for it.

pmcb55: D'oh - yeah, of course - sorry. Yeah, it was the 'non-browser-ness' of the Refresh Token that piqued my attention here, but I mistakenly read the logic there as being `||` instead of `&&` for some reason.

PR #1081, opened 2021-02-20
Get issuer canonical IRI from its own config

NSeydoux:
The user provides an issuer IRI, which allows discovering the issuer
configuration. So far, this user-provided IRI was stored on login, and
used to validate the ID token. However, this IRI may be correct, but
still differ from the issuer claim of the ID token, e.g. depending on
whether it includes a trailing slash or not. Now, the IRI that gets
stored is the canonical IRI obtained from the IdP's config, which
enforces it to be aligned with the ID token's issuer claim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ correct issuer, and that it contains some expected values.
   stored in storage, ready to be retrieved again from storage when the login
   flow redirects back to the client application (previously it was only being
   stored if DCR was invoked).
+- The issuer URL associated to the session is now necessarily the __canonical__
+  issuer's URL, instead of potentially including/missing a trailing slash.
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ correct issuer, and that it contains some expected values.
   stored in storage, ready to be retrieved again from storage when the login
   flow redirects back to the client application (previously it was only being
   stored if DCR was invoked).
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ correct issuer, and that it contains some expected values.
   stored in storage, ready to be retrieved again from storage when the login
   flow redirects back to the client application (previously it was only being
   stored if DCR was invoked).
+- The issuer URL associated to the session is now necessarily the __canonical__
```
pmcb55, line 27:
```suggestion
- The issuer URL associated with the session is now necessarily the __canonical__
```

packages/node/src/login/oidc/OidcLoginHandler.ts
```diff
@@ -120,7 +120,7 @@ export default class OidcLoginHandler implements ILoginHandler {
 
     // Construct OIDC Options
     const oidcOptions: IOidcOptions = {
-      issuer: options.oidcIssuer,
+      issuer: issuerConfig.issuer,
```
pmcb55, line 123: Probably worth a comment here again, just 'cos the reasoning for **not** using the `options.oidcIssuer` is kinda subtle, and could easily be missed when looking at how the options are being set here.

pmcb55 (approved):

packages/node/src/login/oidc/OidcLoginHandler.ts
NSeydoux: Added a comment in b4e1bc67d3ca1e11503629cd605608bff98e44f8

PR #1097, opened 2021-02-23
R-enable running build script on install

NSeydoux:
When setting up husky v5, a behaviour of the build toolchain was
changed: the build script was no longer running when running the install
script. Because there are dependencies between the packages of the
monorepo, running this build script is actually useful, and this
re-enables this behaviour.


kay-kim (approved):
lgtm

Vinnl (approved):

PR #1100, opened 2021-02-23
Renamed master branch to main

NSeydoux:
This should replace all references to `master` in the codebase. Before merging this, I'll change the branch protection rules, and use GH's UI to rename the branch so that open PRs get migrated.

### DIFF
--- a/.github/publish-website.yml
+++ b/.github/publish-website.yml
@@ -3,7 +3,7 @@ name: Publish website
 on:
   push:
     branches:
-      - master
+      - main
 
 env:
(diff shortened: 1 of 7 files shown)


Vinnl (approved):
There's no need to manually change the branch protection rules, GitHub does that automatically for you: https://github.com/github/renaming

One thing though: if I search for `master` I also see `conf.py` fields referencing the branch. I think they only apply to links to edit a page (which I think are no longer shown), but to be sure I think it's preferable to update those as well.

Issue #1101, opened 2021-02-23
Client registration sends PKCE header


woutermont:
### Bug description
The client registrar sends a `code_challenge_method` header during Dynamic Client Registration (DCR). This header belongs to PKCE (in combination with a `code_challenge`), which plays no role in this stage. On certain IDPs, like KeyCloak, this makes DCR fail (in casu because the jackson parser cannot handle unrecognized headers). 

### To Reproduce
1. Log in to any IDP using this library
2. Check which headers are sent

### Expected result
Only necessary headers should be sent, so no  `code_challenge_method`.

### Actual result
The `code_challenge_method` header is sent during DCR, where it is not needed.

### Search terms I used
`pkce` `registration`

### Impacted package
`oidc-client-ext` `solid-client-authn-browser`

### Environment
```
System:
    OS: macOS 11.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 144.26 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.4 - ~/.nvm/versions/node/v14.15.4/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.10 - ~/.nvm/versions/node/v14.15.4/bin/npm
  Browsers:
    Brave Browser: 87.1.18.77
    Chrome: 88.0.4324.182
    Edge: 88.0.705.74
    Firefox: 84.0.2
    Safari: 14.0.3
  npmPackages:
    @inrupt/eslint-config-base: 0.0.4 => 0.0.4 
    @inrupt/eslint-config-lib: 0.1.1 => 0.1.1 
    @types/dotenv-flow: ^3.1.0 => 3.1.0 
    @types/estree: 0.0.46 => 0.0.46 
    @types/jest: ^26.0.13 => 26.0.20 
    @types/node-fetch: ^2.5.6 => 2.5.8 
    @typescript-eslint/eslint-plugin: ^4.14.2 => 4.15.2 
    @typescript-eslint/parser: ^4.14.2 => 4.15.2 
    dotenv-flow: ^3.2.0 => 3.2.0 
    eslint: ^7.19.0 => 7.20.0 
    eslint-config-airbnb-base: ^14.2.0 => 14.2.1 
    eslint-config-prettier: ^7.2.0 => 7.2.0 
    eslint-plugin-import: ^2.22.1 => 2.22.1 
[...]

PR #1102, opened 2021-02-23
fix: no PKCE headers in client registration

woutermont:
This PR fixes bug #1101.


NSeydoux:
For security reasons, Github prevents PRs from forks to run the CI. I re-applied your commit with my commiter permission in https://github.com/inrupt/solid-client-authn-js/pull/1103.

PR #1106, opened 2021-02-24
Do not enable session restore by default

Vinnl:
Attempting to restore a session loses the application state, so it
should be a conscious decision by developers to do so, and to take
the opportunity to preserve and restore the state themselves.

This fixes a bug introduced in 1.6.0.

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -375,7 +375,10 @@ describe("Session", () => {
         .fn()
         .mockResolvedValue(undefined);
       const mySession = new Session({ clientAuthentication });
-      await mySession.handleIncomingRedirect("https://some.redirect/url");
+      await mySession.handleIncomingRedirect({
+        url: "https://some.redirect/url",
+        restorePreviousSession: true,
+      });
       expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
     });
 
@@ -408,7 +411,10 @@ describe("Session", () => {
         .mockResolvedValue("https://some.issuer/");
 
       const mySession = new Session({ clientAuthentication });
-      await mySession.handleIncomingRedirect("https://some.redirect/url");
+      await mySession.handleIncomingRedirect({
+        url: "https://some.redirect/url",
+        restorePreviousSession: true,
+      });
 
       expect(window.localStorage.getItem(KEY_CURRENT_URL)).toEqual(
         "https://mock.current/location"
@@ -440,7 +446,10 @@ describe("Session", () => {
         .fn()
         .mockResolvedValue(undefined);
       const mySession = new Session({ clientAuthentication });
-      await mySession.handleIncomingRedirect("https://some.redirect/url");
+      await mySession.handleIncomingRedirect({
+        url: "https://some.redirect/url",
+        restorePreviousSession: true,
+      });
       expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
     });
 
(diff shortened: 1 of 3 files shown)


packages/browser/src/Session.ts
```diff
@@ -179,7 +193,8 @@ export class Session extends EventEmitter {
    * using the browser's current location.
    */
   handleIncomingRedirect = async (
-    url: string = window.location.href
+    url: string = window.location.href,
+    options: IHandleIncomgingRedirectOptions = {}
```
Vinnl, line 197: Actually, I'm thinking, now that `url` is optional, maybe we should make `options` the first argument and have `url` be one of its parameters (and allow it to be a string for backwards compatibility)?

Vinnl: Now added this as 57de44c7 - let me know what you think.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -485,6 +497,81 @@ describe("Session", () => {
...
+        .mockResolvedValue(undefined);
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect({
+        url: "https://some.redirect/url",
+        restorePreviousSession: false,
+      });
+
+      expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
```
NSeydoux, line 573:
```suggestion
      expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
      expect(clientAuthentication.login).not.toHaveBeenCalled();
```
I think adding this to the test would verify more striclty if the silent authentication has not been triggered.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -485,6 +497,81 @@ describe("Session", () => {
...
+        });
+      clientAuthentication.handleIncomingRedirect = jest
+        .fn()
+        .mockResolvedValue(undefined);
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.redirect/url");
+
+      expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
```
NSeydoux, line 534: See the comment below to check that silent authentication has not been triggered.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -485,6 +497,81 @@ describe("Session", () => {
         redirectUrl: "https://some.redirect/url",
       });
     });
+
+    it("does nothing if the developer has not enabled session restore", async () => {
```
NSeydoux, line 501:
```suggestion
    it("does nothing if the developer has not explicitly enabled silent authentication", async () => {
```

packages/browser/__tests__/Session.spec.ts
```diff
@@ -485,6 +497,81 @@ describe("Session", () => {
...
+        .mockResolvedValue(undefined);
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.redirect/url");
+
+      expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
+    });
+
+    it("does nothing if the developer has disabled session restore", async () => {
```
NSeydoux, line 537:
```suggestion
    it("does nothing if the developer has disabled silent authentication", async () => {
```

packages/browser/__tests__/Session.spec.ts
NSeydoux: Since the API option is `restorePreviousSession`, which is better for developers, I understand why this mentions this term too. However, I find it a little clearer to call it silent authentication from an internal point of view, but that may be because I'm just more familiar with the term.

NSeydoux (approved):

packages/browser/__tests__/Session.spec.ts
Vinnl: Actually that was my initial plan, but then I saw that other "it does nothing" tests added this check :) 4d5b2132

packages/browser/__tests__/Session.spec.ts
Vinnl: No makes sense, we're also using it for the function name: b3518237

PR #1109, opened 2021-02-25
Re-enable workaround with disabled silent refresh

Vinnl:
This reverts d7332cf00317a7a3bc1f3684c01916bd21f89ace, but adds
that the Resource Server cookie-based workaround is only enabled
when silent token refresh is not enabled, to ensure that that is a
non-breaking bugfix.

This means I only added a single test (Session "does not attempt to use the workaround if the long-term solution (silent refresh) is enabled") and the second condition in

```js
    if (
      typeof storedSessionCookieReference === "string" &&
      options.restorePreviousSession !== true
    ) {
```

### DIFF
--- a/packages/browser/src/sessionInfo/SessionInfoManager.ts
+++ b/packages/browser/src/sessionInfo/SessionInfoManager.ts
@@ -55,9 +55,21 @@ export async function clear(
   sessionId: string,
   storage: IStorageUtility
 ): Promise<void> {
+  const storedSessionCookieReference = await storage.get(
+    "tmp-resource-server-session-info"
+  );
+  const reference = JSON.parse(storedSessionCookieReference ?? "{}");
+  const { webId } = reference;
+  if (webId !== undefined) {
+    const webIdAsUrl = new URL(webId);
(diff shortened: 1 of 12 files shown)


packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -55,9 +55,21 @@ export async function clear(
...
+  if (webId !== undefined) {
+    const webIdAsUrl = new URL(webId);
+    const resourceServerIri = webIdAsUrl.origin;
+    await storage.clearResourceServerSessionInfo(resourceServerIri);
+  }
   await Promise.all([
     storage.deleteAllUserData(sessionId, { secure: false }),
     storage.deleteAllUserData(sessionId, { secure: true }),
+    // FIXME: This is needed until the DPoP key is stored safely
+    storage.delete("clientKey", { secure: false }),
```
NSeydoux, line 72:
```suggestion
```

This does not need to be reverted (it should even have been removed long before the workaround cleanup ^^)

NSeydoux (approved):

PR #1110, opened 2021-02-25
Don't fail the build when coverage upload fails

Vinnl:
The upload-artifact action is notoriously unreliable, but it is not
important enough to break the build when it fails. Rather, we can
just restart the build in the rare occasion that we do need those
artifacts and the upload fails.

### DIFF
--- a/.github/workflows/cd-preview.yml
+++ b/.github/workflows/cd-preview.yml
@@ -106,6 +106,7 @@ jobs:
         VERSION_NR: ${{ needs.dev-release-npm.outputs.version-nr }}
     - name: Archive Parcel build artifacts
       uses: actions/upload-artifact@v2.2.2
+      continue-on-error: true
       with:
         name: parcel-dist
         path: .github/workflows/cd-packaging-tests/bundler-parcel/dist
@@ -129,6 +130,7 @@ jobs:
         VERSION_NR: ${{ needs.dev-release-npm.outputs.version-nr }}
     - name: Archive Webpack build artifacts
       uses: actions/upload-artifact@v2.2.2
+      continue-on-error: true
       with:
         name: webpack-dist
         path: .github/workflows/cd-packaging-tests/bundler-webpack/dist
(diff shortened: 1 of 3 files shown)


pmcb55:
Great, as this is a pain-in-the-ass. But is it possible to make it 'retry-on-error' as opposed to 'continue-on-error'?
For instance, when you say "**_we can just restart the build in the rare occasion that we do need those artifacts and the upload fails_**" - does that mean that on those rare occasions, we need to **_remember_** to go check if indeed they were uploaded or not (since the build won't tell us anymore - i.e., it'll always just continue regardless).
In other words, if it's one of those rare occasions, we can't just assume the artifacts were uploaded if the build passed - we'd need to know that if they aren't there, and yet the build passed, we may need to just re-run the build, or go into the CI logs and see if it was this particular upload job that failed.
In other other words (!) - it sounds like this would need to be documented somewhere too!

Vinnl:
Agreed that that would be desirable, but unfortunately GitHub Actions doesn't support that, as far as I can see. (I mean, it doesn't even support restarting _individual_ jobs :roll_eyes: )

As for needing documentation: I think it speaks for itself. If you go looking for the coverage reports in the CI job, and instead you see that the job has failed but continued anyway, that's the obvious thing to do. I'm not sure where I could write those docs that would be a more obvious place that you'd look for than that :) I can add a comment to the CI scripts at least, if that's what you're thinking of?

pmcb55:
Yeah, exactly - just documenting in the CI scripts is a good start.
But how about creating a README.md in the `./github` folder, and describing all the CI-specific things we do (would be great for new-comers to the codebase!)...?
I've no idea if that's a 'good' or a 'bad' practice to be dumping files in the folder 'owned' by GitHub, but it sounds like a great place to record all these little potential gotcha's, e.g., why we don't run the browser tests on node v12 or v10 (i.e., it makes perfect sense why, but it confused both me and Jack separately when we just looked at the high-level CI jobs, and saw v12 and v10 passing, but v14 failing), or why we switched to using explicit version numbers on GitHub actions like `- uses: actions/checkout@v2.3.4` (see, I can't remember the reason for that now!).
We could also document in there what we do around .codesandbox, and the fact that we run tests from deployed npm packages too. So yeah, seems there's lots of CI-related information 'in our heads' at the moment that needs to written up.

NSeydoux (approved):

NSeydoux:
I created a ticket to document the CI design choices.

PR #1111, opened 2021-02-26
Removed duplication between redirectUr(i|l)

NSeydoux:
Parts of the code reads and writes the redirect URI in storage with the key `redirectUri`, and some other parts with the key `redirectUrl`. This causes confusion, and ends up with potentially creating a bug when trying to use one in silent authentication when the other is the one that has been registered as valid by the Solid Identity Provider, leading to the user being displayed an `invalid_grant` screen instead of being redirected to the app.

Technically, either `redirectUrl` or `redirectUri` are valid choices.
`redirectUrl` has been picked because it is the name of the parameter in
the login options, so picking `redirectUri` woud need either breaking
change, or to keep some room for error by having to rename the field in
the JSON object at some point. Note that the actual OAuth parameter name
is `redirect_uri`, which isn't camel cased, so a renaming would be
necessary anyways.

### DIFF
--- a/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
+++ b/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
@@ -102,15 +102,15 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
   }
 
   async handle(
-    redirectUrl: string
+    inputRedirectUrl: string
   ): Promise<ISessionInfo & { fetch: typeof fetch }> {
-    if (!(await this.canHandle(redirectUrl))) {
+    if (!(await this.canHandle(inputRedirectUrl))) {
       throw new Error(
-        `AuthCodeRedirectHandler cannot handle [${redirectUrl}]: it is missing one of [code, state].`
+        `AuthCodeRedirectHandler cannot handle [${inputRedirectUrl}]: it is missing one of [code, state].`
       );
     }
 
-    const url = new URL(redirectUrl);
+    const url = new URL(inputRedirectUrl);
     // The type assertion is ok, because we checked in canHandle for the presence of a state
     const oauthState = url.searchParams.get("state") as string;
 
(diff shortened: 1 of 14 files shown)


NSeydoux:
I tested this against a local deployment of [Penny](https://gitlab.com/vincenttunru/penny), and it looks like it resolves the issue, but I'd prefer that you confirm that @Vinnl.

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -102,15 +102,15 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
 
   async handle(
-    redirectUrl: string
+    inputRedirectUrl: string
   ): Promise<ISessionInfo & { fetch: typeof fetch }> {
-    if (!(await this.canHandle(redirectUrl))) {
+    if (!(await this.canHandle(inputRedirectUrl))) {
       throw new Error(
-        `AuthCodeRedirectHandler cannot handle [${redirectUrl}]: it is missing one of [code, state].`
+        `AuthCodeRedirectHandler cannot handle [${inputRedirectUrl}]: it is missing one of [code, state].`
```
pmcb55, line 109: Hhmmm... interesting code smell here - i.e., this `handle()` method is explaining why a completely different method, `canHandle()`, has returned `false`. Clearly it's not this function's job to be doing that, as once a developer tweaks the `canHandle()` method to add some new 'I can't handle this' condition, and naturally doesn't realize they also need to update this completely different method accordingly, this error message becomes dangerously misleading.
No easy fix here I don't think, but a good example of why the LIT always passed a mutable 'Result' Dataset as an input parameter all along the request processing call stack - i.e., so that each and every processing stage could independently report it's behaviour all within a single 'Result' object (just by using RDF), and why nested errors are so important too (although in this particular case, `canHandle()` is not throwing an error - so the mutable, extensible, self-descriptive 'Result' object would come in here instead).
In this case, I suppose we should add a comment to the `canHandle()` return statement pointing out that the `handle()` method is currently describing this 'missing one of [code, state]' behaviour...?

pmcb55 (approved):

Vinnl:
Can confirm that I no longer experience the issue using this branch :+1:

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: I think the easiest approach would be to change the way the handler pattern is currently implemented to have a separate validation function (currently `canHandle`) that throws when the validation fails, with an error message describing the failure. Then, `canHandle` just has to call this function, and catch the error if a failure happens to still return false (because it's expected that `canHandle` may return false without stopping the app), and `handle` can also call this function, and not catch the error if any. This way, only the validation function implements the validation logic and the attached error message. Since that's something we'd need to apply to all implementations of the handler pattern, I'll add a ticket to this end.

PR #1125, opened 2021-03-02
Enable ESS cookie auth opt out

NSeydoux:
So far, logging in to any Identity Provider would lead to the Resource
Server hosting the WebID being looked up for a `/session` endpoint. This
causes several issues:
- It makes additional, non-required requests if the server does not have
a session endpoint, or if the app supports silent authentication
- It has caused annoying issues when looking up the missing endpoint for
some servers
- It's a solution we want to deprecate anyways.

To resolve this, both the endpoint lookup and the cookie-based auth can
be disabled by specifying an additional option to
`handleIncomingRedirect`, `useEssSession`. It defaults to true (for
backward-compatibility), but it is considered false if silent
authentication is enabled.

Note that this uses an entry in local storage to share the state between
the Session where the option is specified and the RedirectHandler where
the session lookup actually happens. This isn't ideal, but it prevents
from changing internal APIs, only to change them back when this isn't
supported anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### New features
 
-- New options `useEssSession` for `session.handleIncomingRedirect`: If set to false, 
-the `/session` endpoint isn't looked up, and cookie-based auth is disabled. The 
+- New option `useEssSession` for `session.handleIncomingRedirect`: Control to 
+enable and disble the behaviour introduced in 1.4.0. If set to false, the
+`/session` endpoint isn't looked up, and cookie-based auth is disabled. The 
 behaviour is similar when `restorePreviousSession` is true.
 
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### New features
+
+- New options `useEssSession` for `session.handleIncomingRedirect`: If set to false, 
+the `/session` endpoint isn't looked up, and cookie-based auth is disabled. The 
+behaviour is similar when `restorePreviousSession` is true.
+
 ### Bugfixes
 
 #### browser
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -419,6 +419,29 @@ describe("Session", () => {
         ).toHaveBeenCalledTimes(1);
       });
 
+      it("does not attempt to use the workaround if it is explicitly disabled", async () => {
+        mockLocalStorage({
+          "tmp-resource-server-session-info": JSON.stringify({
+            webId: "https://my.pod/profile#me",
+            sessions: {
+              "https://my.pod/": { expiration: 9000000000000 },
+            },
+          }),
+        });
+        const clientAuthentication = mockClientAuthentication();
+        clientAuthentication.handleIncomingRedirect = jest.fn();
+        const mySession = new Session({ clientAuthentication });
+        await mySession.handleIncomingRedirect({
+          url: "https://some.url",
+          useEssSession: false,
+        });
+        expect(mySession.info.isLoggedIn).toBe(false);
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### New features
+
+- New options `useEssSession` for `session.handleIncomingRedirect`: If set to false,
```
Vinnl, line 12:
```suggestion
- New option `useEssSession` for `session.handleIncomingRedirect`: If set to false, 
```

And maybe a short mention of how this behaviour was introduced in 1.4.0 so people can reference back to that would be helpful too :)

packages/browser/src/Session.ts
```diff
@@ -65,6 +65,17 @@ export interface IHandleIncomgingRedirectOptions {
...
   restorePreviousSession?: boolean;
+
+  /**
+   * Inrupt's Enterprise Solid Server sets a cookie to allow the browser to
+   * access private resources on a Pod. In order to mitigate the logout-on-refresh
+   * issue on the short term, the server also implemented a session endpoint
+   * enabling the client app to know whether the cookie is set. If your app
+   * supports the newest session restore approach, or if you want to prevent
+   * your app to look up the session endpoint on your resource server, this
+   * option may be set to false. It defaults to true for backward compatibility.
```
Vinnl, line 76: I'd think I'd only mention that you might want to set this to `false` if you want to prevent this lookup, and then explicitly call out separately that this value will always be overridden to `false` if `restorePreviousSession` is `true`. Since it can be confusing when an option doesn't have the stated effect.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -255,12 +255,18 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
-      tokens.webId,
-      authFetch,
-      this.storageUtility
-    );
+    // TODO: This is a temporary workaround. When deprecating the cookie-based auth,
+    // this should be all cleared.
+    const essWorkaroundDisabled =
+      window.localStorage.getItem("tmp-resource-server-session-enabled") ===
+      "false";
+    if (!essWorkaroundDisabled) {
```
Vinnl, line 263: Ha, so if the disabled value that represents whether the enabled LocalStorage option is equal to the string "false" is not true, this block holds? :P 

(It's readable just fine, I was just amused by the quadruple negative :) )

packages/browser/__tests__/Session.spec.ts
```diff
@@ -419,6 +419,29 @@ describe("Session", () => {
...
+              "https://my.pod/": { expiration: 9000000000000 },
+            },
+          }),
+        });
+        const clientAuthentication = mockClientAuthentication();
+        clientAuthentication.handleIncomingRedirect = jest.fn();
+        const mySession = new Session({ clientAuthentication });
+        await mySession.handleIncomingRedirect({
+          url: "https://some.url",
+          useEssSession: false,
```
Vinnl, line 436: I suppose another test to verify that it does not attempt to use the workaround if it is enabled but silent authentication is enabled too might be a good addition, or is that very complex due to having to mock silent authentication?

Vinnl (approved):

CHANGELOG.md
NSeydoux: Thanks for the suggestions :) 6b9de5b95196d8ba9c55ffa51e18da875248887f

packages/browser/src/Session.ts
NSeydoux: Clarified in b2f97b9b387a44bac4191ddb5586f86ed61bd549

packages/browser/__tests__/Session.spec.ts
NSeydoux: Good point, and it's not that difficult to mock: 27cb103ec419270d73d6283dfd389e26b3a70e4a. The assumption is that if the workaround is used, the function returns before calling `ClientAuthentication::handleIncomingRedirect`, the result of which will lead either to being logged in, or to attempt silent authentication. Also, I added tests to verify the value of the variable in local storage.

PR #1130, opened 2021-03-04
Fix typo in interface name

Vinnl:
Technically someone could be using the interface somewhere in the code and this would be breaking, but in practice I'm quite sure nobody does, so as far as I'm concerned we don't need to preserve the one with the typo. Happy to add it if you disagree though.


NSeydoux (approved):

NSeydoux:
No I agree, the release is recent enough, and the need to use this interface slim enough that it's not worth the extra work.

PR #1143, opened 2021-03-05
Enable lerna-audit in CI

NSeydoux:
lerna requires a specific audit configuration, enabled by the package
lerna-audit. It now runs in CI along with the regular `npm audit`.
The script is also now named `lerna-audit` instead of `audit` to avoid
the confusion with the `npm audit` command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
           E2E_TEST_ESS_POD: ${{ secrets.E2E_TEST_ESS_POD }}
       - run: npx prettier --check "{packages/*/src,packages/*/__tests__}/**"
       - run: npm audit --audit-level=moderate
+      - run: npm run lerna-audit -- --audit-level=moderate
       - name: Archive code coverage results
         uses: actions/upload-artifact@v2.2.2
         continue-on-error: true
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -38,6 +38,7 @@ jobs:
           E2E_TEST_ESS_POD: ${{ secrets.E2E_TEST_ESS_POD }}
       - run: npx prettier --check "{packages/*/src,packages/*/__tests__}/**"
       - run: npm audit --audit-level=moderate
+      - run: npm run lerna-audit -- --audit-level=moderate
```
pmcb55, line 41: This looks like we should perhaps have a Lerna section in our 'Coding Guidelines' - i.e., I assume this should be our setup for all Lerna projects at Inrupt (i.e., I believe Arne is using Lerna too), and also to answer why 'moderate'...?

pmcb55 (approved):

.github/workflows/ci.yml
NSeydoux: To be honest, I copy-pasted the line above, so I'm not sure about `moderate`, but having general Lerna guidelines sounds like a good thing, even if it is just to capitalize knowledge.

PR #1154, opened 2021-03-09
Discover support for solid-oidc in OIDC provider

NSeydoux:
This is the preliminary step to figure out how to interact with a Solid
Identity Provider: when fetching its description as part of the OIDC
discovery process (see
https://openid.net/specs/openid-connect-discovery-1_0.html for more
informaiton), support for Solid-OIDC is specified with the
`solid_oidc_supported` key, as specified in
https://solid.github.io/authentication-panel/solid-oidc/#discovery.

Additional PRs will take this configuration into account to perform DCR
or present a client WebID, the latter being only supported by Solid-OIDC
conformant OIDC providers.

# Checklist

### DIFF
--- a/packages/core/src/login/oidc/IIssuerConfig.ts
+++ b/packages/core/src/login/oidc/IIssuerConfig.ts
@@ -67,4 +67,5 @@ export interface IIssuerConfig {
   requireRequestUriRegistration?: boolean;
   opPolicyUri?: string;
   opTosUri?: string;
+  solidOidcSupported?: string;
 }
--- a/packages/core/src/login/oidc/IIssuerConfig.ts
+++ b/packages/core/src/login/oidc/IIssuerConfig.ts
@@ -67,5 +67,5 @@ export interface IIssuerConfig {
   requireRequestUriRegistration?: boolean;
   opPolicyUri?: string;
   opTosUri?: string;
-  solidOidcSupported?: boolean;
+  solidOidcSupported?: "https://solidproject.org/TR/solid-oidc";
 }
--- a/packages/core/src/login/oidc/IIssuerConfig.ts
+++ b/packages/core/src/login/oidc/IIssuerConfig.ts
@@ -67,5 +67,5 @@ export interface IIssuerConfig {
   requireRequestUriRegistration?: boolean;
   opPolicyUri?: string;
   opTosUri?: string;
-  solidOidcSupported?: string;
+  solidOidcSupported?: boolean;
 }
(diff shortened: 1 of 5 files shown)


packages/core/src/login/oidc/IIssuerConfig.ts
```diff
@@ -67,4 +67,5 @@ export interface IIssuerConfig {
   requireRequestUriRegistration?: boolean;
   opPolicyUri?: string;
   opTosUri?: string;
+  solidOidcSupported?: boolean;
```
Vinnl, line 70: Hmm, am I misreading the spec if I think this should be:

```suggestion
  solidOidcSupported?: "https://solidproject.org/TR/solid-oidc";
```

?

Vinnl (commented):
Just added a question that I think would affect the entire PR, so probably best to clarify that first before I review further :)

Edit: Just saw d53ffecf2165ce4ce3547fc36b30163c9176d041 so I presume not, but would be good for the review to know why not.

packages/core/src/login/oidc/IIssuerConfig.ts
pmcb55: Yeah, it's definitely a good point - I'm assuming Nic is making the assumption that the presence of *any* value means that Solid OIDC is supported by the IdP, and the particular value isn't relevant (the Solid OIDC spec could make that clearer perhaps, by explicitly stating that a IdP MUST NOT provide any value at all if they don't support it (i.e., stating "solid_oidc_supported: "Not supported"" would be non-conformant). I'd be Ok with that, but I would definitely add a comment explaining that. I created a spec issue [here](https://github.com/solid/specification/issues/249).

acoburn: You'll want to verify that the _value_ is equal to that particular URL, not just that the key exists.

NSeydoux: You're absolutely right, that's actually why I initially made it a string, and somehow I was jumping aound this morning and forgot about that \o/

NSeydoux: Fixed in https://github.com/inrupt/solid-client-authn-js/pull/1154/commits/cbc5de1f922b34b675a070e88122067704bfc537

Vinnl (approved):
I presume that validation that the issuer config is a valid `IIssuerConfig` will happen elsewhere?

NSeydoux:
Indeed, the conformance of the issuer configuration is already checked, this just adds an additional supported field. If I'm not mistaken, unknown fields are accepted, they just should be discarded by the client, and we're using our underlying OIDC libraries to validate the spec conformance of the client configuration we get.

PR #1158, opened 2021-03-10
Client WebID support for the browser login

NSeydoux:
This adds support for client WebID in the login process. There are a
certain number of configurations to consider: either the IdP supports
Solid-OIDC or not, and either the client provides a client_id or not.

If no client_id is provided, dynamic client registration is required,
regardless of the IdP. Most clients fall into this category today. If a
client ID is provided and it doesn't look like an IRI, or if it is an
IRI and the IdP supports Solid-OIDC, the client can be considered
already registered and there is no additional interaction required.

The library only has to be careful when the client provides a client
WebID, but the IdP soes not support Solid-OIDC.In that case, presenting
the client_id to the IdP would fail, because it would not look up the
app profile, but expect the given client_id to have previously been
registered. In that case, the client must go through DCR, and will be
considered like a public client by the resource server.

Note to reviewers: This commit only applies to `@inrupt/solid-client-authn-browser`.
The node equivalent code will be submitted in a subsequent commit.

# Checklist

### DIFF
--- a/packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
@@ -71,12 +71,14 @@ describe("OidcLoginHandler", () => {
     expect(actualHandler.handle.mock.calls).toHaveLength(1);
   });
 
-  it("should lookup client ID if not provided", async () => {
+  it("should retrieve client ID from storage if one is not provided", async () => {
     const actualHandler = defaultMocks.oidcHandler;
     const mockedStorage = new StorageUtility(
       mockStorage({}),
       mockStorage({
         "solidClientAuthenticationUser:mySession": {
+          // The value of the client ID doesn't matter here, and it could be a WebID.
+          // This checks it gets passed through from storage to the handler.
           clientId: "some client ID",
         },
       })
@@ -148,6 +150,8 @@ describe("OidcLoginHandler", () => {
       clientRegistrar: new ClientRegistrar(mockedEmptyStorage),
       issuerConfigFetcher: mockIssuerConfigFetcher({
         ...IssuerConfigFetcherFetchConfigResponse,
+        // Solid-OIDC is not supported by the Identity Provider, so the provided
+        // client WebID cannot be used, and the client must go through DCR instead.
         solidOidcSupported: undefined,
       }),
     });
@@ -157,7 +161,7 @@ describe("OidcLoginHandler", () => {
       oidcIssuer: "https://arbitrary.url",
       redirectUrl: "https://app.com/redirect",
       tokenType: "DPoP",
-      clientId: "https://my.app/registration",
+      clientId: "https://my.app/registration#app",
     });
 
     const calledWith = actualHandler.handle.mock.calls[0][0];
@@ -208,17 +212,17 @@ describe("OidcLoginHandler", () => {
       oidcIssuer: "https://arbitrary.url",
       redirectUrl: "https://app.com/redirect",
       tokenType: "DPoP",
-      clientId: "https://my.app/registration",
+      clientId: "https://my.app/registration#app",
     });
 
     const calledWith = actualHandler.handle.mock.calls[0][0];
-    expect(calledWith.client.clientId).toBe("https://my.app/registration");
+    expect(calledWith.client.clientId).toBe("https://my.app/registration#app");
 
     const storedClientId = await mockedStorage.getForUser(
       "mySession",
       "clientId"
     );
-    expect(storedClientId).toEqual("https://my.app/registration");
+    expect(storedClientId).toEqual("https://my.app/registration#app");
   });
 
   it("should save client ID, secret and name if given as input options", async () => {
--- a/packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
+++ b/packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
@@ -148,7 +148,7 @@ describe("OidcLoginHandler", () => {
       clientRegistrar: new ClientRegistrar(mockedEmptyStorage),
       issuerConfigFetcher: mockIssuerConfigFetcher({
         ...IssuerConfigFetcherFetchConfigResponse,
-        solidOidcSupported: false,
+        solidOidcSupported: undefined,
       }),
     });
 
@@ -199,7 +199,7 @@ describe("OidcLoginHandler", () => {
       clientRegistrar: new ClientRegistrar(mockedStorage),
       issuerConfigFetcher: mockIssuerConfigFetcher({
         ...IssuerConfigFetcherFetchConfigResponse,
-        solidOidcSupported: true,
+        solidOidcSupported: "https://solidproject.org/TR/solid-oidc",
       }),
     });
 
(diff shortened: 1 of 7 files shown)


packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
```diff
@@ -77,7 +77,7 @@ describe("OidcLoginHandler", () => {
       mockStorage({}),
       mockStorage({
         "solidClientAuthenticationUser:mySession": {
-          clientId: "https://some.app/registration",
+          clientId: "some client ID",
```
pmcb55, line 80: Why change this from a URI to a non-URI, since isn't this test just checking that whatever was stored is just being passed through? If the intent is really to use a non-URI, I'd change the value from "some client ID" (which tells me nothing), to something like "value of client ID doesn't matter - just checking it gets passed through from storage to the handler", or something like that...
But if the client ID is generally meant to be a URI (in fact a WebID), then I'd change the value to be a URI, but still communicate the intent of the test, e.g., "https://clientAppWebID.com/value-stored-in-storage-gets-passed-to-handler".

packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
```diff
@@ -152,6 +189,38 @@ describe("OidcLoginHandler", () => {
...
+        solidOidcSupported: true,
+      }),
+    });
+
+    await handler.handle({
+      sessionId: "mySession",
+      oidcIssuer: "https://arbitrary.url",
+      redirectUrl: "https://app.com/redirect",
+      tokenType: "DPoP",
+      clientId: "https://my.app/registration",
```
pmcb55, line 211: Just to emphasise that client ID's are expected to be WebID's (should app WebID's end with `#app`...??):
```suggestion
      clientId: "https://my.app/profile/card#app",
```

packages/browser/src/login/oidc/OidcLoginHandler.ts
```diff
@@ -44,6 +44,68 @@ import {
...
+    (issuerConfig.solidOidcSupported === false && isValidUrl(options.clientId))
+  ) {
+    // If no client_id is provided, the client must go through DCR. Whether the
+    // Identity Provider supports Solid-OIDC or not will only change the value
+    // of the provided client_id.
+    // The tricky case is if a client WebID (here, a client_id looking like an IRI)
+    // is provided, but the Identity Provider does not support Solid-OIDC. In this
+    // case, we must discard the provided client_id, and still go through dynamic
+    // client registration, because the Identity Provider will not recognize the
+    // provided client_id.
```
pmcb55, line 76: I *think* this wording might be clearer (in conjunction with the new wording in the comment below too) - but see what you think...
```suggestion
    // If no client_id is provided, the client must go through DCR. If a
    // client_id is provided and it looks like a URI, yet the Identity Provider does *not* support Solid-OIDC, then
    // we also perform DCR (and discard the provided client_id).
```

packages/browser/src/login/oidc/OidcLoginHandler.ts
```diff
@@ -44,6 +44,68 @@ import {
...
+        sessionId: options.sessionId,
+        clientName: options.clientName,
+        redirectUrl: options.redirectUrl,
+      },
+      issuerConfig
+    );
+  }
+  // If the Identity Provider is Solid-OIDC compliant, or if the client_id isn't
+  // an IRI and therefore has been previously registered to the IdP, the client
+  // registration information just need to be stored to be retrieved after redirect.
```
pmcb55, line 88:
```suggestion
  // If a client_id was provided, and the Identity Provider is Solid-OIDC compliant (or it's not
  // compliant but the client_id isn't an IRI (meaning this client probably already registered with
  // the IdP via DCR)), then the client registration information needs to be stored so that it can
  // be retrieved later after redirect.
```

pmcb55 (approved):
Looks great - just some suggested comment tweaks...

packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
```diff
@@ -129,7 +129,44 @@ describe("OidcLoginHandler", () => {
     expect(mockedOidcModule.registerClient).toHaveBeenCalled();
   });
 
-  it("should save client ID if given one as an input option", async () => {
+  it("should perform DCR if a client WebID is provided, but the target IdP does not support Solid-OIDC", async () => {
+    const mockedOidcModule = jest.requireMock("@inrupt/oidc-client-ext");
+    mockedOidcModule.registerClient = jest.fn().mockResolvedValue({
+      clientId: "some dynamically registered ID",
```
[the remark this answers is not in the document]
NSeydoux: As for the previous comment, I'd rather have a separate comment, but keep the value more compact. 0f5f5d58f80952d1905cb73a4e7aad9b6016e3ed adds a comment.

packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
NSeydoux: I added a hash fragment in the client WebIDs in 0f5f5d58f80952d1905cb73a4e7aad9b6016e3ed, but I'd prefer to avoid associating the notion of WebID to the `/profile/card` IRI template, which is merely a convention, but isn't something that we should necessarily expect.

packages/browser/src/login/oidc/OidcLoginHandler.ts
NSeydoux: 0f4c45b5d22c2d0fa8f3d8ccc9055cce155fae14

packages/browser/src/login/oidc/OidcLoginHandler.ts
NSeydoux: 0f4c45b5d22c2d0fa8f3d8ccc9055cce155fae14

NSeydoux:
I think this PR could be closed, and replaced by https://github.com/inrupt/solid-client-authn-js/pull/1165. Do you agree ?

packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
pmcb55: Yeah, sure - my suggestion was just to make a point really, i.e., I wouldn't have suffixed the example value with "...just checking it gets passed through from storage to the handle" either. I totally agree about not using the value as a full-blown comment, but I would use the value to succinctly show intent, so "value doesn't matter" is something I use a lot (i.e., it's very generic, but communicates intent (from the test's perspective)). The value "some client ID" works generally too (although I'd suggest "Client ID doesn't matter" is a just a tiny bit better again!), but just in this particular case it kinda jumped out at me 'cos I expected a client ID to be WebID - but in this case (i.e., a generic authn library), it can indeed be a WebID or any arbitrary string value, so it's fine.

packages/browser/__tests__/login/oidc/OidcLoginHandler.spec.ts
pmcb55: Sure, the `#app` is better (since it's the fragment part that makes the IRI a non-information resource, which is what a WebID really is). Personally, I still like conforming to the current convention of `/profile/card` (and extending that convention with `#app` for client app WebID's maybe), as I think having a convention (whatever that convention might be) still helps in communicating the intent (i.e., if I see `/profile/card` on *anything*, I'll automatically assume it's a WebID), and I think there's inherent value in that. But I'm still fine with you not doing so either (since it is 'just' a convention at the moment), and I've never really liked how long adding `/profile/card` makes a WebID anyways (and needing to remember whether it's `/profile/card` or `/card/profile` !).

PR #1159, opened 2021-03-10
Normalize redirect IRI in node package

NSeydoux:
- The IRI wasn't normalized on input
- The IRI parameters added by the Solid Identity Provider were not
removed from the redirect_iri specified at the token endpoint. This was causing a mismatch between the provided `redirect_iri` and the expected value, equal to the `redirect_iri` provided during registration.

These two bugs combined could cause the behaviour observed by @jeff-zucker in https://github.com/solid/node-solid-server/issues/1580.

### DIFF
--- a/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -311,6 +311,7 @@ describe("AuthCodeRedirectHandler", () => {
       expect(callback).toHaveBeenCalledWith(
         "https://my.app/redirect",
         { code: "someCode", state: "someState" },
+        // The code verifier comes from the mocked storage.
         { code_verifier: "some code verifier", state: "someState" },
         expect.anything()
       );
--- a/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -196,7 +196,7 @@ describe("AuthCodeRedirectHandler", () => {
       },
     });
 
-  const setupOidcClientMock = (tokenSet: TokenSet) => {
+  const setupOidcClientMock = (tokenSet?: TokenSet, callback?: any) => {
     const { Issuer } = jest.requireMock("openid-client");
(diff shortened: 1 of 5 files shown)


packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -293,6 +293,29 @@ describe("AuthCodeRedirectHandler", () => {
...
+      });
+
+      await authCodeRedirectHandler.handle(
+        "https://my.app/redirect?code=someCode&state=someState"
+      );
+
+      expect(callback).toHaveBeenCalledWith(
+        "https://my.app/redirect",
+        { code: "someCode", state: "someState" },
+        { code_verifier: "some code verifier", state: "someState" },
```
pmcb55, line 314: It's not clear from this test where the value "some code verifier" is coming from (i.e., I can see where the values on the prevous line are coming from (i.e., query params on the URI on line 308), but not this value)? Maybe just a comment to explain that the `code_verifier` value is coming from a mock (I assume!)...?

pmcb55 (approved):

jeff-zucker:
I cloned the solid-client-authn-js repo that inclueds this PR, built a version of generate-oidc-token using that repo, ran the script with the same reported error but a difference in behavior from yesterday -- before I was sent to the solidcommunity.net login page, but now I get a blank page in the browser and the same "login successful" and "mismatched redirect uri" in the console.

NSeydoux:
Thanks for following up ! I can't reproduce the error with this fix in... I cloned https://github.com/inrupt/generate-oidc-token, and ran:
- `npm ci`
- `npm install @inrupt/solid-client-authn-node@main`
- `npm run build`
- `node dist/index.js`
I can get a refresh token without errors then. Can you describe how to reproduce the error you encounter ?

jeff-zucker:
I may have screwed up the test, so if the patch worked for you, please just release it and I'll check from the npm version.  

If you care, this was my process:  I cloned the solid-client-authn-js/main repo; npm saved generate-oidc into it; modified and rebuilt generate-oidc-token/src/index.ts to use that version of solid-client-authn-node; ran node dist/index.js.

jeff-zucker:
Also, forgot to mention that I built both generate-oidc-token and solid-client-authn-js-node before testing.

PR #1187, opened 2021-03-18
Slow down e2e test so that NSS can keep up

NSeydoux:
If the tests are too fast paced, some fail to authenticate to NSS. By
trial and error, 17 seconds (!) seems to be a safe value to wait for
authentication to complete.


Vinnl:
Since tests are succeeding now even without the timeout, I guess this requires further investigation before merging this? The dotenv-flow addition looks good though.

NSeydoux:
Well it looks like the test that went through was a one-off, and all the PRs that are currently open still fail the CI. I suggest we do merge this, which resolved the problem for me locally and looks like it resoves it in CI (the CI on this PR runs using the values from the current branch, right ?), and see if it resolves the issue in other branches.

Vinnl (approved):
OK. I'd still like to know the cause of the slowdowns, and 17 seems a bit too arbitrary to my comfort levels, but it's no biggie. (One other thing we could consider is only adding the huge expiration time for NSS?)

PR #1204, opened 2021-03-24
Run dependabot CI without GitHub secrets

NSeydoux:
Due to
https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/,
Dependabot PRs don't have access to GH secrets. In particular, that
means they can't run e2e tests, and can't deploy a preview app on NPM.
So far, we have manually re-run the dependabot PRs to get access to the
secrets. This commit makes it so that Dependabot PRs run without e2e
tests or NPM preview, so that a human user does not need to be involved.


.github/workflows/ci.yml
```diff
@@ -29,7 +29,16 @@ jobs:
...
+      - run: |
+          # Unfortunately GitHub Actions does not currently let us do something like
+          #     if: secrets.NPM_TOKEN != ''
+          # so simply skip the command if the env var is not set:
+          if [ -n $E2E_TEST_REFRESH_TOKEN ]; then
+          npm run e2e-test
+          fi
+        if:
+          # Testing for the secret is based on a bash expression, which isn't portable. 
+          ${{ matrix.os == 'ubuntu-20.04' }}
```
Vinnl, line 41: This does mean that the end-to-end tests won't run at all on Windows any more, right? How much of a problem do you expect that to be? (I guess there's not much difference between the different OS's?)

.github/workflows/cd-preview.yml
```diff
@@ -43,7 +40,13 @@ jobs:
           VERSION_NR=$(cat packages/browser/package.json | npx json version)
           echo "::set-output name=version-nr::$VERSION_NR"
       - run: npm run build
-      - run: npm run publish-preview -- --dist-tag "$TAG_SLUG" --yes
+      - run: |
+          # Unfortunately GitHub Actions does not currently let us do something like
+          #     if: secrets.NPM_TOKEN != ''
+          # so simply skip the command if the env var is not set:
+          if [ -n $NPM_TOKEN ]; then
+          npm run publish-preview -- --dist-tag "$TAG_SLUG" --yes
```
Vinnl, line 48: Hmm, one problem I foresee here is that the `verify-imports-*` tasks that follow this will fail, because they're trying to install a package that wasn't published. We might need to fall back to just skipping the entire workflow if on a Dependabot branch and it wasn't triggered manually, if that's possible? Or even just if the trigger is Dependabot - it [looks like](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) you can do something like `if: github.actor != 'dependabot'`?

Vinnl (changes requested):
I think the jobs as specified now would still always fail for Dependabot :)

.github/workflows/ci.yml
NSeydoux: That's correct, I figured if the unit tests did run on all OS, the risks of the e2e tests failing only on Windows were relatively low, and at least low enough that it's not really worth working around that by writing a custom portable node script to replace the bash script :).

.github/workflows/cd-preview.yml
NSeydoux: Hmm yeah you're right, that would be an issue. But wouldn't it be possible to use `if: github.actor != 'dependabot` everywhere, instead of testing for the presence of the secrets ? That would remove the non-portability of bash as well.

Vinnl: Yes, that sounds like a good way forward to me.

Vinnl: Oh, and with the ability to trigger the tests manually still :)

Vinnl (approved):
Looks good, though I think it might also need a `workflow_dispatch` trigger to allow running it manually?

NSeydoux:
I was wondering if it would be required, since in the case a user pushes a commit or re-runs a workflow, they effectively become the github actor. I guess however in the case when updating an e2e-related library it's true that it would be convenient to trigger some workflows manually, I'll add it.

Vinnl:
Ah right, so you could re-run the workflows to trigger the other jobs after all. Couldn't hurt to add the manual trigger, but it's no blocker indeed :)

PR #1208, opened 2021-03-25
Skip e2e browser tests for dependabot

NSeydoux:
Skipping the event that triggers the deployment hangs the event waiting
fot the deployment instead of skipping it too. Now, the browser e2e test
are skipped for dependabot too.


Vinnl (approved):
I'm not sure if this will work - i.e. I'm not sure if this job's actor is Dependabot. But no harm in trying.

PR #1228, opened 2021-03-30
Updated dependencies to fix security issue

NSeydoux:
This commit is the result of running `npm audit fix` when NPM detected a security issue.


Vinnl (approved):
I see what you did here, but I'm strongly convinced that the approach I took in https://github.com/inrupt/solid-client-js/pull/953 is by far the superior approach :P

Issue #1242, opened 2021-04-02
Field [sessionId] for user [...] is not stored


NSeydoux:
This issue comes from discussion in https://github.com/inrupt/solid-client-authn-js/issues/423, where a bug unrelated to the initial issue has been discovered by @matthieu-fesselier.

See https://github.com/inrupt/solid-client-authn-js/issues/423#issuecomment-810895015 and the subsequent conversation for more details.

### Search terms you've used

- sessionId
- not stored
  
### Impacted package

Which packages do you think might be impacted by the bug ?

- [X] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

The login process does not complete. The user is redirected fine to the identity provider, but when redirected back to the client app, instead of being logged in, the user is still logged out, and a message "Field [sessionId] for user [<some guid>] is not stored" is displayed.  

### To Reproduce

See linked issue

Vinnl:
Also reported here: https://forum.solidproject.org/t/issues-with-logging-in-when-using-inrupt/4183

NSeydoux:
Due to the nature of the bug, it's hard to reproduce it consistently. @matthieu-fesselier, could you try to see if `npm install @inrupt/solid-client-authn-browser@fixstore-race-condition` resolves your issue ? I can't reproduce the bug on the repo you linked in #423 with this fix, but it's already happening only intermittently without the fix, so it's not like that proves much ^^.

matthieu-fesselier:
It does not, but I actually found the cause of the bug on our side. 

We have a component with a `auto-login` feature, which is supposed to redirect you at your IDP as soon as the page is loaded, if you are not authenticated. 

It seems that, when the bug appears, here is what is happening:
- page loaded
- `handleIncomingRedirect` called
- when resolved, `sessionInfo.isLoggedIn` is `false` -> we call the `login`
- after this, solid-authn-client-js performs the first redirect

When I deactivate the `auto-login` feature, everything works fine. 
Is there something that I can wait to be sure that the lib has completely handled the redirections?

NSeydoux:
[...]

PR #1243, opened 2021-04-02
Prevent race condition on login

NSeydoux:
There appears to be a race condition when logging in, that does not
manifest consistently. When it does, session information aren't stored
properly in local storage, resulting in a failure after redirection.

I suspect this causes #1242.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 #### browser
 
 - Building the browser package is now possible on Windows, thanks to more portable scripts.
+- Asynchronous calls that lead to a redirection when restoring a session are now
+  blocking, to prevent any race condition. 
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Building the browser package is now possible on Windows, thanks to more portable scripts.
 - Asynchronous calls that lead to a redirection when restoring a session are now
-  blocking, to prevent any race condition. 
+  blocking, to prevent the error associated to the message `Field [sessionId] for user [...] is not stored`
+  that gets thrown when the user is redirected back from the identity provider.
 
 The following sections document changes that have been released already:
 
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -545,10 +545,6 @@ describe("Session", () => {
   });
 
   describe("silent authentication", () => {
-    beforeEach(() => {
-      jest.useRealTimers();
-    });
-
     it("does nothing if no previous session is available", async () => {
       mockLocalStorage({});
       mockLocation("https://mock.current/location");
@@ -559,19 +555,15 @@ describe("Session", () => {
       const clientAuthentication = mockClientAuthentication({
         sessionInfoManager: mockSessionInfoManager(mockedStorage),
       });
+      const incomingRedirectPromise = Promise.resolve();
       clientAuthentication.handleIncomingRedirect = jest
         .fn()
-        .mockResolvedValue(undefined);
+        .mockReturnValue(incomingRedirectPromise);
       const mySession = new Session({ clientAuthentication });
       // eslint-disable-next-line no-void
       void mySession.handleIncomingRedirect({
         url: "https://some.redirect/url",
         restorePreviousSession: true,
-      });
-      // We know that handleincomingRedirect is never going to return, so just
-      // wait some time for the async calls to complete.
-      await new Promise((resolve) => {
-        setTimeout(resolve, 500);
       });
       expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
     });
@@ -597,12 +589,16 @@ describe("Session", () => {
       const clientAuthentication = mockClientAuthentication({
         sessionInfoManager: mockSessionInfoManager(mockedStorage),
       });
+      const incomingRedirectPromise = Promise.resolve();
       clientAuthentication.handleIncomingRedirect = jest
         .fn()
-        .mockResolvedValue(undefined);
+        .mockReturnValueOnce(incomingRedirectPromise);
+      const validateCurrentSessionPromise = Promise.resolve(
+        "https://some.issuer/"
+      );
       clientAuthentication.validateCurrentSession = jest
         .fn()
-        .mockResolvedValue("https://some.issuer/");
+        .mockReturnValue(validateCurrentSessionPromise);
 
       const mySession = new Session({ clientAuthentication });
       // eslint-disable-next-line no-void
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 #### browser
 
 - Building the browser package is now possible on Windows, thanks to more portable scripts.
+- Asynchronous calls that lead to a redirection when restoring a session are now
+  blocking, to prevent any race condition.
```
Vinnl, line 16: As a developer using the library, "race condition" doesn't tell me a lot. Maybe list the error that should go away? That way, if you encountered the error you'll know it's been fixed, and if you didn't, you'll know it doesn't apply to you :)

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
```diff
@@ -86,48 +86,45 @@ export default class AuthorizationCodeWithPkceOidcHandler
...
-          // We use the OAuth 'state' value (which should be crypto-random) as
-          // the key in our storage to store our actual SessionID. We do this
-          // 'cos we'll need to lookup our session information again when the
-          // browser is redirected back to us (i.e. the OAuth client
-          // application) from the Authorization Server.
-          // We don't want to use our session ID as the OAuth 'state' value, as
-          // that session ID can be any developer-specified value, and therefore
-          // may not be appropriate (since the OAuth 'state' value should really
-          // be an unguessable crypto-random value).
+    try {
```
Vinnl, line 89: To verify: this change just removes an unused import and rewrites this to async/await syntax - there's no behavioural change I need to look out for, right?

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
```diff
@@ -86,48 +86,45 @@ export default class AuthorizationCodeWithPkceOidcHandler
...
-          .catch((err: unknown) => {
-            // eslint-disable-next-line no-console
-            console.error(err);
-          })
-      );
-    });
+      redirector.redirect(signingRequest.url.toString(), {
+        handleRedirect: oidcLoginOptions.handleRedirect,
+      });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Vinnl, line 123: I think with `unknown` this might not be necessasry? (But be sure to check - might just be an ESLint warning, causing GitHub to add a million "check annotations" again...)

```suggestion
```

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
```diff
@@ -86,48 +86,45 @@ export default class AuthorizationCodeWithPkceOidcHandler
...
+    } catch (err: unknown) {
+      // eslint-disable-next-line no-console
+      console.error(err);
```
Vinnl, line 126: By the way (irrelevant to this PR, so feel free to continue anyway): why are we even logging this to the console? Isn't it better to just either let it bubble up (potentially up to the console), or not log it at all if it's not relevant to the developer?

packages/browser/__tests__/Session.spec.ts
```diff
@@ -559,10 +563,16 @@ describe("Session", () => {
...
+      // eslint-disable-next-line no-void
+      void mySession.handleIncomingRedirect({
         url: "https://some.redirect/url",
         restorePreviousSession: true,
       });
+      // We know that handleincomingRedirect is never going to return, so just
+      // wait some time for the async calls to complete.
+      await new Promise((resolve) => {
+        setTimeout(resolve, 500);
+      });
```
Vinnl, line 575: I don't understand what's going on here... Is the problem that it asynchronously writes to local storage, so we'll have to be able to await that in some way? The "some time" and `useRealTimers` here makes me nervous - wouldn't it just be enough to wait for the next tick in the event loop, i.e. `await Promise.resolve()`?

Vinnl (commented):
I need a bit more context about the test approach to be able to properly review that, sorry :/

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
NSeydoux: Absolutely, this used a mix of `await` and `.then`, which can sometimes result in confusing code, so at least now things are clearer. I initially suspected such a confusion caused by this could be causing the issue, and that wasn't the case, but I still think it's cleaner now :).

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
NSeydoux: That's a good question. Initially, it was meant to capture errors thrown by `oidc-client`, and help us debug them. Just swallowing the error didn't feel right, but I think also bubbling up the error was leaking from the underlying library. I guess we should still thrown an error wrapping up the underlying issue.

packages/browser/__tests__/Session.spec.ts
NSeydoux: The issue here is that ` void mySession.handleIncomingRedirect(...)` only resolves if no redirect is happening, which means that it does not resolve in "normal" conditions, so that the client code which looks like `await session.login()` is actually blocked until the actual redirect happens.

This piece of code is intended to work around this never returning promise. On one hand, we can't await it, but on the other some events in ` void mySession.handleIncomingRedirect(...)` are actually asynchronous, so not acknowledging that could introduce race conditions in the tests. I didn't realize using `await Promise.resolve()` could be a good fix for this, I'll try it out.

NSeydoux: Unfortunately, it isn't. My assumption is that `await Promise.resolve()` skips one tick on the event loop, but that  asynchronous operations happen for several ticks (I'm not very familiar with how the NodeJS event loop works, so if what I say doesn't make sense that's why ^^). I tested locally "sleeping" for 50ms instead of 500 ms, and it worked too, but my thinking was that putting a long wait is sure not to rely on the computer running the tests being actually quick, and even some old tired laptop can clear the bar easy ^^

packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
NSeydoux: Fixed in https://github.com/inrupt/solid-client-authn-js/pull/1243/commits/18980ee5072d3c6a207bd715908ae77c2f2860fc

CHANGELOG.md
NSeydoux: Rephrased in https://github.com/inrupt/solid-client-authn-js/pull/1243/commits/f2b2a864f260ba4e10b23a6367749f16e1213c60

packages/browser/__tests__/Session.spec.ts
Vinnl: Right, so it were multiple Promises that were being awaited, so that's indeed multiple ticks. I've added a commit that explicitly awaits those Promises, as a suggestion: 3d2cd3df74a172046a3f7ff6781920edc1778a0c

NSeydoux: I see, yeah that looks much better than waiting for an arbitrary duration ! Thanks.

Vinnl (approved):
I wonder if it would be possible to verify that indeed the Promise does not resolve when silent authentication applies :thinking: Maybe something with a `finally()` call? Not terribly important I don't think, so good to go as far as I'm concerned.

NSeydoux:
That's a very good question, but I couldn't figure out if doing so was possible. I'll keep a note for this.

PR #1274, opened 2021-04-14
ID token signing algorithm negotiation

NSeydoux:
This enforces negotiation of the algorithm used to sign ID tokens. In its discovery document (`/.well-known/openid-configuration`), the OIDC provider must specify a list of supported signature algorithms with the `id_token_signing_alg_values_supported` property. When registering, a client may specify a signing algorithm that they want used to sign the ID Token the OIDC provider will send. 

The default behaviour is to use `RS256`, but this default behaviour is only a suggestion, not a requirement, which means we cannot rely on the default. Case in point, the ESS ID broker defaults to `ES256`, which breaks the client library if not handled properly. Now, we specify a list of preferred signing algorithms (currently `["ES256", "RS256"]`), and the library selects the most preferred algorithm in the list of algorithms supported by the OIDC issuer.  

Note: the ID token verification is only updated in the `node` library, because the `browser` library doesn't require the signing algorithm to be specified to verify. 

# New feature description

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Bugfixes
+
+#### node and oidc
+
+- The OIDC issuer profile is used to negociate the preferred signature algorithm
+for ID tokens.
+
+#### node
+
+- During client registration, the client explicitly registers both the 'refresh_token'
+and the 'authorization_code' grants, instead of only relying on scopes to get 
+refresh tokens.
+
 The following sections document changes that have been released already:
 
 ## 1.7.3 - 2021-04-09
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,10 @@ for ID tokens.
 
 #### node
 
-- During client registration, the client explicitly registers both the 'refresh_token'
-and the 'authorization_code' grants, instead of only relying on scopes to get 
-refresh tokens.
+- During client registration, the client explicitly specifies both the 'refresh_token'
+and the 'authorization_code' grants as part of its profile, instead of only relying
+on scopes to get refresh tokens. Depending on the Identity Provider, the former 
+behaviour could result in not getting refresh tokens.
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 #### node and oidc
 
-- The OIDC issuer profile is used to negociate the preferred signature algorithm
+- The OIDC issuer profile is used to negotiate the preferred signature algorithm
 for ID tokens.
 
 #### node
--- a/packages/core/src/login/oidc/IClientRegistrar.test.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.test.ts
@@ -121,7 +121,7 @@ describe("determineSigningAlg", () => {
     expect(determineSigningAlg(["RS256"], ["RS256"])).toBe("RS256");
   });
 
-  it("returns undefined if there are no matches", () => {
+  it("returns null if there are no matches", () => {
     expect(determineSigningAlg(["RS256"], ["ES256"])).toBeNull();
   });
 });
--- a/packages/core/src/login/oidc/IClientRegistrar.test.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.test.ts
@@ -123,5 +123,6 @@ describe("determineSigningAlg", () => {
 
   it("returns null if there are no matches", () => {
     expect(determineSigningAlg(["RS256"], ["ES256"])).toBeNull();
+    expect(determineSigningAlg(["RS256"], [])).toBeNull();
   });
 });
--- a/packages/core/src/login/oidc/IClientRegistrar.test.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.test.ts
@@ -21,7 +21,7 @@
 
 import { it, describe, expect } from "@jest/globals";
 import { IIssuerConfig, ILoginOptions, IStorageUtility } from "../..";
-import { handleRegistration } from "./IClientRegistrar";
+import { determineSigningAlg, handleRegistration } from "./IClientRegistrar";
 
 describe("handleRegistration", () => {
   it("should perform DCR if a client WebID is provided, but the target IdP does not support Solid-OIDC", async () => {
@@ -107,5 +107,21 @@ describe("handleRegistration", () => {
     );
     expect(clientRegistrar.getClient).not.toHaveBeenCalled();
     expect(storageUtility.setForUser).toHaveBeenCalled();
+  });
+});
+
+describe("determineSigningAlg", () => {
+  it("returns the preferred algorithm of the supported list", () => {
+    expect(
+      determineSigningAlg(["ES256", "HS256", "RS256"], ["ES256", "RS256"])
+    ).toBe("ES256");
+    expect(determineSigningAlg(["ES256", "HS256", "RS256"], ["RS256"])).toBe(
+      "RS256"
+    );
+    expect(determineSigningAlg(["RS256"], ["RS256"])).toBe("RS256");
+  });
+
+  it("returns undefined if there are no matches", () => {
+    expect(determineSigningAlg(["RS256"], ["ES256"])).toBeNull();
   });
 });
--- a/packages/core/src/login/oidc/IClientRegistrar.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.ts
@@ -58,6 +58,22 @@ function isValidUrl(url: string): boolean {
   }
 }
 
+export function determineSigningAlg(
+  supported: string[],
+  preferred: string[]
+): string | null {
+  let preferredSupportedAlg = null;
+  preferred.some((signingAlg) => {
+    if (supported.includes(signingAlg)) {
+      preferredSupportedAlg = signingAlg;
+      // returnning true breaks the .some() iteration.
+      return true;
+    }
+    return false;
+  });
+  return preferredSupportedAlg;
+}
+
 export async function handleRegistration(
   options: ILoginOptions,
   issuerConfig: IIssuerConfig,
--- a/packages/core/src/login/oidc/IClientRegistrar.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.ts
@@ -65,7 +65,7 @@ export function determineSigningAlg(
   return (
     preferred.find((signingAlg) => {
       return supported.includes(signingAlg);
-    }) || null
+    }) ?? null
   );
 }
 
--- a/packages/core/src/login/oidc/IClientRegistrar.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.ts
@@ -62,16 +62,11 @@ export function determineSigningAlg(
   supported: string[],
   preferred: string[]
 ): string | null {
-  let preferredSupportedAlg = null;
-  preferred.some((signingAlg) => {
-    if (supported.includes(signingAlg)) {
-      preferredSupportedAlg = signingAlg;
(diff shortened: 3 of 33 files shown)


packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -58,6 +58,22 @@ function isValidUrl(url: string): boolean {
...
+  let preferredSupportedAlg = null;
+  preferred.some((signingAlg) => {
+    if (supported.includes(signingAlg)) {
+      preferredSupportedAlg = signingAlg;
+      // returnning true breaks the .some() iteration.
+      return true;
+    }
+    return false;
+  });
+  return preferredSupportedAlg;
```
ajacksified, line 74: This could be written more simply using `.find`:

```
export function determineSigningAlg(
  supported: string[],
  preferred: string[]
): string | null {
  return preferred.find((signingAlg) => {
    return supported.includes(signingAlg);
  }) || null;
}
```

packages/oidc/src/common/negociation.ts
```diff
@@ -0,0 +1,42 @@
...
+  let preferredSupportedAlg = null;
+  preferred.some((signingAlg) => {
+    if (supported.includes(signingAlg)) {
+      preferredSupportedAlg = signingAlg;
+      // returnning true breaks the .some() iteration.
+      return true;
+    }
+    return false;
+  });
+  return preferredSupportedAlg;
```
ajacksified, line 39: Same here - logic can be simplified with `find`

ajacksified (changes requested):

packages/core/src/login/oidc/IClientRegistrar.ts
NSeydoux: d7febeaa8dc187ada7ce826058fd42a4c1dbe7b5

packages/oidc/src/common/negociation.ts
NSeydoux: 498eeaad6f60d848cf6118e3103b531e89f1cc25

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Bugfixes
+
+#### node and oidc
+
+- The OIDC issuer profile is used to negociate the preferred signature algorithm
```
pmcb55, line 14:
```suggestion
- The OIDC issuer profile is used to negotiate the preferred signature algorithm
```

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+for ID tokens.
+
+#### node
+
+- During client registration, the client explicitly registers both the 'refresh_token'
```
Vinnl, line 19: Minor, but as a developer who just wants to be able to connect to people's Pods, I have no idea how this affects me :P How do I know if this bug affected me, and if it did, how?

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -58,6 +58,17 @@ function isValidUrl(url: string): boolean {
...
 }
 
+export function determineSigningAlg(
+  supported: string[],
+  preferred: string[]
+): string | null {
+  return (
+    preferred.find((signingAlg) => {
+      return supported.includes(signingAlg);
+    }) || null
```
Vinnl, line 68: Theoretically shouldn't cause any problems here, but generally good practice to avoid casting if unnecessary.

```suggestion
    }) ?? null
```

packages/core/src/login/oidc/IClientRegistrar.test.ts
```diff
@@ -109,3 +109,19 @@ describe("handleRegistration", () => {
...
+    expect(
+      determineSigningAlg(["ES256", "HS256", "RS256"], ["ES256", "RS256"])
+    ).toBe("ES256");
+    expect(determineSigningAlg(["ES256", "HS256", "RS256"], ["RS256"])).toBe(
+      "RS256"
+    );
+    expect(determineSigningAlg(["RS256"], ["RS256"])).toBe("RS256");
+  });
+
+  it("returns undefined if there are no matches", () => {
```
Vinnl, line 124:
```suggestion
  it("returns null if there are no matches", () => {
```

packages/node/src/login/oidc/ClientRegistrar.ts
```diff
@@ -32,10 +32,40 @@ import {
...
+  PREFERRED_SIGNING_ALG,
 } from "@inrupt/solid-client-authn-core";
 import { Client, Issuer } from "openid-client";
 import { configToIssuerMetadata } from "./IssuerConfigFetcher";
 
+export function negotiateClientSigningAlg(
+  issuerConfig: IIssuerConfig,
+  clientPreference: string[]
+): string {
+  if (issuerConfig.idTokenSigningAlgValuesSupported === undefined) {
```
Vinnl, line 45: I'm having a bit trouble navigating the codebase from the PR interview, but have you tested this using an `issuerConfig` in which this property is not declared, rather than undefined? i.e.

```
issuerConfig = {}
```

rather than

```
issuerConfig = { idTokenSigningAlgValuesSupported: undefined }
```

Off the top of my head I'm not sure whether the same holds for object properties as for variables, but if it does, the former might be problematic (which can be solved by doing a check for `typeof ... === "undefined"`).

Vinnl: Actually, might be good to use  `Array.isArray()` anyway, just to also guard against some more defined-but-invalid values.

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -207,6 +213,28 @@ describe("TokenRefresher", () => {
...
+      }),
+    });
+
+    const refreshedTokens = await refresher.refresh(
+      "mySession",
+      "some refresh token",
+      mockJwk()
+    );
+
+    expect(refreshedTokens.access_token).toEqual(mockDpopTokens().access_token);
```
Vinnl, line 235: Classic :)

```suggestion
    expect(refreshedTokens.access_token).toBe(mockDpopTokens().access_token);
```

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -80,9 +82,16 @@ export default class TokenRefresher implements ITokenRefresher {
       { sessionId },
       oidcContext.issuerConfig
     );
+    if (clientInfo.idTokenSignedResponseAlg === undefined) {
```
Vinnl, line 85: Likewise here as the other comment about comparing with `undefined`.

packages/oidc/src/common/negotiation.ts
```diff
@@ -0,0 +1,37 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+// FIXME: this is copy-pasted from @inrupt/solid-client-authn-core (as is all the rest of this directory).
+// Initially, it was aimed that @inrupt/oidc-client-ext would not depend on the core package, and only
+// on the underlying OIDC library, but it is beginning to be impractical. This should be resolved in a
+// future PR.
```
Vinnl, line 25: Maybe also mention here what the resolution should be in that future PR, i.e. should it be consolidated into -core, or extracted into the individual packages?

Vinnl (approved):

packages/core/src/login/oidc/IClientRegistrar.test.ts
pmcb55, line 124: What if no preference is provided at all...? Needs an explicit test, just to communicate the behaviour (is it 'null' or first preferred entry...?)

packages/oidc/src/common/negotiation.ts
NSeydoux: Well I wasn't sure what the resolution should be really :). My initial thought was to have `oidc-client-ext` simply depend on `solid-client-authn-core`, but I think there was a reason not to do that earlier, so I preferred not to say anything in the comment, and consolidate this properly later.

CHANGELOG.md
NSeydoux: 725b5c335a8a16de7ecfaa20c0235a30a85398b5

packages/core/src/login/oidc/IClientRegistrar.ts
NSeydoux: 77abf75ad79c767e18d6515ff269cafdbb260c0e

packages/node/src/login/oidc/ClientRegistrar.ts
NSeydoux: a90e36b19f5ac153d425356eb01a746b5f52f1a5

packages/node/src/login/oidc/refresh/TokenRefresher.ts
NSeydoux: a90e36b19f5ac153d425356eb01a746b5f52f1a5

packages/core/src/login/oidc/IClientRegistrar.test.ts
NSeydoux: 0cb556ce03d197a6ebb51075ed4de50e0a57d8d1

packages/oidc/src/common/negotiation.ts
Vinnl: Ah haha, yeah that makes it hard to note down the resolution :P

vanhumbeecka:
I just bumped onto an issue that is probably related to this ticket:

While using `@inrupt/solid-client-authn-node` for server-side authentication, everything works fine when I use the `inrupt.net` pods (probably using RSA?). But fails when using the `broker.pods.inrupt.com` pods (probably using ES?).

My stacktrace shows:
```
11:43:11.376Z ERROR survey-aggregator: Uncaught Error (body={}, host=localhost, method=GET, params={}, path=/handle-redirect)
    RPError: unexpected JWT alg received, expected RS256, got: ES256
        at Client.validateJWT (/Users/andries/projects/consolidate/survey-aggregator/node_modules/openid-client/lib/client.js:814:13)
        at Client.validateIdToken (/Users/andries/projects/consolidate/survey-aggregator/node_modules/openid-client/lib/client.js:693:60)
        at Client.callback (/Users/andries/projects/consolidate/survey-aggregator/node_modules/openid-client/lib/client.js:469:18)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)
        at async AuthCodeRedirectHandler.handle (/Users/andries/projects/consolidate/survey-aggregator/node_modules/@inrupt/solid-client-authn-node/dist/login/oidc/redirectHandler/AuthCodeRedirectHandler.js:79:24)
        at async ClientAuthentication.handleIncomingRedirect (/Users/andries/projects/consolidate/survey-aggregator/node_modules/@inrupt/solid-client-authn-node/dist/ClientAuthentication.js:71:34)
        at async Session.handleIncomingRedirect (/Users/andries/projects/consolidate/survey-aggregator/node_modules/@inrupt/solid-client-authn-node/dist/Session.js:45:35)
        at async AuthController.handleRedirect (/Users/andries/projects/consolidate/survey-aggregator/dist/modules/auth/AuthController.js:50:13)
```

NSeydoux:
Hi @vanhumbeecka, thanks for reporting this ! It looks like this is absolutely related to this issue. The fix should have been released with `v1.7.4` this morning, which version of the library are you depending on ?

vanhumbeecka:
Hi @NSeydoux , I'm using `v1.7.2`. 
I just upgraded to `v1.7.4` to confirm, and it works now :)

PR #1284, opened 2021-04-19
Add refresh token handling hooks

NSeydoux:
In order to allow for some use cases where using the storage isn't the
most practical solution, this makes it possible to run some custom code
when getting an access token. Two paths are considered:
- when initially logging in, when the refresh token is returned as a
response from the token endpoint. In this case,
`session.handleincomingRedirect` accepts a new parameter,
`handleRefreshToken`, which allows the caller to run custom code at this
point. An example usage for this would be to replace the current
behaviour of `generate-oidc-token`, which looks into the storage to get
the refresh token, and which could instead use this handler.
- when using a refresh token to get a new access token. In this case,
the refresh token may be rotated by the identity provider. The
`session.login` function takes for this use case a new parameter,
`handleRefreshTokenRotation`, which allows the caller to execute custom
code to manage the new refresh token, e.g. in the case the used storage
isn't adapted to handling the token. This would be useful to us in the
case of end-to-end tests, where we could share the new refresh token to
other tests via an environment variable. This could also be useful to NodeJS scripts, in the case they don't implement a persistent storage.

# Checklist

### DIFF
--- a/packages/core/src/login/ILoginOptions.ts
+++ b/packages/core/src/login/ILoginOptions.ts
@@ -51,5 +51,5 @@ export default interface ILoginOptions extends ILoginInputOptions {
    * This callback will be called if, during the refresh token flow, the refresh token is rotated by the Solid Identity
    * Provider. In this case, the provided function is called with the new token as a parameter.
    */
-  handleRefreshTokenRotation?: (token: string) => unknown;
+  onNewRefreshToken?: (newToken: string) => unknown;
 }
--- a/packages/core/src/login/ILoginOptions.ts
+++ b/packages/core/src/login/ILoginOptions.ts
@@ -47,4 +47,9 @@ export default interface ILoginOptions extends ILoginInputOptions {
   prompt?: string;
   // Force the token type to be required (i.e. no longer optional).
   tokenType: "DPoP" | "Bearer";
+  /**
+   * This callback will be called if, during the refresh token flow, the refresh token is rotated by the Solid Identity
+   * Provider. In this case, the provided function is called with the new token as a parameter.
+   */
+  handleRefreshTokenRotation?: (token: string) => unknown;
 }
--- a/packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
+++ b/packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
@@ -34,7 +34,7 @@ export type RedirectResult = ISessionInfo & { fetch: typeof fetch };
  * @hidden
  */
 type IRedirectHandler = IHandleable<
-  [url: string, handleRefreshToken?: (token: string) => unknown],
+  [url: string, onNewRefreshToken?: (newToken: string) => unknown],
   RedirectResult
 >;
 export default IRedirectHandler;
--- a/packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
+++ b/packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
@@ -33,5 +33,8 @@ export type RedirectResult = ISessionInfo & { fetch: typeof fetch };
 /**
  * @hidden
  */
-type IRedirectHandler = IHandleable<[string], RedirectResult>;
+type IRedirectHandler = IHandleable<
+  [url: string, handleRefreshToken?: (token: string) => unknown],
+  RedirectResult
+>;
 export default IRedirectHandler;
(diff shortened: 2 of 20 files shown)


packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
```diff
@@ -33,5 +33,8 @@ export type RedirectResult = ISessionInfo & { fetch: typeof fetch };
 /**
  * @hidden
  */
-type IRedirectHandler = IHandleable<[string], RedirectResult>;
+type IRedirectHandler = IHandleable<
+  [url: string, handleRefreshToken?: (token: string) => unknown],
```
acoburn, line 37: how does the `handleRefreshToken?` method differ from `handleRefreshTokenRotation?`

NSeydoux: In essence, they don't, and I'd have preferred if using only one would be sufficient. However, they are used in two different contexts: since they aren't tied to the session, they need to be injected in the two entry points when they may be used, that is either when the refresh token is received from the token endpoint initially (hence when calling `handleIncomingRedirect`), or whenever a refresh token is used to get a new access token (hence when calling `login`).

However, writing this, I realize the callback could be specified as part of the `Session` constructor, and then injected behind the scenes by the `Session` object to the calls to the `ClientAuthorization` methods. I'll make a PR in that sense, it will make a cleaner API.

packages/core/src/login/ILoginOptions.ts
```diff
@@ -47,4 +47,9 @@ export default interface ILoginOptions extends ILoginInputOptions {
   prompt?: string;
   // Force the token type to be required (i.e. no longer optional).
   tokenType: "DPoP" | "Bearer";
+  /**
+   * This callback will be called if, during the refresh token flow, the refresh token is rotated by the Solid Identity
+   * Provider. In this case, the provided function is called with the new token as a parameter.
+   */
+  handleRefreshTokenRotation?: (token: string) => unknown;
```
Vinnl, line 54: This callback doesn't really do any "handling" of the rotation, right? And there's nothing we're doing with the return value?

```suggestion
  onNewRefreshToken?: (newToken: string) => void;
```

packages/node/src/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -137,6 +137,25 @@ describe("buildBearerFetch", () => {
     );
   });
 
+  it("returns a fetch that calls the refresh token handler if appropriate", async () => {
+    const fetch = jest.requireMock("cross-fetch");
+    fetch.mockResolvedValue({ status: 401 });
+    const tokenSet = mockDefaultTokenSet();
+    tokenSet.refresh_token = "some refreshed refresh token";
```
Vinnl, line 144: Just to make this slightly less confusing (also in other tests):

```suggestion
    tokenSet.refresh_token = "some rotated refresh token";
```

packages/node/src/authenticatedFetch/fetchFactory.ts
```diff
@@ -81,8 +82,10 @@ export function buildBearerFetch(
       currentAccessToken = tokenSet.access_token;
       if (tokenSet.refresh_token) {
         // If the refresh token is rotated, update it in the closure.
-        // TODO: Plug this into external storage, if one is provided.
         currentRefreshOptions.refreshToken = tokenSet.refresh_token;
+        if (currentRefreshOptions.handleRefreshToken !== undefined) {
```
Vinnl, line 86: A positive check on the type is safer (also down below):

```suggestion
        if (typeof currentRefreshOptions.handleRefreshToken === "function") {
```

Vinnl (approved):
Implementation-wise this looks good :+1: , but `handle` seems like a misnomer; it's more like an event listener. That also meshes better with the approach of defining it once in the session constructor, and having it called regardless of whether the refresh token is the first one, or a rotated one. And it also makes it clearer that we're not really expecting any return value from the handler.

packages/node/src/authenticatedFetch/fetchFactory.ts
NSeydoux: 99010293006619e1288c0791b81a55a5c7ace116

packages/node/src/authenticatedFetch/fetchFactory.spec.ts
NSeydoux: That's clearer, but it's less funny to say ^^ 838255e114924476169ffe933a1cae2e20f908cf

ajacksified (approved):

packages/node/src/authenticatedFetch/fetchFactory.spec.ts
Vinnl: Can't argue with that :P

packages/core/src/login/ILoginOptions.ts
Vinnl: I see you left the return type to be `unknown`, implying that SCAN is going to do something with that return type. Is there a reason for not making it `void`, or did you just overlook that? :)

PR #1290, opened 2021-04-21
Disable ESS-specific session workaround by default

Vinnl:
The calls to /session that resulted in 404s were confusing to
developers, and a less brittle workaround is now available with
`restorePreviousSession`. ESS has also disabled the option by
default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Bugfixes
+
+- The workaround to maintain sessions in 1.6.1 has been disabled by default;
+  this means you should see no more (failed) calls to `/session` in your network
+  console. If you want to make sure a session is preserved across page reloads,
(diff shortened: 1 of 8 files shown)


NSeydoux (changes requested):
There's one thing that would be good to add in this PR as part of disabling this behaviour by default: currently, the `fetch` function for `ClientAuthentication` is initialized to a locally defined `fetchWithCookies`, which is a regular fetch with `credentials: "include"`. Including credentials is only required in the case when we're relying on a cookie for authenticated fetch, and it's causing CORS issues in many cases where it's not providing any value. Ideally, in `Session.ts`, in the code dedicated to the workaround, if the ESS cookie is used for authentication then the session should override its `clientAuthentication`'s `fetch` with `fetchWithCookies`.

Vinnl:
Ah, gotcha - I'd assumed that was automatically avoided. Will change that tomorrow, thanks.

Vinnl:
OK, done in aa98bba39780cf7d35821e434a42054b7ecd3139.

NSeydoux (approved):

PR #1329, opened 2021-05-04
Implement support for client credentials

NSeydoux:
The client credential handle only considers requests where both the
client ID and secret are defined. Only confidential clients are considered
for client credentials, public client (registered through DCR) cannot go 
through this grant. 

if `session.login` is called with the appropriate parameters (that is, a client 
ID an secret), the client credential flow is triggered and directly returns a valid
session, i.e. no redirection is needed.

in order to make it easier to test, and to implement something that is long
overdue, this also adds a new fetch factory that applies to both bearer and
DPoP tokens. 

# Checklist

### DIFF
--- a/packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
+++ b/packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
@@ -30,14 +30,28 @@
 import {
   IOidcHandler,
   IOidcOptions,
-  NotImplementedError,
   LoginResult,
+  IStorageUtility,
+  ISessionInfo,
 } from "@inrupt/solid-client-authn-core";
+import { JWK } from "jose";
+import { Issuer } from "openid-client";
+import { inject, injectable } from "tsyringe";
+import { buildDpopFetch } from "../../../authenticatedFetch/fetchFactory";
+import { configToIssuerMetadata } from "../IssuerConfigFetcher";
+import { getWebidFromTokenPayload } from "../redirectHandler/AuthCodeRedirectHandler";
+import { ITokenRefresher } from "../refresh/TokenRefresher";
 
 /**
  * @hidden
  */
+@injectable()
 export default class ClientCredentialsOidcHandler implements IOidcHandler {
+  constructor(
+    @inject("node:tokenRefresher") private tokenRefresher: ITokenRefresher,
+    @inject("node:storageUtility") private storageUtility: IStorageUtility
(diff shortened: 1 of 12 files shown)


ajacksified (changes requested):
There is an extremely narrow scope of when client credentials are useful. Should we even support this? I was under the impression that we would not spend any time at all on this and instead look for a long-term solution for handling tests and non-web-client application authentication.

This seems to be an antipattern in Solid applications.

packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
```diff
@@ -30,19 +30,102 @@
...
+          oidcLoginOptions.issuer
+        }]: ${JSON.stringify(tokens)} is missing 'access_token'.`
+      );
+    }
+
+    if (tokens.id_token === undefined) {
+      throw new Error(
+        `Invalid response from Solid Identity Provider [${
+          oidcLoginOptions.issuer
+        }]: ${JSON.stringify(tokens)} is missing 'id_token'`
```
Vinnl, line 105: I'm channelling my inner Pat here:

```suggestion
        }]: [${JSON.stringify(tokens)}] is missing 'access_token'.`
      );
    }

    if (tokens.id_token === undefined) {
      throw new Error(
        `Invalid response from Solid Identity Provider [${
          oidcLoginOptions.issuer
        }]: [${JSON.stringify(tokens)}] is missing 'id_token'`
```

Vinnl (approved):

packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
NSeydoux: Haha I left the square brackets out on purpose, because `JSON.stringify(tokens)` will produce a JSON string enclosed in curly brackets, and I think adding square brackets around that will make it look like an array, which it is not :)

Vinnl: Fair enough :)

edwardsph (commented):
Just a small comment on: "As such, it should only be used in specific cases."
I wonder if we need to be clearer about "specific cases" - anyone who wants to use this method could just say my case is a "specific case", in fact - aren't all cases "specific cases"? Can we add some more scope to this comment to identify the relevant class of situations where this may be a reasonable choice.

PR #1338, opened 2021-05-06
Add architecture documentation

NSeydoux:
In order to make it easier to onboard or navigate the codebase, an
architecture document is introduced, loosly based on the recommandations
from https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html.

This document aims at:
- Pointing to helpful resources regarding OAuth/OIDC. It does not aim at
being itself an OIDC resource.
- Documenting design patterns/architectural specificities of the library, in the current case handler pattern and dependency injection.
- Providing a code map giving a high-level idea of what happens where in
the code.

Note that the SVG diagrams are automatically generated from the `.pluml` files using PlantUML. Versioning generated files isn't great, but they should change rarely enough that it isn't a problem. Also, GH renders the SVG diagrams in the 'Rich diff' mode for MD files, which should make this more convenient to review. It sometimes doesnot render them initially when I enable the rich diff, but refreshing the page then fixes it.


ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+In the context of this library, a request is an API to execute some OIDC-related operation. Handlers will determine whether they can handle the request based on the options specified by the user when making the call.
+
+#### Login
+
+Logging in is the operation that goes from the Client to the Issuer. It may result in a redirection of the user to the Issuer, but that is not always the case (in the `-node` module in particular). Handlers for the login operation are located in `packages/*/src/login/oidc/oidcHandlers/*Handler.ts`.
```
Vinnl, line 60: "goes from" is a bit unclear to me here. Does this mean "the Client has to get the Resource Owner to visit a webpage of the Issuer, e.g. via redirection (in a browser-based app) or by having them click a link or scan a QR code)"?

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Incoming redirect
+
+OIDC is based on redirection of the user's browser between the Client and the Issuer. The incoming redirect happens after the Resource Owner has authenticated to the Issuer, and the Issuer redirects them to the Client's `redirect_uri` with some relevant information in the query params. Handlers for the incoming redirect are located in `packages/*/src/login/oidc/redirectHandler/*Handler.ts`.
```
Vinnl, line 64: I'm going to suggest a rewording here that keeps the events in chronological order. I hope that what I'm saying is correct :crossed_fingers: :

```suggestion
At the Issuer webpage, the Resource Owner authenticates (e.g. by entering a username and a password), after which the Issuer sends them to a webpage under the Client app's control (its `redirect_uri`), to which it appends some query parameters that it can use to complete the login flow. This is done when the developer calls `handleIncomingRedirect`, and the Handlers for the incoming redirect are located in `packages/*/src/login/oidc/redirectHandler/*Handler.ts`.
```

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+});
+container.register<IOidcHandler>("browser:oidcHandlers", {
+  useClass: AuthorizationCodeWithPkceOidcHandler,
+});
+container.register<IOidcHandler>("browser:oidcHandlers", {
+  useClass: ClientCredentialsOidcHandler,
+});
+```
+
+Here, `AggregateOidcHandler` is the handler aggregator (as defined in the Handler Pattern section), and `RefreshTokenOidcHandler`, `AuthorizationCodeWithPkceOidcHandler` and `ClientCredentialsOidcHandler` are its underlying handlers. When receiving a request, `AggregateOidcHandler` will first call to its instance of `RefreshTokenOidcHandler` to check if it can handle it. If so, the instance of `RefreshTokenOidcHandler` will handle the request, and the instances of `AuthorizationCodeWithPkceOidcHandler` or `ClientCredentialsOidcHandler` will not be called. This means that it is important to declare the dependencies from the most specialized to the most generic, because if a fallback handler that can handle all requests is declared first, the other more specialized handlers will not be called.
```
Vinnl, line 89: Wait, if the order in which the Handlers are added to `container` determines which gets called, then why does `AggregateOidcHandler` need to get called first which I'd _also_ assume to have a list of Handlers to determine which gets called? (In other words, what does `AggregateOidcHandler` do here that `container` doesn't?)

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+can be tested as follows: 
+```
+const refreshTokenOidcHandler = new RefreshTokenOidcHandler(
+  someMockedTokenRefresher,
+  someMockedStorageUtility
+);
+```
+
+## Mapping OIDC flows to the code
+
```
Vinnl, line 116: This is very useful - what might make it even more so is to add in which use cases which flow is used?

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+### Auth code flow
+
+![Module dependencies](./documentation/diagrams/auth_code_flow.svg)
+
+- Discovery: `packages/*/src/login/oidc/IssuerConfigFetcher.ts`
+- Registration: `packages/*/src/login/oidc/ClientRegistrar.ts`
+- OIDC handler: `packages/*/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts`
+- Handle incoming redirect: `packages/*/src/login/oidc/redirectHandler/*Handler.ts`
```
Vinnl, line 124: Might be good to mention here why you can't list the specific Handler here. (Because I don't know :P)

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Refresh flow
+
+![Module dependencies](./documentation/diagrams/refresh_flow.svg)
+
+- OIDC handler: `packages/*/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts`
+
+Note that in this case, no redirection happens: the Access Token is received directly
```
Vinnl, line 132:
```suggestion
Note that in this case, no redirection happens (i.e. there's only a Back channel): the Access Token is received directly
```

Vinnl (commented):
This is great! I added some suggestions, which I hope are correct :P (If not, that's a learning opportunity for me I suppose.)

I haven't approved it because the others should [read it](https://github.com/inrupt/solid-client-authn-js/blob/chore/architecture/ARCHITECTURE.md) too :)

ARCHITECTURE.md
NSeydoux: That's certainly a possibility, however I'd like to have a discussion first: wouldn't that get rather deep into explaining OIDC/OAuth, which is information available in greater details elsewhere, and stray away from presenting how the codebase relates to OIDC/OAuth ?

Vinnl: Hmm, I was thinking of a single comment relating it to Solid (e.g. "this is used by Solid apps running in the browser to login" and "this is used by Solid apps that run on the server or natively to maintain a session"), but perhaps I'm misunderstanding how specific to certain use cases these flows are and I should just read up on OIDC again :P

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+- [How To Securely Implement Authentication in Single Page Applications](https://betterprogramming.pub/how-to-securely-implement-authentication-in-single-page-applications-670534da746f), by Dennis Stötzel
+
+### Solid-OIDC
+
+Solid extends the OIDC protocol in order to make it fit into a decentralized ecosystem. The [Solid-OIDC specification](https://solid.github.io/authentication-panel/solid-oidc/) is the reference document to learn about this extension.
+
+### A short glossary
+
+Here is a list of terms having a specific meaning in the context of OIDC:
+- **Resource Server**: The server hosting private resources. In our case, a Solid server. A Resource Server receives requests authenticated with an Access Token
```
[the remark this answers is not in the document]
NSeydoux: I implemented part of this change in ce3d056ab989aeff0b592f0ad12f3f558f831b9f, but I'm not sure about the first proposed sentence. Technically, the notion of Pod is quite vague, and it is currently not part of the spec (as far as I know). It is introduced via the notion of Pod Root and Pod Owner, but one Resource Server will serve many Pods, and most Pods will contain resources that may be accessed by users that are not the Pod Owner. Therefore, I'm afraid that associating the notion of a Resource Server with a WebID is a bit restrictive, even for the sake of explanation.

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+### The Handler pattern
+
+Important components of this library are based on the Handler design pattern: given data contained in a request, a set of classes implementing a similar API will declare whether or not they may handle said request. Handlers declare two functions: `canHandle(request)` and `handle(request)`. `canHandle` simply returns a boolean indicating the ability of the handler to handle the request, and `handle` actually processes it. All of the handlers for a given type of requests are tracked by a handler aggregator, which will have the first handle for which `canHandle` returns `true` process the request. The handler aggregator has the same API as the handlers it aggregates, and brokers the request to the underlying handlers. More on that in the Dependency Injection section.
+
+In the context of this library, a request is an API to execute some OIDC-related operation. Handlers will determine whether they can handle the request based on the options specified by the user when making the call.
```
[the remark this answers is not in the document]
NSeydoux: That's because it was meant to say "a request is an API call" :p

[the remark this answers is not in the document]
Vinnl: Ah well, that _is_ a lot clearer :P

ARCHITECTURE.md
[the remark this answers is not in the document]
Vinnl: That's fair enough :+1:

ARCHITECTURE.md
NSeydoux: Note that the label for the containers of the aggregator and the underlying handlers differ (respectively `browser:oidcHandler` and `browser:oidcHandlers`, with an 's'). In this case, the order in which the dependencies are declared is irrelevant, because they do not relate to the same container. The Aggregator implements the class from `packages/core/src/util/handlerPattern/AggregateHandler.ts`, and uses the `@injectAll` annotation to receive all the handlers registered to a given container. If this comment clarifies things, I'll add it to the documentation.

Vinnl: Aha! Yes, that's definitely worth pointing out - at least the `s` part.

ARCHITECTURE.md
NSeydoux: 7b69d5233e25714ca2f18500cf2cdded6f198ea2 is a bit more specific about the handlers that may get called, but really it depends on their `canHandle` conditions. Typically, the list will grow if we implement error handling as I suggest in the dedicated ticket, by adding a dedicated handler that `canHandle` incoming redirects where the redirect IRI contains an `error` query param.

ARCHITECTURE.md
NSeydoux: Well it's more that I'm afraid I would give a truncated view of which flows may be used in which cases, while really I don't expect things to be Solid-specific here, but to hold in any OIDC context. However, if you think it may be useful, I could details which flows are explicitly NOT supported, e.g. the refresh flow in the browser, or explain why using the client credential flow is not something that works well with Solid.

NSeydoux: Actually I'll do that, because figuring out what things are absent is always trickier than finding the things that are there

Vinnl: Yeah absolutely, sounds like a good idea.

ARCHITECTURE.md
Vinnl: Ah, that sounds like a good approach to error handling. How about just mentioning something like "the specific Handler depends on which Handler's `canHandle()` method first returns `true`"?

ARCHITECTURE.md
NSeydoux: 78e9ddd6ab6e555a1f559a7807465b4dd7e12cb1

ARCHITECTURE.md
NSeydoux: 78e9ddd6ab6e555a1f559a7807465b4dd7e12cb1

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+will handle the request, and the instances of `AuthorizationCodeWithPkceOidcHandler`
+or `ClientCredentialsOidcHandler` will not be called. This means that it is
+important to declare the dependencies from the most specialized to the most generic,
+because if a fallback handler that can handle all requests is declared first,
+the other more specialized handlers will not be called.
+
+Note that the label for the containers of the aggregator and the underlying handlers
+differ (respectively `browser:oidcHandler` and `browser:oidcHandlers`, with an 's',
+in the above example). In this case, the order in which the dependencies are
+declared is irrelevant, because they do not relate to the same container. The
```
kay-kim, line 169: For my own education -  What does "they do not relate to the same container" mean?  I see the same container registering the handlers.

kay-kim (commented):
Hey @NSeydoux  - I had to pause since I'm running off to get my hair cut (the first time in over a year - I know!?!)  
I wasn't sure about who the target audience is nor the style for this document (definitely recognizing the more relaxed nature of our github docs) and generally, it reads nicely.  So, I just made some comments here and there but definitely feel free to ignore any and all.  

I'll finish up the last bit after the hair chop.

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+- a request by the Client to the Issuer's token endpoint
+Handlers for the login operation are located in `packages/*/src/login/oidc/oidcHandlers/*Handler.ts`.
+
+#### Incoming redirect
+
+The incoming redirect is an operation initiated by the Issuer.
+At the Issuer webpage, the Resource Owner authenticates (e.g. by entering a username
+and a password), after which the Issuer sends them to a webpage under the Client
+app's control (its `redirect_uri`), to which it appends some query parameters that
+it can use to complete the login flow. This is done when the developer calls
```
kay-kim, line 124: the two "it" instances are a bit ambiguous.

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+Aggregator implements the class from `packages/core/src/util/handlerPattern/AggregateHandler.ts`,
+and uses the `@injectAll` annotation to receive all the handlers registered to a
+given container.
+
+#### Mocks and tests
+
+Dependency injection makes the codebase more flexible, because it is only at
+runtime that each component will be presented with the dependencies it declared,
+which means it is easier to add a dependency to a component without changing the
+whole codebase.
```
kay-kim, line 179: Maybe break into 2 sentences, something like:

> Dependency injection makes the codebase more flexible. Because each component is presented with its declared dependencies at runtime, it becomes easier to add a dependency to a component without changing the whole codebase.

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+
+However, mocking dependency injection in test code wouldn't bring any value.
```
kay-kim, line 181: Maybe just tweak to:

> However, for testing a component, mocking dependency injection in test code ...
> Instead, to test the object, construct the object ...

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Mapping OIDC flows to the code
+
+### Unsupported flows
+
+There are some OIDC flows that we intentionally don't plan on supporting:
+- The [Implicit flow](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth),
+which is widely recognized as having security issues, and doesn't bring value compared
+to the Auth Code flow.
+- The Refresh flow **in a browser context**. We do implement the efresh flow for
```
kay-kim, line 211: efresh -> Refresh

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+NodeJS, but there are limitations that prevent it from being applicable in a
+browser context:
+  - Users must always be prompted when a refresh token is requested, which prevents
+  silent authentication.
+  - There is no secure place where the Refresh Token may be stored in the browser
+  across page reload,  which means the token would be lost on reload, which defeats
+  the purpose of having a long-lived token anyway. The 
+  [best security practices document for OAuth](https://tools.ietf.org/html/draft-ietf-oauth-security-topics-18#section-4.13)
+  contains some recommandations on handling Refresh Tokens securely.
```
kay-kim, line 220: recommendations

ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+### Helpful resources
+
+Here is a list of things that helped this library's developers get a better understanding of OAuth/OIDC: 
+- [OAuth 2 in Action](https://www.manning.com/books/oauth-2-in-action), by Justin Richer and Antonio Sanso
+- [OAuth masterclass](https://www.youtube.com/watch?v=egfyV2NV9Mw), by Justin richer
+- [How To Securely Implement Authentication in Single Page Applications](https://betterprogramming.pub/how-to-securely-implement-authentication-in-single-page-applications-670534da746f), by Dennis Stötzel
+
+### Solid-OIDC
+
+Solid extends the OIDC protocol in order to make it fit into a decentralized ecosystem. The [Solid-OIDC specification](https://solid.github.io/authentication-panel/solid-oidc/) is the reference document to learn about this extension.
```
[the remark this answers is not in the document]
NSeydoux: 3775f329b5e58f2eef502e41b5fe872233c7033e

ARCHITECTURE.md
NSeydoux: 4a9e1d4bc4eb45508cf141ac1ac1318ee824fba8

ARCHITECTURE.md
```diff
@@ -0,0 +1,216 @@
+# Solid-client-authn-* architecture
+
+This document summerizes the architecture of the `@inrupt/solid-client-authn-*`
```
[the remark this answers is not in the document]
NSeydoux: Feels less like summer though

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+means this single git repository actually hosts several NPM modules which are
+related to each other. The following diagram shows an overview of the modules and their relations.
+
+![Module dependencies](./documentation/diagrams/module_map.svg)
+
+The two modules grouped under the "Client libraries" label are the
+modules we expect our customers to import. As their names imply, each  of these 
+modules is specific to a given environment (NodeJS or the browser). However, they
+both have a very similar API and architecture, and mostly differ by their main
+dependency, namely the third-party library implementing the
```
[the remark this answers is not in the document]
NSeydoux: Does c202c32ba5a68b3daeb7169ad730836e4f2df9a3 flow better ?

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+
+The four modules are available in the [packages directory](./packages).
+
+## OAuth2.0/OpenID Connect
+
+This library aims at helping developers authenticating users to their applications
+using the [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
+protocol (often abbreviated OIDC). OIDC is a protocol based on the
+[OAuth2.0](https://tools.ietf.org/html/rfc6749) framework. In order to understand
+how the library internally works, some understanding of OAuth/OIDC is preferable.
```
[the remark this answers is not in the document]
NSeydoux: It feels a little bit like removing the section and moving the subsections as preface to the following section would kind of bloat the next section, while keeping a pretty similar logical structure... However, what do you think about moving the section `Mapping OIDC flows to the code` up, and merging it with `0Auth2.0/OpenID Connect` ?

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+Here is a list of terms having a specific meaning in the context of OIDC:
+- **Resource Server**: The server hosting private resources. In our case, a Solid
+server. A Resource Server receives requests authenticated with an Access Token.
+Example: https://pod.inrupt.com. 
+- **Resource Owner**: the user, who owns some private resources 
+- **OIDC issuer**: the Solid Identity Provider, which issues Access Tokens, ID
+tokens, and Refresh Tokens. These tokens tell the Resource Server that the user
+has control over a certain identity (WebID), which can then use that information
+to decide whether to give or deny access. Example: https://broker.pod.inrupt.com.
+- **Client**: the application the Resource Owner uses to access its resources on
```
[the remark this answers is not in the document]
NSeydoux: The notion of Resource Owner in the OIDC sense is broader than the notion of Pod Owner in Solid: anyone with a WebID is a Resource Owner, in the authentication sense. Whether a particular Resource Owner is authorized to access a particular Resource is a different matter, up to the authozization system, so ACP or ACL, which is outside of the scope of this document. Do you think that's a helpful precision to add ?

ARCHITECTURE.md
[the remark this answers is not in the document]
kay-kim: Oh ... I think that would be really great.

ARCHITECTURE.md
```diff
@@ -0,0 +1,241 @@
...
+using the [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
+protocol (often abbreviated OIDC). OIDC is a protocol based on the
+[OAuth2.0](https://tools.ietf.org/html/rfc6749) framework. In order to understand
+how the library internally works, some understanding of OAuth/OIDC is preferable.
+
+### Helpful resources
+
+Here is a list of things that helped this library's developers get a better
+understanding of OAuth/OIDC: 
+- [OAuth 2 in Action](https://www.manning.com/books/oauth-2-in-action), by Justin Richer and Antonio Sanso
```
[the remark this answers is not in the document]
kay-kim: That works.  Definitely am against duplicating (and if the diagrams of the auth code flow moves on up to this section, then, the diagrams can act as a nice precursor to "hey, here are some resources that may help").

ARCHITECTURE.md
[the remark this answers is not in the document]
kay-kim: Yes please.  Because the Resource Owner definition follows that of the Resource Server - which ends with example https://pod.inrupt.com, it because easier to make the mental jump to equate Resource Owner to Pod Owner.  So, explicitly stating that it's anyone with a WebID clarification would definitely help.

ARCHITECTURE.md
NSeydoux: It's a concept really specific to TSyringe, so you don't need to worry too much about it :). What I'm trying to mean is that the order in which the `container` object registers dependencies isn't relevant  in the case they are registered with different labels. Maybe that's actually the way this should be phrased :D

ARCHITECTURE.md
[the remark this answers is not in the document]
NSeydoux: Done in 3da9f3dfd00e33393dafa1688b9dfe4d1cf6e839

ARCHITECTURE.md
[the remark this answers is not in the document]
NSeydoux: Done in 3da9f3dfd00e33393dafa1688b9dfe4d1cf6e839

NSeydoux:
BTW, how did the haircut go :D ?

kay-kim:
> BTW, how did the haircut go :D ?

Oh, it went well.  There was so much hair on the floor after she cut it -- it was the size of a small lhasa apso.  It feels so civilized to get a hair cut now :D

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+As their names imply, each  of these modules is specific to a given environment 
+(NodeJS or the browser). However, they both have very similar APIs and architecture,
+and mostly differ by their main dependency, namely the third-party library implementing the
+[OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) protocol.
+
+- `@inrupt/solid-client-authn-node` depends on [`openid-client`](https://github.com/panva/node-openid-client/).
+- `@inrupt/solid-client-authn-browser` depends on `@inrupt/oidc-client-ext`. 
+`@inrupt/oidc-client-ext` in turn depends on
+[`oidc-client`](https://github.com/IdentityModel/oidc-client-js),
+extending it with features we need that it does yet implement, namely support for
```
kay-kim, line 26: it does yet implement -> it does not yet implement  ?

Although, I wonder if we can just shorten to 

extending it with support for DPoP tokens and Dynamic Client Registration

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+[DPoP tokens](https://tools.ietf.org/html/draft-ietf-oauth-dpop-01) and 
+[Dynamic Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html).
+
+The four modules are available in the standard Lerna [packages directory](./packages).
+
+## OAuth2.0/OpenID Connect
+
+The client libraries aim to help developers authenticating the users of their applications
```
kay-kim, line 34: maybe (?) help developers authenticate their application users via the OpenID Connect .. ?

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+protocol (often abbreviated OIDC). OIDC is an industry-wide standard protocol based on the
+[OAuth2.0](https://tools.ietf.org/html/rfc6749) framework. In order to understand
+how our libraries work internally, some understanding of OAuth/OIDC is definitely preferable.
+
+### A short glossary
+
+Here is a list of terms having a specific meaning in the context of OIDC:
+- **Resource Owner**: the user, who owns resources, some of which are private.
+The notion of Resource Owner in the OIDC sense is broader than the notion of Pod Owner
+in Solid: anyone with a WebID is a Resource Owner, in the authentication sense.
```
kay-kim, line 45: Probably don't need ", in the authentication sense"  since we stated "in the OIDC sense" earlier in the sentence.

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+Whether a particular Resource Owner is **_authorized_** to access a particular Resource
+is a different matter, up to the Solid authorization system (e.g., ACP or ACL), which is
+outside the scope of this document.
+- **Resource Server**: the server hosting resources owned by the Resource Owner.
+In our case, a Solid server. A Resource Server receives requests authenticated
+with an Access Token. Example: https://pod.inrupt.com. 
+- **OIDC issuer**: the Solid Identity Provider, which issues Access Tokens, ID
+tokens, and Refresh Tokens. These tokens tell the Resource Server that the user
+has control over a certain identity (WebID), which can then use that information
```
kay-kim, line 54: which can then use ->  which the Server can then use ... (?)

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+to decide whether to give or deny access. Example: https://broker.pod.inrupt.com.
+- **Client**: the application the Resource Owner uses to access resources on
+a Resource Server. Those resources may be owned directly by the Resource Owner
+(i.e., resources in a Pod created by the resource owner), or on a Resource Server
+hosting someone else's resources, but to which this client application and user
+have been explicitly granted access).
```
kay-kim, line 60: extraneous ")"

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+
+- Solid-OIDC also makes the support for [Key-bound Access Tokens](https://tools.ietf.org/html/draft-fett-oauth-dpop-04)
+(referred to as DPoP tokens) mandatory: it is only optional in traditional OIDC, where
+Bearer tokens are the default option. DPoP tokens cannot be replayed by a Resource
+Server to another Resource Server, which is an important security feature in a
+decentralized ecosystem such as Solid's.
+
+### Mapping OIDC flows to the code
+
+#### Unsupported flows
```
kay-kim, line 84: Should this section follow the auth code flow and the refresh code flow? so that we hit the supported flows first?

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+Various components, such as the Login and Incoming Redirect components, are based
+on the Handler design pattern. Given data contained in a request, a set of classes
+implementing a similar API will declare whether or not they may handle said request.
+
+- **Handlers** declare two functions:
+  - `canHandle(request)`
+  - `handle(request)`
+  
+  `canHandle()` simply returns a boolean indicating the ability of the handler to handle the 
+  request, and `handle()` actually processes it.
```
kay-kim, line 158: I would maybe shoehorn these in with the bullet points above

- `canHandle(request)`, which returns a boolean indicating the handler's ability to handle the request
- `handle(request)`, which actually processes the request

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+  
+- All the handlers for a given type of request are tracked by a **Handler Aggregator**,
```
kay-kim, line 160: Maybe flip the sentence to start with Handler Aggregator

- **Handler Aggregator** tracks all the handlers for a given type of request. The handler aggregator invokes the first handler for which ...

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+
+### Dependency injection
+
+An important architectural component of this library is dependency injection,
+implemented here using [TSyringe](https://github.com/Microsoft/tsyringe). Dependencies
+are declared in `packages/*/src/dependencies.ts`.
+
+#### Declaring order
+
+The order in which the dependencies are declared (for a given container) **matters**.
```
kay-kim, line 199: **matters** when the dependencies have the same label (?)

ARCHITECTURE.md
```diff
@@ -0,0 +1,265 @@
...
+will handle the request, and the instances of `AuthorizationCodeWithPkceOidcHandler`
+and `ClientCredentialsOidcHandler` will not be called. This means that it is
+important to declare the dependencies from the most specialized to the most generic,
+because if a fallback handler that can handle all requests is declared first,
+the other more specialized handlers will never be called.
+
+Note that the label for the containers of the aggregator and the underlying handlers
+differ in the example above:
+- `browser:oidcHandler` (without an 's') for the aggregator.
+- `browser:oidcHandlers` (with an 's'), for the aggregated handlers.
```
kay-kim, line 231: I would probably move this up a little?  

Here, `AggregateOidcHandler` is the handler aggregator (as defined in the Handler
Pattern section), and `RefreshTokenOidcHandler`, `AuthorizationCodeWithPkceOidcHandler`
and `ClientCredentialsOidcHandler` are its underlying handlers. 

Note that the label for the aggregator containers differ from that for the underlying handlers (or something like that):
- `browser:oidcHandler` (without an 's') for the aggregator
- `browser:oidcHandlers` (with an 's') for the aggregated handlers.

When receiving a
request, `AggregateOidcHandler` will first invoke its instance of `RefreshTokenOidcHandler`
to check if it can handle it. If so, that instance of `RefreshTokenOidcHandler`
will handle the request, and the instances of `AuthorizationCodeWithPkceOidcHandler`
and `ClientCredentialsOidcHandler` will not be called. This means that it is
important to declare the dependencies from the most specialized to the most generic,
because if a fallback handler that can handle all requests is declared first,
the other more specialized handlers will never be called.

// then we can just flow into the sections when the order isn't relevant

Vinnl (approved):

PR #1339, opened 2021-05-06
Added dependency from oidc to core package

NSeydoux:
Initially, `oidc-client-ext` was meant to be unrelated to anything
solid-specific, which is why it wasn't dependant on
`solid-client-authn-core`. However, the `core` package mostly containing
type definitions and shared utils functions between the browser and node
modules, it isn't a big dependency on `oidc-client-ext`, but it reduces
dependency, so the benefits outwheight the costs.

Also, migrating to josev@3 will increase the portion of shared code
related to crypto and dpop, making the split between core and oidc even
more painful.

This changes nothing to the public API.

Note: I'll update the architecture diagram in #1338 accordingly.


packages/browser/package-lock.json
```diff
@@ -9,8 +9,6 @@
       "version": "1.8.1",
       "license": "MIT",
       "dependencies": {
-        "@inrupt/oidc-client-ext": "^1.8.1",
```
Vinnl, line 12: I don't understand how this leads to changes in the lockfile of `-browser`? Shouldn't it still be depending on the stable version of `-oidc` that does not yet depend on `-core`? (I really don't know what Lerna is doing, but I feel like it messes up the lockfiles continuously and practically makes them worthless :/ )

Vinnl (approved):

packages/browser/package-lock.json
NSeydoux: It's a bit worrying, but I don't know either. I did run `npm install` in the oidc module, and then `npm ci` at the top level, which triggers `lerna bootstrap`, and I think that may be what messes up the `package-lock`, but I'm not sure, nor can I tell if that's the right thing to do. I'll dig into separately.

PR #1346, opened 2021-05-07
Fix lodash vulnerability

NSeydoux:
This reaplies #1341 and #1343 , and also prevents https://www.npmjs.com/advisories/1677 from impacting us until a fix is published by only running `npm audit` on production dependencies. Running `npm audit` on dev dependencies could be done in a pre-commit hook (which is easier to ignore than a CI configuration).

See https://github.com/inrupt/solid-client-authn-js/pull/1347 to keep running the same audit level at a different stage of the development pipeline.


package.json
```diff
@@ -9,11 +9,11 @@
     "clean": "rimraf packages/*/dist/ packages/*/node_modules/ packages/*/coverage/ packages/*/umd/",
     "build-api-docs": "lerna run build-api-docs",
     "format-all": "prettier --write \"packages/*/src/**\" \"packages/*/__tests__/**\"",
-    "lerna-audit": "lerna-audit",
+    "lerna-audit-production": "lerna exec \"npm audit --production\"",
```
ajacksified, line 12: Why production and not also devdeps?

Vinnl: I'm not sure if that's a good idea - devDependencies can still impact downstreams. For example, if one of our build tools is compromised, it can inject code into the production bundle. Or if it's used to compromise a developer machine or our CI environment, it might get hold of credentials that can be used to push code to downstreams.

Vinnl: After discussing with Nic elsewhere, the real issue here is that `npm audit` fails on an issue in a dependency for which no fix is available yet, and it's not possible to set it up such that it only fails when there's actually something we can do to resolve it.

Thus, I think that if we mark this explicitly as a temporary workaround to deal with that specific issue, and when it can be reverted, then applying this is fine.

PR #1355, opened 2021-05-10
Clear all node_modules when running npm ci

NSeydoux:
Before running `lerna bootstrap`, this removes all the 'node_modules' in
all submodules, because that's what a regular `npm ci` does, and doing
it manually resolves issues that are happening otherwise.


NSeydoux:
The problem this intended to solve no longer manifests itself. Let's wait and see.

PR #1356, opened 2021-05-10
Migrate browser code to jose@3.x

NSeydoux:
This migrates the browser module to jose@3.x. There are several consequences to this change, explaining the size of the PR: 
- It enables migrating some code from environment-specific modules to the core module, which is why the node module is impacted too. Now, DPoP key pair generation or DPoP header creation is implemented once in the core module.
- The dependencies of both `oidc-client-ext` and `browser` changed, which modified the package-locks, which largely contributes to the size of the PR
- It looks like a recent change in the linter modified multiple files when running tests
- The node builtins are no longer required in the webpack configuration

An important note is that this upgrade requires Webpack 5, because `jose` only uses an `exports` map to make its submodules available. Having an `index` file with a bulk export would have more retro-compatibility, this is a conversation we'll engage with the library maintainer.

- [ ] Relevant documentation, if any, has been written/updated. (N/A)
- [ ] The changelog has been updated, if applicable. (N/A)
- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/). (Mostly)


Vinnl:
Superseded by #1426.

PR #1360, opened 2021-05-11
Set sessionInfo after completing silent auth

Vinnl:
After completing silent authentication in an iframe, the `info` property on the Session wasn't updated. It is now, and I've extracted the code for doing so, which is shared with `handleIncomingRedirect`, into the `setSessionInfo` method, primarily so that there's a single place to change to make sure the session is silently refreshed rather than terminated when it expires.

- [x] The changelog has been updated, if applicable. N/A

### DIFF
--- a/packages/browser/__tests__/Session.spec.ts
+++ b/packages/browser/__tests__/Session.spec.ts
@@ -1020,5 +1020,49 @@ describe("Session", () => {
         inIframe: true,
       });
     });
+
+    it("sets the updated session info after silently refreshing", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      const incomingRedirectPromise = Promise.resolve({
+        isLoggedIn: true,
+        webId: "https://some.pod/profile#me",
+        sessionId: "someSessionId",
+        expirationDate: 961106400,
+      });
+      clientAuthentication.handleIncomingRedirect = jest
+        .fn()
+        .mockReturnValueOnce(incomingRedirectPromise);
+
+      const windowAddEventListener = jest.spyOn(window, "addEventListener");
+      // ../src/iframe is mocked for other tests,
+      // but we need `setupIframeListener` to actually be executed
+      // so that the callback gets called:
+      const iframeMock = jest.requireMock("../src/iframe");
+      const iframeActual = jest.requireActual("../src/iframe");
+      iframeMock.setupIframeListener.mockImplementationOnce(
+        iframeActual.setupIframeListener
+      );
+      const mySession = new Session({ clientAuthentication });
+
+      // `window.addEventListener` gets called once with ("message", handler)
+      // — get that handler.
+      const messageEventHandler = windowAddEventListener.mock
+        .calls[0][1] as EventListener;
+      const mockedEvent = {
+        origin: window.location.origin,
+        source: null,
+        data: {
+          redirectUrl: "http://arbitrary.com",
+        },
+      } as MessageEvent;
+      // This handler will call `clientAuthentication.handleIncomingRedirect`,
+      // which will return our sessionInfo values:
+      messageEventHandler(mockedEvent);
+
+      await incomingRedirectPromise;
+      expect(mySession.info.webId).toBe("https://some.pod/profile#me");
+      expect(mySession.info.sessionId).toBe("someSessionId");
+      expect(mySession.info.expirationDate).toBe(961106400);
+    });
   });
 });
--- a/packages/browser/src/Session.ts
+++ b/packages/browser/src/Session.ts
@@ -126,6 +126,12 @@ export async function silentlyAuthenticate(
   }
 }
 
+function isLoggedIn(
+  sessionInfo?: ISessionInfo
+): sessionInfo is ISessionInfo & { isLoggedIn: true } {
+  return !!sessionInfo?.isLoggedIn;
+}
+
 /**
  * A {@link Session} object represents a user's session on an application. The session holds state, as it stores information enabling acces to private resources after login for instance.
  */
@@ -187,7 +193,21 @@ export class Session extends EventEmitter {
       };
     }
     // Listen for messages from children iframes.
-    setupIframeListener(this.clientAuthentication.handleIncomingRedirect);
+    setupIframeListener(async (redirectUrl: string) => {
+      const sessionInfo = await this.clientAuthentication.handleIncomingRedirect(
+        redirectUrl
+      );
+
+      // If silent authentication was not successful, do nothing;
+      // the existing session might still be valid for a while,
+      // and will expire by itself.
+      if (!isLoggedIn(sessionInfo)) {
+        return;
+      }
+      // After having revalidated the session,
+      // make sure to apply the new expiration time:
+      this.setSessionInfo(sessionInfo);
+    });
     // Listen for the 'tokenRenewal' signal to trigger the silent token renewal.
     this.on("tokenRenewal", () =>
       silentlyAuthenticate(this.info.sessionId, this.clientAuthentication, {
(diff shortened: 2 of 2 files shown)


packages/browser/src/Session.ts
```diff
@@ -187,7 +193,16 @@ export class Session extends EventEmitter {
       };
     }
     // Listen for messages from children iframes.
-    setupIframeListener(this.clientAuthentication.handleIncomingRedirect);
+    setupIframeListener(async (redirectUrl: string) => {
+      const sessionInfo = await this.clientAuthentication.handleIncomingRedirect(
+        redirectUrl
+      );
+
+      if (!isLoggedIn(sessionInfo)) {
```
pmcb55, line 201: If it's possible for this handler to be invoked when already logged in, then should we still call `setSessionInfo()` regardless, i.e., maybe we were called because we want to reset our session's access token timer to a new expiration date?
Or to state this another way, is there a very specific reason that we don't want to call `setSessionInfo()`? (If so, then a comment is warranted perhaps?)

Vinnl: That's a good question. @NSeydoux, do you know the answer? 'cause I don't...

packages/browser/src/Session.ts
```diff
@@ -424,4 +431,17 @@ export class Session extends EventEmitter {
...
     this.on("sessionRestore", callback);
   }
+
+  private setSessionInfo(
+    sessionInfo: ISessionInfo & { isLoggedIn: true }
+  ): void {
+    this.info.isLoggedIn = sessionInfo.isLoggedIn;
+    this.info.webId = sessionInfo.webId;
+    this.info.sessionId = sessionInfo.sessionId;
+    if (typeof sessionInfo.expirationDate === "number") {
```
NSeydoux, line 441:
```suggestion
    this.info.sessionId = sessionInfo.sessionId;
    this.info.expirationDate = sessionInfo.expirationDate;
    if (typeof sessionInfo.expirationDate === "number") {
```
This will allow, in an upcoming ticket, to change the behaviour of the callback set by `setTimeout` to checking if the current expiration date (`this.info.expirationDate`) is close to `Date.now()` (with a certain tolerance). This way, if silent authentication happens in an `iframe`, the callback will not log the session out.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -1020,5 +1020,74 @@ describe("Session", () => {
...
+        redirectUrl: "https://some.redirect/url",
+        tokenType: "DPoP",
+      });
+      clientAuthentication.validateCurrentSession = jest
+        .fn()
+        .mockReturnValue(validateCurrentSessionPromise);
+      const incomingRedirectPromise = Promise.resolve({
+        isLoggedIn: true,
+        webId: "https://some.pod/profile#me",
+        sessionId: "someSessionId",
```
NSeydoux, line 1054:
```suggestion
        sessionId: "someSessionId",
        expirationDate: 961106400
```
This sets a new expiration date for the access token held by the session. It happens to be the release date of Pokemon yellow in Europe, arguably an excellent day, but that's purely coincidental.

packages/browser/__tests__/Session.spec.ts
```diff
@@ -1020,5 +1020,74 @@ describe("Session", () => {
...
+        },
+      } as MessageEvent;
+      // This handler will call `clientAuthentication.handleIncomingRedirect`,
+      // which will return our sessionInfo values:
+      messageEventHandler(mockedEvent);
+
+      await incomingRedirectPromise;
+      await validateCurrentSessionPromise;
+      expect(mySession.info.webId).toBe("https://some.pod/profile#me");
+      expect(mySession.info.sessionId).toBe("someSessionId");
```
NSeydoux, line 1090:
```suggestion
      expect(mySession.info.sessionId).toBe("someSessionId");
      expect(mySession.info.expirationDate).toBe(961106400);
```
I'm repeating myself but... what a date !

packages/browser/__tests__/Session.spec.ts
```diff
@@ -1020,5 +1020,74 @@ describe("Session", () => {
...
+        data: {
+          redirectUrl: "http://arbitrary.com",
+        },
+      } as MessageEvent;
+      // This handler will call `clientAuthentication.handleIncomingRedirect`,
+      // which will return our sessionInfo values:
+      messageEventHandler(mockedEvent);
+
+      await incomingRedirectPromise;
+      await validateCurrentSessionPromise;
```
NSeydoux, line 1088:
```suggestion
```

packages/browser/src/Session.ts
```diff
@@ -187,7 +193,16 @@ export class Session extends EventEmitter {
...
-    setupIframeListener(this.clientAuthentication.handleIncomingRedirect);
+    setupIframeListener(async (redirectUrl: string) => {
+      const sessionInfo = await this.clientAuthentication.handleIncomingRedirect(
+        redirectUrl
+      );
+
+      if (!isLoggedIn(sessionInfo)) {
+        return;
+      }
+      this.setSessionInfo(sessionInfo);
```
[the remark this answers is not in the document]
Vinnl: Yes, because I'd actually forgotten as well: 4dd2b96c

packages/browser/__tests__/Session.spec.ts
```diff
@@ -1020,5 +1020,74 @@ describe("Session", () => {
         inIframe: true,
       });
     });
+
+    it("sets the updated session info after silently refreshing", async () => {
+      const sessionId = "mySession";
+      mockLocalStorage({
+        [KEY_CURRENT_SESSION]: sessionId,
+      });
```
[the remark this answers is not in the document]
Vinnl: Good point, I was kind of intimidated by all the mocking ceremony in the other test, but I should've checked that actually most of it isn't needed for this test: 0b4e6ddc

packages/browser/__tests__/Session.spec.ts
```diff
@@ -1020,5 +1020,74 @@ describe("Session", () => {
...
+      const validateCurrentSessionPromise = Promise.resolve({
+        issuer: "https://some.issuer",
+        clientAppId: "some client ID",
+        clientAppSecret: "some client secret",
+        redirectUrl: "https://some.redirect/url",
+        tokenType: "DPoP",
+      });
+      clientAuthentication.validateCurrentSession = jest
+        .fn()
+        .mockReturnValue(validateCurrentSessionPromise);
```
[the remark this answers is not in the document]
Vinnl: Good point, I was kind of intimidated by all the mocking ceremony in the other test, but I should've checked that actually most of it isn't needed for this test: 0b4e6ddc

packages/browser/src/Session.ts
Vinnl: Whoops yeah, that's kinda important: 4134dc9f

packages/browser/__tests__/Session.spec.ts
Vinnl: 4134dc9f

packages/browser/__tests__/Session.spec.ts
Vinnl: 4134dc9f

packages/browser/__tests__/Session.spec.ts
Vinnl: 0b4e6ddc

packages/browser/src/Session.ts
NSeydoux: Well technically, the redirection failing does not mean that the session isn't logged out. Imagine that a user initially logs in an gets an access token valid for 30 minutes. If we try to perform a silent authentication in an iframe 10 minutes before the expiration, and it fails, then the session isn't logged out for another 10 minutes, at which point the callback we set initially when receiving the token fires and the session gets properly logged out, emitting a logout signal, clearing the storage and all. This is exaggerated, but the idea is that if the silent authentication is successful, we want to update the session info in particular with the new expiration date, but if it failed we don't do anything and have dedicated code properly logging out the session.

So long story short I think the current implementation is the "right" thing to do, but we may add a comment specifying that the proper logout happens in a callback set when receiving the access token.

Vinnl: Makes sense, thanks! Added comment: 0d15be9a

packages/browser/src/Session.ts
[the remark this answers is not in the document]
NSeydoux: "Your comments are for your future self" is one of the best pieces of advice I've had from a CS teacher :D

packages/browser/__tests__/Session.spec.ts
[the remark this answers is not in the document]
NSeydoux: That's fair, setting up the mocks where there are so much side effects is kind of involved

PR #1388, opened 2021-05-25
Fix issue identified by npm audit

Vinnl:
This is blocking other PR pipelines from succeeding.


package-lock.json
```diff
@@ -1,18474 +1,7 @@
 {
   "name": "root",
-  "lockfileVersion": 2,
```
Vinnl, line 3: I don't know why this was set to version 2 (which I think is npm@7's) - according to the .nvmrc we should be using Node 14 which I think comes with npm@6 still. Another data point that maybe it's time to switch to Volta, which allows us to lock the package manager version specifically.

VirginiaBalseiro (approved):

PR #1398, opened 2021-05-27
Update browser-based E2E tests for broker changes

Vinnl:
Unfortunately using the `.withText` selector didn't seem to work
for one reason or another. Investigating why didn't seem worth the
effort, hence just matching on the DOM structure for now.

### DIFF
--- a/e2e/browser/helpers/authorizeClientApp.ts
+++ b/e2e/browser/helpers/authorizeClientApp.ts
@@ -15,9 +15,9 @@ export async function authorizeNss() {
 }
 
 export async function authorizeEss() {
-  const approveButton = Selector("button").withText("Approve");
+  // Ideally we'd find the button using the `.withText` Selector method,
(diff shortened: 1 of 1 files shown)


e2e/browser/helpers/authorizeClientApp.ts
```diff
@@ -15,9 +15,6 @@ export async function authorizeNss() {
 }
 
 export async function authorizeEss() {
-  const approveButton = Selector("button").withText("Approve");
+  const approveButton = Selector('button[form="approve"]');
```
pmcb55, line 18: Made just a quick comment to the effect that the 'better' selection isn't working (but we don't know why yet)...?

pmcb55 (approved):

e2e/browser/helpers/authorizeClientApp.ts
Vinnl: Fair enough! 750b6070

PR #1411, opened 2021-05-31
Remove ajv, validateSchema and safeGet

Vinnl:
They appear not to be used, and add quite a bit to the bundle size. @pmcb55 Of course I'm a bit nervous just assuming that the code is actually unused - could you give a second opinion as to whether that's likely or not?

It does shave about 130kB from our bundled size, bringing it from [ridiculously big](https://bundlephobia.com/result?p=@inrupt/solid-client-authn-browser@1.8.2-main-893142833-6561-1622469620.0) to [still ridiculously big](https://bundlephobia.com/result?p=@inrupt/solid-client-authn-browser@1.8.2-choreremove-ajv-893209213-6562-1622470990.0) (but slightly less so) :P


pmcb55 (approved):
Yeah, I checked through the code, and indeed, the 'safeGet' isn't being used anywhere. The whole idea of trying to validate a providers OIDC config meta-data seems weird anyway (i.e., as opposed to just [duck-typing](https://en.wikipedia.org/wiki/Duck_typing)), so happy to see this removed.

Vinnl:
Great, thanks for the check!

PR #1426, opened 2021-06-07
Migrate browser code to jose@3.x

Vinnl:
# New feature description

This re-applies the changes from #1356, but uses `@inrupt/jose-legacy-modules` instead of jose@3 directly, to keep it working in tools that do not support package exports yet, like Webpack 4. As mentioned in that PR:

> - It enables migrating some code from environment-specific modules to the core module, which is why the node module is impacted too. Now, DPoP key pair generation or DPoP header creation is implemented once in the core module.
> - The dependencies of both `oidc-client-ext` and `browser` changed, which modified the package-locks, which largely contributes to the size of the PR
> - The node builtins are no longer required in the webpack configuration

I also did not apply the migration to Webpack 5 (since this should work with 4 too), nor the Prettier changes.

~~I would also like to try this in a real app and see if anything breaks, but that needn't hold up reviewing this.~~ Confirmed working in Penny.

# Checklist


packages/core/src/authenticatedFetch/dpopUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+import { v4 } from "uuid";
+
+/**
+ * Normalizes a URL in order to generate the DPoP token based on a consistent scheme.
+ *
+ * @param audience The URL to normalize.
+ * @returns The normalized URL as a string.
+ * @hidden
+ */
+function normalizeHttpUriClaim(audience: string): string {
```
pmcb55, line 38: Personally, for such a generic function, I'd just name it more explicitly (i.e., the function itself has nothing to do with DPoP per se, or 'uri claims'). I would probably attempt to explain that it can be used, for example, to normalize URLs for reasons of comparability (at least as far as DPoP is concerned, perhaps for other uses too):
```suggestion
function removeHashUsernameAndPassword(url: string): string {
```

packages/core/src/authenticatedFetch/dpopUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+  const cleanedAudience = new URL(audience);
+  cleanedAudience.hash = "";
+  cleanedAudience.username = "";
+  cleanedAudience.password = "";
+  return cleanedAudience.toString();
+}
+
+export type DpopKeyPair = {
```
pmcb55, line 46: Seems like just a KeyPair to me (i.e., is there anything DPoP-specific in it...?)
```suggestion
export type KeyPair = {
```

packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
```diff
@@ -41,3 +47,84 @@ type IRedirectHandler = IHandleable<
...
+      `Malformed JWKS for [${issuerIri}] at [${jwksIri}]: ${e.message}`
+    );
+  }
+  return jwk;
+}
+
+/**
+ * Extract a WebID from an ID token payload based on https://github.com/solid/webid-oidc-spec.
+ * Note that this does not yet implement the user endpoint lookup, and only checks
+ * for webid or IRI-like sub claims.
```
pmcb55, line 80:
```suggestion
 * for 'webid' or IRI-like 'sub' claims.
```

pmcb55 (approved):

PR #1435, opened 2021-06-08
Fix session restore loop

Vinnl:
This PR fixes the bug where `handleIncomingRedirect` does not resolve when attempting, but failing (e.g. when the session on the IDP has expired), to silently authenticate. This, in turn, causes solid-ui-react to never set `sessionRequestInProgress` to `false`, which then caused PodBrowser to keep attempting to login.

### DIFF
--- a/packages/browser/src/iframe.ts
+++ b/packages/browser/src/iframe.ts
@@ -23,9 +23,18 @@
 // which initializes it, and `setupIframeListener`, which reads from it. It avoids
 // creating a DOM element with identifying attributes such as an ID which may clash
 // with some other element set up by the library user.
-const iframe: HTMLIFrameElement = window.document.createElement("iframe");
-iframe.setAttribute("hidden", "true");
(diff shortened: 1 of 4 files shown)


packages/browser/src/iframe.ts
```diff
@@ -52,7 +62,15 @@ export function redirectInIframe(redirectUrl: string): void {
 export function setupIframeListener(
   handleIframeRedirect: (redirectUrl: string) => Promise<unknown>
 ): void {
+  // If an app is doing server-side rendering, setting up an iframe listener
+  // is pointless (since there is no iframe, and there will be no expiring
+  // session to renew) and might even through a `window` not defined error,
```
VirginiaBalseiro, line 67: typo: throw

VirginiaBalseiro (approved):
Just a typo :)

packages/browser/src/iframe.ts
Vinnl: Heh, whoops. Fixed.

PR #1447, opened 2021-06-15
Removed TSyringe completely from Node package

pmcb55:
Chore to start removing TSyringe completely - this PR removes it from the Node package (i.e., both `tsyringe` and `reflect-metadata` uninstalled from the `package.json` (and then `npm ci` re-run from root directory to kick-off Lerna)).
All tests still pass, when run from both the `packages/node` folder and from the root.
Basically the change here was to simply manually construct all required class instances directly in the function `getClientAuthenticationWithDependencies()`.


Vinnl (approved):
This is excellent! I was looking at removing TSyringe just for a single class and indeed concluded that that would involve removing it practically everywhere, but I didn't realise that that was actually doable without rewriting most of the codebase/tests.

Next steps will be changing the classes so as not to take their dependencies as parameters to the constructor, but to just `import` them directly - but now that this work is done, that can far more easily be done in a gradual fashion.

Btw, I mentioned that I was going to try this in Penny, but of course since this is in `-node`, I can't... I then wanted to try it using solid-client's end-to-end test, but that's reliant on #1329 (which is bound to have lots of merge conflicts), so unfortunately I couldn't do that either.

pmcb55:
What would the benefit be of removing the constructor params and using `import` directly instead? Maybe it's just my Java-ness, but passing in dependencies still seems to make sense to me (e.g., for passing in mocks, **_potentially_** new implementations, etc.)

Vinnl:
The primary benefit is in the browser, in that only the necessary code gets instantiated - if the developer does not use a particular path at all, the code can be removed altogether.

It's not necessary for mocks, since you can replace modules altogether. If we need new implementations, I generally have a preference anyway for defining the common API when needed, rather than trying to predict those use cases in advance and sacrificing dead code elimination when we can avoid it.

pmcb55:
Hmmm... Ok thanks for the elaboration Vinnie. Not being an JS expert, I'll defer on the whole 'ctor-params' vs 'direct import' question(!).
But I think it's an interesting **_general_** design decision for which I think input from @ajacksified and @RubenVerborgh might be interesting. (We could discuss internally (e.g., via Slack, Google, Confluence) or right here publicly (since we have a nice example now of a complex object creation in this PR (in the method `getClientAuthenticationWithDependencies()` of  file `packages/node/src/dependencies.ts`)).

packages/node/src/dependencies.ts
```diff
@@ -141,16 +54,53 @@ export function getClientAuthenticationWithDependencies(dependencies: {
...
+    clientRegistrar
+  );
+
+  return new ClientAuthentication(
+    // TODO: I don't think we need an Aggregate login handler here at all, since we only use one
+    //  anyways...!
+    new AggregateLoginHandler([
+      new OidcLoginHandler(
+        storageUtility,
+        new AggregateOidcHandler([
```
RubenVerborgh, line 80: Perhaps move this one up and assign to a variable? Just for readability.
Same with the other deep nestings.

RubenVerborgh:
I think the right choices have been made here.

We keep the benefits of injectable code, but remove the bloat of TSyringe.

The injection is now performed manually, which seems fine given the complexity of the code.

As a comparison, we follow the same injectable strategy with the Community Solid Server, but given that we have many more instantiations, we choose to not write the injection code manually but rather have Components.js perform that injection for us through configuration. In this library, we do not seem to have reached that tipping point yet.

An attention point for the future is to make sure we keep the injectable nature of the code, which is also great for testability.

RubenVerborgh:
> Next steps will be changing the classes so as not to take their dependencies as parameters to the constructor, but to just import them directly - but now that this work is done, that can far more easily be done in a gradual fashion.

I would hold back on that.

PR #1459, opened 2021-06-20
Removed TSyringe from browser package

pmcb55:
One thing I notice is that the `.codesandbox` project still has references to `reflect-meta-data` and `tysringe` in it's `package-lock.json` (even after I run `npm ci` and even `npm i`), but it appears it's `package.json` is way out-of-date in relation to SCAB:
```
  "dependencies": {
    "@inrupt/solid-client": "0.3.0",
    "@inrupt/solid-client-authn-browser": "^0.1.1",
:
:
```
But I don't want to (i.e., 'don't know **_how_** to **_correctly_**') screw around with those version numbers in this PR.


Vinnl (approved):
I'm not sure whether the CodeSandbox is actually used, so it's probably fine to leave it be for now - but since you're not messing with the `package.json`, I'll revert the changes to the sandbox's `package-lock.json`.

PR #1469, opened 2021-06-23
Bind fetch to window

Vinnl:
This PR fixes bug #944.

- [ ] I've added a unit test to test for potential regressions of this bug. - Didn't know how to, but let me know if you have suggestions here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Bugs fixed
+
+#### browser
+
+- Trying to call `Session.fetch` for a Session that had not yet authenticated
+  would result in the following error:
+    'fetch' called on an object that does not implement interface Window.
+
 The following sections document changes that have been released already:
 
 ## 1.9.0 - 2021-06-16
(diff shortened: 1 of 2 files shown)


Vinnl:
@matthieu-fesselier If you have time, it'd be great if you could verify whether this solves the issue you described in https://github.com/inrupt/solid-client-authn-js/issues/944#issuecomment-842262770?

You can try out the fix by running:

    npm install @inrupt/solid-client-authn-browser@fix944-fetch-window

RubenVerborgh:
Should we also/instead fix it in FallbackRedirectHandler?

If that's the one passing around the regular fetch function, then that's the one to blame. Plus we can then also unit test.

Simply said, we should never pass around window.fetch because it is a member method rather than a function. As a protection, we could also have a separate file exposing a bound window.fetch and type it differently (the original type should have a `this`!) to avoid the mistake.

Vinnl:
I think the problem is in the binding to ClientAuthentication, not in the passing around of `fetch`.

That said, yes, the way we generate `fetch` definitely needs overhauling. However, it's currently spread out over a codebase that I'm having trouble fitting in my mind, so I'd like to do that together with Nicolas.

RubenVerborgh:
> not in the passing around of `fetch`.

`fetch` is a method (https://fetch.spec.whatwg.org/#fetch-method) so should not be passed around a function.

> it's currently spread out over a codebase

The separate file approach might be useful; so we can lint out any usage of `fetch` elsewhere.

ajacksified (approved):
This solves the immediate issue; we can create a separate tech debt ticket to discuss how fetch should be handled across the codebase.

PR #1560, opened 2021-07-30
Factor fetch into core package

NSeydoux:
This is preliminary work to enable refresh tokens being used in the browser. The `fetchFactory.ts` file, which was environment-specific, has now been moved into the core package, since the main difference between its environment-specific implementations was the presence of the refresh token management. Moving the node behaviour into the core module enables reusing its already implemented refresh token management code.

# Checklist

- [ ] The changelog has been updated, if applicable. **N/A, this is purely internal**


packages/core/jest.config.js
```diff
@@ -2,4 +2,8 @@ module.exports = {
   ...require("../../jest.config"),
   // Allows running tests from packages/core/
   testEnvironment: "<rootDir>/../../tests/environment/customEnvironment.js",
+  moduleNameMapper: {
+    // Here, the OIDC package is used, but any package depending on jose@3.x would do.
+    "^jose/(.*)$": "<rootDir>/node_modules/jose/dist/node/cjs/$1",
+  },
```
Vinnl, line 8: This shouldn't be necessary in the `-core` package, because we should not be using plain `jose` there - even before #1516, as this package is a dependency of `-browser`.

packages/oidc/src/dpop/tokenExchange.ts
```diff
@@ -88,17 +89,11 @@ function hasExpiresIn(
...
   idToken: string;
   webId: string;
-  refreshToken?: string;
   dpopKey?: KeyPair;
-  expiresIn?: number;
-};
-
-export type TokenEndpointDpopResponse = TokenEndpointResponse & {
-  dpopKey: KeyPair;
 };
```
Vinnl, line 97: Is this no longer useful?

packages/node/src/login/oidc/refresh/TokenRefresher.ts
```diff
@@ -107,30 +122,25 @@ export default class TokenRefresher implements ITokenRefresher {
...
-      throw new Error(
-        `The Identity Provider [${issuer.metadata.issuer}] did not return an access token on refresh.`
-      );
-    }
+    const tokenSet = tokenSetToTokenEndpointResponse(
+      await client.refresh(refreshToken, {
+        // openid-client does not support yet jose@3.x, and expects
+        // type definitions that are no longer present. However, the JWK
+        // type that we pass here is compatible with the API, hence the `any`
+        // assertion.
```
Vinnl, line 130: I don't see `any` anymore :)

```suggestion
        // type that we pass here is compatible with the API, hence the type
        // assertion.
```

Vinnl (approved):
All looks fairly straightforward; just make sure that `-core` does not depend on `jose`.

packages/core/jest.config.js
NSeydoux: 30eba5f9348fc6eecbf3e124e83eea4f26397248 It wasn't actually depending on `jose`, but only due to the module mapping that tricked jest into having the appropriate behaviour ^^

packages/oidc/src/dpop/tokenExchange.ts
NSeydoux: Not really: I extended the `fetchFactory` so that it exports a single function (`buildAuthenticatedFetch`) which accepts a `KeyPair | undefined` rather than the factory exporting two functions, `buildBearerFetch` and `buildDpopFetch`. This way, the type checking is done in the core module only, rather than in each environment-specific module.

Vinnl: Gotcha, makes sense :+1:

PR #1561, opened 2021-07-30
Fix/node/submodule import

NSeydoux:
Using submodules import for jose creates interoperability issues with
parts of the ecosystem that are not up-to-speed yet with this mechanism,
e.g. jest. It is manageable internally, but it actually also impacts our
dependents, which we want to avoid as much as possible.

To work around the issue, all dependencies to jose go through an
intermediary packages which re-exports jose submodules as legacy
modules.

**Note to reviewers**: To avoid conflicts, this has been implemented on top of https://github.com/inrupt/solid-client-authn-js/pull/1560, so it may be easier to review that first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - A transitive dependency used submodule exports, which aren't supported yet by
 significant parts of the ecosystem, such as Jest. With an internal change, we enabled
-using @inrupt/solid-client-authn-node without modifying Jest configuration of
-our dependents. 
+using @inrupt/solid-client-authn-node without encountering submodule exports.
 
 The following sections document changes that have been released already:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Bugfixes
+
+#### node
+
+- A transitive dependency used submodule exports, which aren't supported yet by
+significant parts of the ecosystem, such as Jest. With an internal change, we enabled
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 ## Unreleased
 
+### Bugfixes
+
+#### node
+
+- A transitive dependency used submodule exports, which aren't supported yet by
+significant parts of the ecosystem, such as Jest. With an internal change, we enabled
+using @inrupt/solid-client-authn-node without modifying Jest configuration of
+our dependents.
```
Vinnl, line 17: Since there are more reasons than just Jest why this could be problematic, I'd stick to only mentioning it as an example of what can go wrong.

```suggestion
using @inrupt/solid-client-authn-node without encountering submodule exports.
```

packages/node/package.json
```diff
@@ -30,11 +30,11 @@
     "typescript": "^4.0.5"
   },
   "dependencies": {
+    "@inrupt/jose-legacy-modules": "*",
```
Vinnl, line 33: Why would we make this a wildcard dependency? It's just republished whenever `jose` is with a matching version number, so we can keep it up to date using Dependabot just like we do today.

```suggestion
    "@inrupt/jose-legacy-modules": "0.0.3-3.14.3",
```

(I suppose using the `^` could get us to suddenly match a major version bump, so pinning it to a specific version might be a good idea. Alternatively, we could rework jose-legacy-modules to switch the versions around, i.e. start with the jose version and then append the fork's version, since there's not going to be breaking changes in what the fork does anyway (I think).

Vinnl (approved):

packages/node/package.json
NSeydoux: Oops, it wasn't intentional, I wonder if that's NPM default behaviour for dependencies using unexpected version numbers format.

PR #1563, opened 2021-08-02
Remove all reference to REFRESH_TOKEN

dgbarclay:
Remove all refresh token references, secrets updated for repo.


NSeydoux (approved):

Vinnl:
@NSeydoux Could you verify if the change in 1a83d4635ce6dd91eb9b2300dbb360cd5780532b makes sense? I'm not sure whether the top-level Jest E2E config is used for anything other than `-node` (I was considering just removing it, and changing the `e2e-test` script into `cd packages/node; npm run e2e-test` if not), but as far as I can see they always run in Node and thus don't need to mock the browser environment? Without this change, the `TextEncoder` global will not be available, which jose assumes to be.

NSeydoux:
Jest indeed only runs the Node e2e tests, so it makes sense to move this to the top-level configuration. We may also consider moving the e2e tests altogether in an `e2e/node` folder, side-by-side with the `e2e/browser` folder, wher testcafé tests run. And only the node tests assume the `TextEncoder` global, right ?

Vinnl:
It's jose that assumes `TextEncoder`, regardless of the environment. It should usually be present in both Node and the browser, but I'm guessing that jsdom doesn't implement it, and therefore that the tests fail if calling jose in that environment. I'll move the tests to `/e2e`, that's where I was looking for them first anyway. (Although moving `/e2e/browser` to `/packages/browser` would make sense to me too. It's primarily the inconsistency that's confusing.)

NSeydoux:
Right, if it's more than moving the code around it may not be worth the effort

PR #1570, opened 2021-08-03
Move test files side-by-side with code

NSeydoux:
This aligns testing practices in the browser package with the other packages, namely by moving the test files side-by-side with the tested code. While doing this, it also deletes some files that are no longer used (the browser fetch having been replaced by the core fetch)


Vinnl (approved):
This is going to save so much jumping around when doing PR reviews :heart:

PR #1576, opened 2021-08-04
Call callbacks on events

NSeydoux:
This changes the way we manage user-provided callbacks, in particular the ones regarding refresh token rotation (`onNewResfreshToken`) and error-handling (`onError`). Instead of having the user specify the callback as a parameter to either the session constructor, as it is currently the case for `onNewResfreshToken`, or as a parameter to `handleincomingRedirect` as it is currently the case for `onError` in the browser. 

In the internal architecture, this change involves that instead of passing around some callbacks (of which there could be many), we pass around a single event emitter (which happens to be the base `Session` object, but that's an implementation detail), and emit events in different situation through this emitter. Users can subscribe to events using the session's `onXXX` methods, just like they do for `onLogin` or `onLogout`.

For backwards compatibility, the Node session should still accept the `onNewRefreshToken` callback as a constructor parameter until next major release, but we should mark it as deprecated, and it will under the cover register the callback to the appropriate event anyway. 

The browser session hasn't been released yet with the `onError` callback as a `handleIncomingRedirect` parameter, so I'd suggest we just remove it from there and implement an `onError` method instead.

# New feature description

# Checklist


packages/browser/src/Session.ts
```diff
@@ -120,16 +115,19 @@ export async function silentlyAuthenticate(
...
+        sessionId,
+        prompt: "none",
+        oidcIssuer: storedSessionInfo.issuer,
+        redirectUrl: storedSessionInfo.redirectUrl,
+        clientId: storedSessionInfo.clientAppId,
+        clientSecret: storedSessionInfo.clientAppSecret,
+        tokenType: storedSessionInfo.tokenType ?? "DPoP",
+        inIframe: options.inIframe,
+      },
+      getDefaultSession()
```
NSeydoux, line 129:
```suggestion
      this
```
You don't want to necessarily use the default session, but rather the current one, be it the default or one that has been created manually by the app.

packages/browser/src/login/oidc/redirectHandler/ErrorOidcHandler.spec.ts
```diff
@@ -68,29 +70,35 @@ describe("ErrorOidcHandler", () => {
...
       expect(mySession.isLoggedIn).toEqual(false);
       expect(mySession.webId).toBeUndefined();
-      expect(mockCallback).toHaveBeenCalledWith("error", "unable_to_fetch");
+      expect(mockEmit).toHaveBeenCalledWith(
+        EVENTS.ERROR,
+        "error",
+        "unable_to_fetch"
+      );
     });
     it("calls the onError callback if given with just error parameters", async () => {
```
NSeydoux, line 89:
```suggestion
    });
    
    it("calls the onError callback if given with just error parameters if no description is provided", async () => {
```
Just to be extra-clear :)

packages/browser/src/login/oidc/redirectHandler/ErrorOidcHandler.ts
```diff
@@ -51,20 +53,14 @@ export class ErrorOidcHandler implements IRedirectHandler {
...
-    // The argument is ignored, but must be present to implement the interface
-    _onToken?: (newToken: string) => unknown,
-    onError?: (
-      error: string | null,
-      errorDescription?: string | null
-    ) => unknown
+    eventEmitter?: EventEmitter
   ): Promise<ISessionInfo & { fetch: typeof fetch }> {
-    if (typeof onError === "function") {
+    if (eventEmitter) {
```
NSeydoux, line 58:
```suggestion
    if (eventEmitter !== undefined) {
```
This won't change the behaviour here, but it's technically a stricter test, given we know that `eventEmitter` is either an object of type `EventEmitter` or `undefined`, and not some "falseish" value like an empty string for instance.

packages/browser/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -231,14 +235,15 @@ describe("TokenRefresher", () => {
...
         "mySession",
-        "some refresh token",
+        "some old refresh token",
         await mockKeyPair(),
-        callback
+        mockEmitter
       );
-      expect(callback).toHaveBeenCalled();
+
+      expect(mockEmit).toHaveBeenCalled();
```
NSeydoux, line 246: Here, I would test that the callback has been called with the appropriate arguments while we are at it, to verify that the actual refresh token value has been passed (it's writing this test for the node version that I realised I wasn't actually passing the refresh token value to the callback, hurray for test-driven development ^^)

packages/browser/src/Session.ts
dgbarclay: "this" doesn't work in this instance because it is outside of the Session class... is there a method to get the default session alongside the manually created one, or do we just need to pass the session into silentlyAuthenticate as a param??

packages/browser/src/Session.ts
```diff
@@ -120,16 +115,19 @@ export async function silentlyAuthenticate(
...
+        sessionId,
+        prompt: "none",
+        oidcIssuer: storedSessionInfo.issuer,
+        redirectUrl: storedSessionInfo.redirectUrl,
+        clientId: storedSessionInfo.clientAppId,
+        clientSecret: storedSessionInfo.clientAppSecret,
+        tokenType: storedSessionInfo.tokenType ?? "DPoP",
+        inIframe: options.inIframe,
+      },
+      this
```
NSeydoux, line 129: My bad, this should not be `this`, but instead a `Session` instance should be passed to the `silentlyAuthenticate` function, meaning that calls to `silentlyAuthenticate` will need to be passed `this` if called from a Session object.

PR #1577, opened 2021-08-04
Run audit fix

NSeydoux:
Caused by https://www.npmjs.com/advisories/1770


Vinnl (approved):
That's a whole lot of lockfile changes for a single vulnerable dependency :/

Issue #1285, opened 2021-04-19
Login failure with authn-node


jeff-zucker:
### Impacted package

solid-client-authn-node and possibly generate-oidc-token

### Bug description

Can not login.

### To Reproduce
1.  Get a token with generate-oidc-token
2.  Run this script  (which worked fine previously):
```javascript
import { Session } from "@inrupt/solid-client-authn-node";                      
const {getCredentials} = require('../../getCredentials');                       
                                                                                
async function login() {                                                        
  const sess = new Session();                                                   
  const creds = await getCredentials('jeff-zucker@inrupt.com');                 
  await sess.login(creds);                                                      
  return session;                                                               
}                                                                               
                                                                                
(async ()=>{                                                                    
  let s = await login();                                                        
  console.log(s)                                                                
})()                                                                            
```                                                                                
 
### Expected result

console log of the session object

### Actual result
```
(node:192246) UnhandledPromiseRejectionWarning: Error: Invalid refresh credentials: OPError: invalid_grant (Invalid refresh_token)
    at refreshAccess (/home/jeff/Dropbox/Web/solid/solid-node-client/node_modules/@inrupt/solid-client-authn-node/dist/login/oidc/oidcHandlers/RefreshTokenOidcHandler.js:49:15)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async RefreshTokenOidcHandler.handle (/home/jeff/Dropbox/Web/solid/solid-node-client/node_modules/@inrupt/solid-client-authn-node/dist/login/oidc/oidcHandlers/RefreshTokenOidcHandler.js:79:28)
    at async ClientAuthentication.login (/home/jeff/Dropbox/Web/solid/solid-node-client/node_modules/@inrupt/solid-client-authn-node/dist/ClientAuthentication.js:26:33)
    at async Session.login (/home/jeff/Dropbox/Web/solid/solid-node-client/node_modules/@inrupt/solid-client-authn-node/dist/Session.js:15:31)
    at async login (/home/jeff/Dropbox/Web/solid/solid-node-client/drafts/login-scan.js:7:3)
    at async /home/jeff/Dropbox/Web/solid/solid-node-client/drafts/login-scan.js:12:11
[...]

PR #1594, opened 2021-08-09
Properly handle client credentials

NSeydoux:
This PR fixes bug #1285. 

Going through the refresh token flow with a dynamically registered
client requires providing a client ID and secret, which are the
conditions to trigger the client credentials flow. However, the client
credential flow only works for __statically__ registered clients, which
leads to a bug.


packages/node/src/dependencies.ts
```diff
@@ -45,6 +54,51 @@ import ClientRegistrar from "./login/oidc/ClientRegistrar";
...
+        storageUtility,
+        new Redirector()
+      ),
+    ]),
+    issuerConfigFetcher,
+    clientRegistrar
+  );
+};
+
+export const buildRedirectHandler = (
```
Vinnl, line 81: ESLint warning.

Vinnl (approved):
Nice job on the tests :+1:

PR #1611, opened 2021-08-12
Allow 'any' in test files

NSeydoux:
This updates ESLint rules to allow for explicit any in test files,
because it's usually a legitimate use case.


Vinnl (approved):
Heh, that's a lot of exceptions - this is clearly long overdue.

Issue #1623, opened 2021-08-18
Proactive token refresh blocks graceful Node.js exits


rubensworks:
### Search terms you've used
refresh, token, exit, hanging

### Impacted package

Which packages do you think might be impacted by the bug ?

- [ ] solid-client-authn-browser
- [x] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

When using this lib in a Node.js application, the Node.js will not gracefully exit anymore after logging in.
This appears to be caused by the [timeouts that are set to handle token refreshing](https://github.com/inrupt/solid-client-authn-js/blob/main/packages/core/src/authenticatedFetch/fetchFactory.ts#L179-L201), which obviously makes sense.

I was wondering if it would be possible to work around this, by for example

1. disabling token refreshing, or
2. forcefully clearing the latest timeout handler.

### To Reproduce

I've created a [small repo with a bin script](https://github.com/rubensworks/solid-node-auth-helper.js) that demonstrates the problem.
After logging in, the Node process will hang.

Using a packages such as [`why-is-node-running`](https://www.npmjs.com/package/why-is-node-running), it can be observed that it is hanging because of the mentioned timeout handler.

### Environment
Please run

  npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers

in your project folder and paste the output here:

```

  System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 44.97 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.2.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 7.13.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
    Firefox: 90.0.2
    Safari: 13.1
  npmPackages:
    @inrupt/solid-client-authn-node: ^1.11.0 => 1.11.0
    @rubensworks/eslint-config: ^1.0.0 => 1.0.1
    @types/jest: ^26.0.0 => 26.0.24
    @typescript-eslint/eslint-plugin: ^4.2.0 => 4.29.2
    @typescript-eslint/parser: ^4.1.1 => 4.29.2
    coveralls: ^3.0.0 => 3.1.1
[...]

PR #1633, opened 2021-08-20
Clear timeouts on logout

NSeydoux:
The silent authentication running in the background prevents Node from
closing gracefully. This ensures that on logout, the latest timeout set
is cleared, so that no timeout is left hanging after the sessions have
been logged out.

This PR fixes bug #1623 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ so it should not have affected dependants, but it's technically more correct.
 - The `prompt=consent` parameter was missing when redirecting the user to the Solid
 Identity Provider authorization endpoint. This prevented working with the Community
 Solid Server Identity Provider.
+- Proactive refreshing of the token prevented NodeJS from shutting down gracefully.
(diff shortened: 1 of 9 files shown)


pmcb55 (approved):
Just a general comment - but I assume Node will still hang on shutdown if you don't logout first?
Perhaps that's acceptable, but maybe better if we can detect the Node shutdown event, and then also forceably kill the timer if needed to allow full shutdown (although you may have looked into that already).

NSeydoux:
I'll see if this is possible, but I'm not sure it will be, as the notion of Node shutdown is vague in this context: do you know if a dedicated event is sent when Node reaches the end of a script for instance ? I'm not sure the interpreter is able to make a distinction between a hanging `timeout` like the ones preventing it from shutting down and a legitimate async call that should not be interrupted.

pmcb55:
Nope, sorry - I don't know if a dedicated event is sent when Node reaches the end of a script or not. I guess it's one for a backlog investigation ticket...

PR #1701, opened 2021-09-24
Use createRemoteJWKSet for ID token validation

pavol-brunclik-compote:
This PR fixes bug: 
Where when validating ID token against public key from JWKS store at index [0].
In case another private key was used to sign ID token, validation failed.
Reintroduced usage of  createRemoteJWKSet for ID token validation + fixed tests using 'nock' library mocking requests to JWKS endpoint.


packages/core/src/login/oidc/redirectHandler/IRedirectHandler.ts
```diff
@@ -55,29 +56,39 @@ type WithStack = { stack: string };
...
     );
   }
   // The JWKS should only contain the current key for the issuer.
-  let jwk: JWK;
+  let jwk: JWK | undefined;
   try {
-    jwk = (await jwksResponse.json()).keys[0] as JWK;
+    const { keys } = await jwksResponse.json();
+    jwk = (keys as []).find(
+      (key: { kid: string }) => key.kid === idTokenHeader.kid
```
acoburn, line 78: There are some edge cases where this code will not retrieve the correct key.

https://datatracker.ietf.org/doc/html/rfc7517#section-4.5

> When "kid" values are used within a JWK Set, different keys within the JWK Set SHOULD use distinct "kid" values.

In other words, there may be cases where multiple keys in a JWKS will have the same identifier. It is certainly not a typical situation, but I have also encountered live services where this is the case.

Also important is the fact that the `kid` field is optional in a JWS:

https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.4

So there may be situations where the JWK has a `kid` defined and the JWS has no `kid`

The `alg` field _is_ required in a JWS

https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1

but it is optional in a JWK

https://datatracker.ietf.org/doc/html/rfc7517#section-4.4

All this is to say that this PR is an improvement for certain situations, but there are cases where it will miss the correct key (finding nothing) and cases where it will choose the wrong key (multiple JWKs with the same `kid`)

pavol-brunclik-compote: Hello, thank you for information and I fully agree the implementation is incomplete.
It looks to be much more better to use createRemoteJWKSet method from jose library for ID token validation, as already proposed in IRedirectHandler->fetchJWKS FIXME - this should cover much more cases:
- removed fetchJWKS method
- use createRemoteJWKSet directly in jwtVerify
- fixed / updated tests - mocking requests to JWKS endpoint using nock library
- kept the 'Not found' and 'malformed jwks' tests, though this is already covered by jose.

NSeydoux:
This is now fixed by https://github.com/inrupt/solid-client-authn-js/commit/bd14278e0cc34b223f1e7b26cade8fc9fb73e833

PR #1710, opened 2021-09-29
Setup end-to-end tests against dev-next

NSeydoux:
This sets up the end-to-end tests to run against dev-next, the pre-production ESS environment. These tests are currently failing, and fixing them will be done in a different PR.

In particular, these tests look for actual resources in the Pod which are not created at the moment,
causing the tests to fail. Creating the resources requires some extra setup on the
dev-next environment, which shall be done later.

### DIFF
--- a/packages/node/src/e2e/e2e-test.spec.ts
+++ b/packages/node/src/e2e/e2e-test.spec.ts
@@ -81,6 +81,7 @@ describe.each(serversUnderTest)(
     describe("Environment", () => {
       it("contains the expected environment variables", () => {
         expect(rootContainer).not.toBeUndefined();
+        expect(clientId).not.toBeUndefined();
         expect(clientSecret).not.toBeUndefined();
         expect(oidcIssuer).not.toBeUndefined();
       });
@@ -125,7 +126,6 @@ describe.each(serversUnderTest)(
           oidcIssuer,
         });
         const privateResourceUrl = rootContainer;
-        console.log(`Fetching ${privateResourceUrl}`);
         const response = await session.fetch(privateResourceUrl);
         expect(response.status).toEqual(200);
         await expect(response.text()).resolves.toContain("ldp:BasicContainer");
--- a/packages/node/src/e2e/e2e-test.spec.ts
+++ b/packages/node/src/e2e/e2e-test.spec.ts
@@ -36,157 +36,201 @@ config({
   silent: process.env.CI === "true",
 });
 
-// This first test just saves the trouble of looking for a library failure when
-// the environment wasn't properly set.
-describe("Environment", () => {
-  it("contains the expected environment variables", () => {
-    expect(process.env.E2E_TEST_CLIENT_ID).not.toBeUndefined();
-    expect(process.env.E2E_TEST_CLIENT_SECRET).not.toBeUndefined();
-    expect(process.env.E2E_TEST_IDP_URL).not.toBeUndefined();
-  });
-});
+type OidcIssuer = string;
+type ClientId = string;
+type ClientSecret = string;
+type Pod = string;
+type AuthDetails = [Pod, OidcIssuer, ClientId, ClientSecret];
+// Instructions for obtaining these credentials can be found here:
+// https://github.com/inrupt/solid-client-authn-js/blob/1a97ef79057941d8ac4dc328fff18333eaaeb5d1/packages/node/example/bootstrappedApp/README.md
+const serversUnderTest: AuthDetails[] = [
+  // pod.inrupt.com:
+  [
+    // Cumbersome workaround, but:
+    // Trim `https://` from the start of these URLs,
+    // so that GitHub Actions doesn't replace them with *** in the logs.
+    process.env.E2E_TEST_ESS_POD!.replace(/^https:\/\//, ""),
+    process.env.E2E_TEST_ESS_IDP_URL!.replace(/^https:\/\//, ""),
+    process.env.E2E_TEST_ESS_CLIENT_ID!,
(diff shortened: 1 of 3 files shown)


packages/node/src/e2e/e2e-test.spec.ts
```diff
@@ -36,157 +36,201 @@ config({
...
+  (rootContainerDisplayName, oidcIssuerDisplayName, clientId, clientSecret) => {
+    const rootContainer = "https://" + rootContainerDisplayName;
+    const oidcIssuer = "https://" + oidcIssuerDisplayName;
+
+    // This first test just saves the trouble of looking for a library failure when
+    // the environment wasn't properly set.
+    describe("Environment", () => {
+      it("contains the expected environment variables", () => {
+        expect(rootContainer).not.toBeUndefined();
+        expect(clientSecret).not.toBeUndefined();
```
matthieubosquet, line 84:
```suggestion
        expect(clientId).not.toBeUndefined();
        expect(clientSecret).not.toBeUndefined();
```

packages/node/src/e2e/e2e-test.spec.ts
```diff
@@ -36,157 +36,201 @@ config({
...
+
+      it("can fetch a private resource when logged in", async () => {
+        const session = new Session();
+        await session.login({
+          clientId,
+          clientSecret,
+          oidcIssuer,
+        });
+        const privateResourceUrl = rootContainer;
+        console.log(`Fetching ${privateResourceUrl}`);
```
matthieubosquet, line 128:
```suggestion
```
Probably no console.log in the tests?

matthieubosquet (approved):
We're missing secrets in the repo's settings: E2E_DEV_NEXT_USERNAME, E2E_DEV_NEXT_PASSWORD, E2E_TEST_ESS_CLIENT_ID, E2E_TEST_ESS_CLIENT_SECRET, E2E_TEST_ESS_IDP_URL.

Otherwise, I added two comments but pending those, it looks good to me.

packages/node/src/e2e/e2e-test.spec.ts
```diff
@@ -36,157 +36,201 @@ config({
...
-      clientSecret: process.env.E2E_TEST_CLIENT_SECRET,
-      oidcIssuer: process.env.E2E_TEST_IDP_URL,
+type OidcIssuer = string;
+type ClientId = string;
+type ClientSecret = string;
+type Pod = string;
+type AuthDetails = [Pod, OidcIssuer, ClientId, ClientSecret];
+// Instructions for obtaining these credentials can be found here:
+// https://github.com/inrupt/solid-client-authn-js/blob/1a97ef79057941d8ac4dc328fff18333eaaeb5d1/packages/node/example/bootstrappedApp/README.md
+const serversUnderTest: AuthDetails[] = [
```
matthieubosquet, line 46: Could we maybe use the same pattern as in @inrupt/solid-client-js here?

That is: using an array of objects instead of an array of arrays:
https://github.com/inrupt/solid-client-js/blob/09024b5a08f8ab00bd503a36e67cc1efb50e684b/src/e2e-browser/e2e.testcafe.ts#L76-L80

I think it's a bit cleaner than positional arguments, but I don't mind too much. Just a suggestion.

packages/node/src/e2e/e2e-test.spec.ts
```diff
@@ -36,157 +36,201 @@ config({
...
+  // pod.inrupt.com:
+  [
+    // Cumbersome workaround, but:
+    // Trim `https://` from the start of these URLs,
+    // so that GitHub Actions doesn't replace them with *** in the logs.
+    process.env.E2E_TEST_ESS_POD!.replace(/^https:\/\//, ""),
```
matthieubosquet, line 52: Can we extract those replace into a utility function?

```javascript
function trimHttps(x: string): string {
  return x.replace(/^https:\/\//, "");
}
```

packages/node/src/e2e/e2e-test.spec.ts
NSeydoux: unfortunately that's not possible, because these are two different things: in this case, we have to conform to Jest's `describe.each` signature, and an array of array is what is expected. Note that the file you're pointing to contains browser tests, which run with TestCafé, and not Jest, which is why we use a different syntax.

packages/node/src/e2e/e2e-test.spec.ts
NSeydoux: Probably, but do you think that would make the code more readable ?

packages/node/src/e2e/e2e-test.spec.ts
matthieubosquet: Oh, sad. Well, c'est la vie!

packages/node/src/e2e/e2e-test.spec.ts
matthieubosquet: I think it would and more importantly, I opt for DRY by default; `.replace(/^https:\/\//, "")` is arguably a case of code duplication.

However, it's pretty easy to read as is and the "repeat" part only holds because of the generic regex... So frankly I don't mind. ^^,

(I'm much more bothered by all of our CI being a big case of code duplication (and by our, I don't mean it's endemic))

I'll mark as resolved, we have bigger fishes to fry ;)

NSeydoux: Yeah, there IS a lot of duplication across our CI, but I'm not sure how it could be done otherwise... Another thing to consider on the long run :)

Issue #1708, opened 2021-09-28
Unable to pass custom Headers


rubensworks:
### Search terms you've used
headers, fetch init

### Impacted package

Which packages do you think might be impacted by the bug ?

- [ ] solid-client-authn-browser
- [x] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

This library seems assume that the headers field within the init object of a fetch always uses the `Record<string, string>` form, while this can also be a `string[][]` or `Headers`.

Within Comunica, we're consistently passing around `Headers` objects. But when it reaches this auth library, the headers become mangled, which causes several issues (such as being unable to set the content type on PUT/PATCH requests).

Related to #176, #96

### To Reproduce

`session.fetch('...', { headers: new Headers({ 'Content-Type': 'text/turtle' }) })`

### Expected result
Headers are delegated properly

### Actual result
Headers are ignored

### Environment
Please run

  npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers

in your project folder and paste the output here:

```

  System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 23.36 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.2.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 7.13.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 94.0.4606.61
    Firefox: 91.0.2
    Safari: 13.1
  npmGlobalPackages:
    @comunica/actor-init-sparql-file: 1.14.0
    @ldflex/comunica: 3.4.2
    chrome-headless-render-pdf: 1.9.0
    clinic: 8.0.1
    componentsjs-generator: 2.6.0
    generator-comunica: 1.0.0
    http-server: 0.12.3
    ldbc-snb-enhancer: 1.0.2
    n: 6.5.1
    npm-cli-adduser: 1.1.4
    npm: 7.13.0
    rdf-dataset-fragmenter: 1.2.0
    rdf-dereference: 1.5.0
    renovate: 23.20.2
    verdaccio: 3.11.4
    why-is-node-running: 2.2.0
    yo: 3.1.1

```

## Additional information

[...]

PR #1716, opened 2021-10-04
Prevent headers from being mangled

NSeydoux:
This PR fixes bug #1708 .

Instead of using a deconstruction assignment, which doesn't work on all the forms of Headers, this manipulates the fetch headers through the Headers interface, which preserves headers passed in all forms.


packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -50,8 +50,17 @@ import { KeyObject } from "crypto";
...
 import { configToIssuerMetadata } from "../IssuerConfigFetcher";
 
+// Using the cross-fetch polyfill should enforce Headers to be present in the Node environment
+// However, it does not behave as expected, which is why the global.Headers value
+// is manually overwritten.[]
+// import "cross-fetch/polyfill";
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+global.Headers = NodeHeaders;
+
```
matthieubosquet, line 63: This was bugging me a bit.
I think we probably don't need that.
What might be enough is to import both Headers and fetch in the core package fetchFactory: packages/core/src/authenticatedFetch/fetchFactory.ts
That is: `import { Headers, fetch } from "cross-fetch";` (instead of `import { fetch }...`).

The TypeScript compiler gets confused by the first line in "cross-fetch" exports `/// <reference lib="dom" />`:
https://github.com/lquixada/cross-fetch/blob/dcb9279a50f911c4a6e031de44397305a7c948a6/index.d.ts#L1-L14

Which I think makes it believe the whole lib.dom is available ^^,

packages/node/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -50,8 +50,17 @@ import { KeyObject } from "crypto";
 import { fetch as globalFetch } from "cross-fetch";
 
 import { EventEmitter } from "events";
+import { Headers as NodeHeaders } from "node-fetch";
```
matthieubosquet, line 53: We don't need the extra "node-fetch" dependency.

```suggestion
import { fetch as globalFetch } from "cross-fetch";

import { EventEmitter } from "events";
import { Headers as NodeHeaders } from "node-fetch";
```
```suggestion
import { fetch as globalFetch, Headers as NodeHeaders } from "cross-fetch";

import { EventEmitter } from "events";
```

matthieubosquet, line 53: And actually, if I'm right, we might not need so much node specific code.
(I think the correct fix here is not importing Headers at all since it's only there to override global.)

matthieubosquet (approved):
LGTM! :)

PR #1793, opened 2021-11-19
Fix authn when using client identifiers

### LINKED ISSUE #1767: Node and browser versions of the library are out of sync
The Node version gives

```
client_secret_basic client authentication method requires a client_secret
```

where the browser version succeeds.

Details at https://github.com/solid/community-server/issues/838#issuecomment-957273062

NSeydoux:
This fixes #1767.

When using a Client Identifier as defined in Solid-OIDC, a client does not have a secret. Based on https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1, it looks like it is acceptable to only provide a client id, but no secret, in the authentication information stored in the body of the request to the token endpoint.

However, this is currently running into an issue with the underlying OIDC library. A conversation is happening at https://github.com/panva/node-openid-client/discussions/430. In the meantime, an arbitrary secret is given to the library, which works around the issue and is ignored by the server.

### DIFF
--- a/packages/node/src/constant.ts
+++ b/packages/node/src/constant.ts
@@ -22,5 +22,3 @@
 import { SOLID_CLIENT_AUTHN_KEY_PREFIX } from "@inrupt/solid-client-authn-core";
 
 export const KEY_REGISTERED_SESSIONS = `${SOLID_CLIENT_AUTHN_KEY_PREFIX}registeredSessions`;
-// Currently required, pending a response to https://github.com/panva/node-openid-client/discussions/430.
-export const TMP_ARBITRARY_SECRET = "some arbitrary secret";
--- a/packages/node/src/constant.ts
+++ b/packages/node/src/constant.ts
@@ -22,3 +22,5 @@
 import { SOLID_CLIENT_AUTHN_KEY_PREFIX } from "@inrupt/solid-client-authn-core";
 
 export const KEY_REGISTERED_SESSIONS = `${SOLID_CLIENT_AUTHN_KEY_PREFIX}registeredSessions`;
+// Currently required, pending a response to https://github.com/panva/node-openid-client/discussions/430.
+export const TMP_ARBITRARY_SECRET = "some arbitrary secret";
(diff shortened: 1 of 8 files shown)


packages/node/src/constant.ts
```diff
@@ -22,3 +22,5 @@
 import { SOLID_CLIENT_AUTHN_KEY_PREFIX } from "@inrupt/solid-client-authn-core";
 
 export const KEY_REGISTERED_SESSIONS = `${SOLID_CLIENT_AUTHN_KEY_PREFIX}registeredSessions`;
+// Currently required, pending a response to https://github.com/panva/node-openid-client/discussions/430.
+export const TMP_ARBITRARY_SECRET = "some arbitrary secret";
```
ThisIsMissEm, line 26: Perhaps make this value: 
```suggestion
export const TMP_ARBITRARY_SECRET = "arbitrary-secret-see-issue-XXX";
```

Or something such that if you see that client secret crop up, you know what it means. (it's self-descriptive)

packages/node/src/login/oidc/refresh/TokenRefresher.spec.ts
```diff
@@ -206,6 +203,12 @@ describe("TokenRefresher", () => {
 
   it("throws if a refresh token isn't provided", async () => {
     setupDefaultOidcClientMock();
+    const mockedModule = jest.requireMock("../ClientRegistrar") as {
+      negotiateClientSigningAlg: typeof negotiateClientSigningAlg;
+    };
+    mockedModule.negotiateClientSigningAlg = jest
+      .fn(negotiateClientSigningAlg)
+      .mockReturnValue("ES256");
     const mockedStorage = mockRefresherDefaultStorageUtility();
```
ThisIsMissEm, line 212: A small stylistic thing: I'd normally recommend a new line after a multi-line block, specifically when chaining method calls.

ThisIsMissEm (approved):
I'm not entirely familiar with this code, but overall looks good to me.

packages/node/src/constant.ts
NSeydoux: Fortunately I was an idiot and didn't read the documentation well enough, and this bogus client secret is no longer necessary.

RubenVerborgh:
Thank you!

Issue #1657, opened 2021-09-01
Question: What is the best way to use this library within a Web Worker?


rubensworks:
I've been trying to use this library to fetch resources within the browser via a Web Worker, without much success.
I will describe below what I have tried up until now, and was wondering if someone could give me some pointers on what would be the best way to go forward.

**My intended workflow:**

1. Authenticate the user within the browser itself.
2. Serialize the session and send it to the Worker within a message.
3. Deserialize the session within the Worker.
4. Do `fetch` requests using the authenticated session.

Step 1 works without problems, but I'm unsure how to handle steps 2-3.
So far, I tried two approaches:

### Serialization attempt 1

Main browser process:
```javascript
// session is authenticated at this stage
const sessionInfoRaw = session.clientAuthentication.getSessionInfo(session.info.sessionId);
sendToWorker(sessionInfoRaw);
```

(I am aware that `session.clientAuthentication` is private, so this is probably not a good solution)

Worker:
```javascript
const session = new Session({ sessionInfo: sessionInfoRaw });
await session.login({});
```

=> Throws a `window is not defined` error at `BrowserStorage.js:5` (possibly caused by #308).

I can observe that the session already contains the correct `webId` value.

### Serialization attempt 2

Main browser process:
```javascript
// session is authenticated at this stage
const sessionInfoRaw = session.clientAuthentication.getSessionInfo(session.info.sessionId);
sendToWorker({
  refreshToken: sessionInfoRaw.refreshToken,
  clientId: sessionInfoRaw.clientAppId,
  clientSecret: sessionInfoRaw.clientAppSecret,
  oidcIssuer: sessionInfoRaw.issuer,
});
```

Worker:
```javascript
const session = new Session();
await session.login(sessionInfoRaw);
```

=> Throws a `window is not defined` error at `BrowserStorage.js:5` (possibly caused by #308).

In this case, the session contains an undefined `webId` value.

NSeydoux:
I'm afraid the main issue is that I don't believe the session is serializable: sensitive information is kept in-memory in the `fetch` closure, which isn't readable on purpose due to security concerns. This means that the Web Worker would have to do the full authentication process in order to get an authenticated session, which involves a redirection in most cases, so I don't think this could work. 

[...]

PR #1802, opened 2021-11-29
Expose authenticateHeaders method and helpers for Web workers in browsers

rubensworks:
# Checklist

-----

As discussed in #1657, this PR adds two main features, which are fully tested and documented (as JSDoc).
(If there's another place that requires documentation writing around this, please let me know)

### 1. `Session#authenticateHeaders`

This is a new method that exists in both Node.js and browsers, and allows an authenticated headers object to be created based on a given request input.

This method enables authenticated request execution to be delegated to other services, such as Web workers.

This approach does not leak any of the internal secrets of the session object, as it just signs the request, within sharing the internal keys.

Example usage:
```javascript
const authHeaders = session.authenticateHeaders(
  'https://example.org',
  'GET',
  new Headers({ accept: 'text/turtle' }),
);
```

### 2. Web worker helpers

To simplify the usage of this library within Web workers, I've added some helper classes that can be used to handle the communication channel between the main window and a worker.

This communication channel exposes a fetch-like method to the worker, and follows the following flow:

* Worker wants to do an authenticated request.
* Worker sends request information to the main Window.
* Window constructs authenticated headers for the request information.
* Window sends back authenticated headers to the Worker.
* Worker performs request using the authenticated headers.

I've also adapted the [single bundle example](https://github.com/rubensworks/solid-client-authn-js/tree/feature/web-workers/packages/browser/examples/single/bundle) so that authenticated fetch requests can also be done via workers, as an example.

Furthermore, I've tested this approach within the (request-heavy) Comunica query engine, and everything seems to be working (with good performance) there as well: https://data.rubensworks.net/tmp/comunica-solid/

Example usage (Window):
```javascript
   const session = getDefaultSession();
   const worker = new Worker(...);
   const windowToWorkerHandler = new WindowToWorkerHandler(this, worker, session);

   worker.onmessage = async (message) => {
      if (windowToWorkerHandler.onmessage(message)) {
        // This means that the message was taken care of by the handler
      } else {
        // Optionally, take care of any other custom messages that your worker may send.
      }
    };
```

Example usage (Worker):
```javascript
   const workerToWindowHandler = new WorkerToWindowHandler(self);
   self.onmessage = (message => {
    if (workerToWindowHandler.onmessage(message)) {
      // This means that the message was taken care of by the handler
    } else {
      // Optionally, take care of any other custom messages that your worker may receive.
    }


   // Use the authenticated fetch function
  const authFetch = workerToWindowHandler.buildAuthenticatedFetch();
  await authFetch('https://example.org/');
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### New features
+
+#### node and browser
+
+- Expose `Session#authenticateHeaders` method that constructs the necessary headers to perform a given request using
+this session's authentication.
+
+#### browser
+
+- The `WindowToWorkerHandler` and `WorkerToWindowHandler` helpers can be used to setup a communication channel between
+the main window and a Web worker, so that the worker can perform authenticated requests, without any secrets having to
+be shared with this worker. The actual fetch requests still happen within the Web worker for optimal performance.
+
 ### Bugfixes
 
 #### node and browser
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -39,6 +39,7 @@ export default function App() {
   const [issuer, setIssuer] = useState("https://broker.pod.inrupt.com/");
   const [resource, setResource] = useState(webId);
   const [data, setData] = useState(null);
+  const worker = new Worker(new URL("./worker.js", import.meta.url));
 
   // The useEffect hook is executed on page load, and in particular when the user
   // is redirected to the page after logging in the identity provider.
@@ -87,6 +88,13 @@ export default function App() {
       .then(setData);
   };
 
+  const handleFetchWorker = (e) => {
+    e.preventDefault();
+
+    worker.postMessage({ resource });
+    worker.onmessage = ({ data: { text } }) => setData(text);
+  };
+
   return (
     <div>
       <main>
@@ -115,6 +123,9 @@ export default function App() {
             }}
           />
           <button onClick={(e) => handleFetch(e)}>Fetch</button>
+          <button onClick={(e) => handleFetchWorker(e)}>
+            Fetch with Web Worker
+          </button>
         </div>
         <pre>{data}</pre>
       </main>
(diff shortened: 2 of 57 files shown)


NSeydoux:
Thanks for submitting this, I'll have a look into it soon!

NSeydoux:
Hi Ruben, sorry it took so long to respond to your PR !

Thanks for the contribution: it is interesting, and enables additional use cases. However, it is incompatible with our security model, as exposing this a method to sign the headers from the function facilitates token extraction and arbitrary usage of the DPoP key. That means that the code proposed in this PR cannot be merged in the main codebase.

That said, the use case you want to achieve is legitimate, and there have been other cases where only having the fetch authenticated, and having to inject it everywhere, has been painful. Our feeling is that if your use-case is different from the "traditional" approach, where the app logs the user in to simply read and write data to their Pod, it is possible that `@inrupt/solid-client-authn-browser` is actually simply getting in your way, as it is designed to facilitate that mainstream use case as much as possible.

Do you think it would be reasonable for you to drop down one level, and use a library which provide less wrapping around the OIDC interaction, and in particular does not handle the access tokens/keys for you, but instead expects you to manage them ? This way, you would not have to work around the features of `@inrupt/solid-client-authn-browser`.

The thing is, the OIDC library we are currently using is `oidc-client`, and it has recently been deprecated. We're maintaining our own fork, `@inrupt/oidc-client` to make sure there are no security issues, but that's not the greatest state a codebase could be in for you to start using. Our plan is to actually move away from it in the future, and to implement our own low-level OIDC library. It would expose primitives that would be used in `@inrupt/solid-client-authn-browser`, so the migration would be transparent for users of the high-level library, but what we can do is make sure that the design of the low-level API works with your use case. This way, users who just want to make authenticated requests to a Solid pod can still use the high-level library, but users who are more comfortable with OIDC and have a more advanced use case will be able to have more control on the OIDC flow, and will be able to eventually manage the tokens themselves, which would vastly facilitate what you want to do for instance.
Is that a plan that makes sense to you ? In the meantime, you can use your modified fork of solid-client-authn-browser.

rubensworks:
@NSeydoux Thanks for your reply!

> Do you think it would be reasonable for you to drop down one level, and use a library which provide less wrapping around the OIDC interaction
> Is that a plan that makes sense to you ?

Certainly makes sense to me!

I'll follow up with @pmcb55 to see how we want to move forward with publishing this functionality in my fork.
Once this new low-level lib is ready, I'm definitely open to rebuilding this functionality over that.

Closing this PR in the meantime. (I'll place a comment on this thread once this fork has been released)

rubensworks:
For reference, my fork is available here: https://github.com/rubensworks/solid-client-authn-js (has been published to npm as well)

PR #1805, opened 2021-12-03
Chore/remove arborist dependency

NSeydoux:
`lerna-audit` introduced a vulnerability through one of its dependencies, which required a fixed dependency resolution. This is no longer the case, so the fix can be removed.

Edit: There were a couple of additional issues to resolve, including a moedreate security vulnerability, and an issue with an older version of jest that was resolved by upgrading.


NSeydoux:
@ThisIsMissEm the typo in the commit message is now fixed. 

Rebasing on main introduced some issues, making it necessary to update the package locks, ending up with 0a7a5d9 for some reason because the CI kept failing.

ThisIsMissEm:
Screenshot 2021-12-07 at 17 19 52

@NSeydoux I think the commits still need to be rewritten, or you've forgotten to force push? Or are you actually intending to squash merge this (which I personally think would be a bad idea in this case)

NSeydoux:
They still need to be rewritten, I should have done it before asking for reviews sorry about that. And agreed, I intended to rebase merge these since each individual commit makes sense in the history

ThisIsMissEm (approved):

PR #1816, opened 2021-12-06
No longer store the ID token in local storage

NSeydoux:
The ID token becoming a part of the authentication flow and being exchangeable for an Access token, it is now a critical piece of information and should be handled as such. It means, among other things, that it should not be stored in local storage. A future PR will keep it in the fetch closure, like the access token.

### DIFF
--- a/packages/browser/src/ClientAuthentication.ts
+++ b/packages/browser/src/ClientAuthentication.ts
@@ -33,10 +33,8 @@ import {
   IIssuerConfigFetcher,
   ISessionInternalInfo,
   ILoginOptions,
-  fetchJwks,
 } from "@inrupt/solid-client-authn-core";
 import { removeOidcQueryParam } from "@inrupt/oidc-client-ext";
-import { jwtVerify, importJWK } from "jose";
 import { EventEmitter } from "events";
 import { KEY_CURRENT_SESSION } from "./constant";
 
@@ -110,6 +108,10 @@ export default class ClientAuthentication {
     return this.sessionInfoManager.getAll();
   };
 
+  // Collects session information from storage, and returns them. Returns null
+  // if the expected information cannot be found.
+  // Note that the ID token is not stored, which means the session information
+  // cannot be validated at this point.
   validateCurrentSession = async (): Promise<
     (ISessionInfo & ISessionInternalInfo) | null
   > => {
@@ -118,33 +120,14 @@ export default class ClientAuthentication {
       return null;
     }
     const sessionInfo = await this.sessionInfoManager.get(currentSessionId);
-    // Several types of session data are required in order to validate that the ID
-    // token in storage hasn't been tampered with, and has actually been issued
-    // by the issuer present in storage.
     if (
       sessionInfo === undefined ||
-      sessionInfo.idToken === undefined ||
       sessionInfo.clientAppId === undefined ||
       sessionInfo.issuer === undefined
     ) {
       return null;
     }
-    const issuerConfig = await this.issuerConfigFetcher.fetchConfig(
-      sessionInfo.issuer
-    );
-
-    try {
-      const jwk = await fetchJwks(issuerConfig.jwksUri, issuerConfig.issuer);
-      await jwtVerify(sessionInfo.idToken, await importJWK(jwk), {
-        audience: sessionInfo.clientAppId,
-        issuer: issuerConfig.issuer,
-      });
-      return sessionInfo;
-    } catch (e) {
-      // The jwt verification function throws on invalid token.
-      // The error is swallowed, and `null` is eventually returned.
-    }
-    return null;
+    return sessionInfo;
   };
 
   handleIncomingRedirect = async (
--- a/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
+++ b/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
@@ -215,10 +215,6 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
     await this.storageUtility.setForUser(
       storedSessionId,
       {
-        // TODO: We need a PR to oidc-client-js to add parsing of the
-        //  refresh_token from the redirect URL.
-        refreshToken:
-          "<Refresh token that *is* coming back in the redirect URL is not yet being parsed and provided by oidc-client-js in it's response object>",
         webId: tokens.webId,
         isLoggedIn: "true",
       },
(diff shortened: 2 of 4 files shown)


packages/browser/src/ClientAuthentication.ts
```diff
@@ -118,33 +120,14 @@ export default class ClientAuthentication {
       return null;
     }
     const sessionInfo = await this.sessionInfoManager.get(currentSessionId);
```
ThisIsMissEm, line 122: Let's also make sure that any sessionInfo that would've contained an ID Token strips it here, such that `sessionInfo` absolutely does not have an ID Token in it.

packages/core/src/storage/StorageUtility.ts
```diff
@@ -116,9 +114,6 @@ export async function saveSessionInfoToStorage(
   if (refreshToken !== undefined) {
     await storageUtility.setForUser(sessionId, { refreshToken }, { secure });
   }
-  if (idToken !== undefined) {
-    await storageUtility.setForUser(sessionId, { idToken }, { secure });
```
ThisIsMissEm, line 120: I think we may also need to explicitly clear the key for the idToken here too, such that we don't accidentally leave dead data in localstorage

ThisIsMissEm (changes requested):
I think we may need to explicitly clear the previous values and harden this a bit.

packages/core/src/storage/StorageUtility.ts
NSeydoux: Agree on clearing local storage, but I think it could happen in a different place. Session data is currently cleared when the user logs out, but that means that it hangs around if the user just leaves without properly logging out. Note that the session ID is a randomly generated uuid, which means that logging in without logging out multiple times leaves multiple instances of session data hanging around, which isn't great. It's not really an issue on the server side with the Node codebase, because we want users to be able to be logged back in silently, but that's not possible on the client-side because it's unsafe to persist the refresh token in local storage.

All this to say, what we could do instead is, for `solid-client-authn-browser`, on login, go through local storage and clear all session data associated to the same webid. This way, we're sure that the local storage doesn't grow infinitely, and that any stale data is removed from there. If you don't mind, I'd rather do this in a separate PR, because it's pretty much its own thing. Does that sound good to you ?

packages/browser/src/ClientAuthentication.ts
NSeydoux: Good point, 73b077e08cd762496aafbb937b50572ecb76ae8d removes the `idToken` property from the session info type, and enforces that it isn't  looked for in storage.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -215,10 +215,6 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
     await this.storageUtility.setForUser(
       storedSessionId,
       {
-        // TODO: We need a PR to oidc-client-js to add parsing of the
-        //  refresh_token from the redirect URL.
-        refreshToken:
-          "<Refresh token that *is* coming back in the redirect URL is not yet being parsed and provided by oidc-client-js in it's response object>",
```
ThisIsMissEm, line 221: We're no longer storing this long-term right?

packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -197,7 +192,6 @@ export class SessionInfoManager implements ISessionInfoManager {
       webId,
       isLoggedIn: isLoggedIn === "true",
       redirectUrl,
-      idToken,
       refreshToken,
```
ThisIsMissEm, line 195: Should we be validating that the refreshToken is a JWT here? (I assume it is)

packages/core/src/storage/StorageUtility.ts
ThisIsMissEm: I think I need more time to better understand this, but I think we could also store a `session:${id}:expiresAfter` date/time, and then flush (clear) all sessions where expiresAfter is `< Date.now()`

I would also like to see this behaviour very clearly documented, if it is not already, as this aspect of the SDK is certainly where we can get really bad privacy bugs forming. e.g., when you sign-in, should there be a "remember me" checkbox, and if it's not checked then we don't store stuff past the current browsing session — single sessionId mode)

ThisIsMissEm (approved):
Two concerns, but I think we're good to merge this for now.

packages/core/src/storage/StorageUtility.ts
NSeydoux: +1 on the documenting this, I was just thinking about adding a section in the `ARCHITECTURE.md` about exactly what gets stored where.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Exactly, the refresh token is only kept in the fetch function's closure in the browser module, because since it allows getting an access token, it is as critical.

packages/browser/src/sessionInfo/SessionInfoManager.ts
NSeydoux: It isn't necessarily: the refresh token, like the access token, is opaque to the client, it's an implementation detail of the OIDC provider.

packages/core/src/storage/StorageUtility.ts
NSeydoux: And also, expiring the session is a bit challenging: what I would be afraid of is to end up in a situation where a session which is still active gets logged out. The session being "active" means it has a valid access token basically, which happens internally in the fetch closure. What could be done is to inject a callback all the way into the fetch so that they have a reference to the storage and keep the expiry date current. I'll try to design this properly so that we can talk about it :)

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
ThisIsMissEm: SessionManager still persists the refresh token: https://github.com/inrupt/solid-client-authn-js/blob/73b077e08cd762496aafbb937b50572ecb76ae8d/packages/browser/src/sessionInfo/SessionInfoManager.ts#L160

NSeydoux: Good catch. I confirm it's not stored, it's only injected into the fetch closure there: https://github.com/inrupt/solid-client-authn-js/blob/1dd3ea0d50d0d3f8d1787ce8e826d179e319f472/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts#L210. But we still look it up in storage, which is wrong, I'll remove this as part of the PR too.

PR #1840, opened 2021-12-09
Remove node fetch dependency

NSeydoux:
Our packages already depend on cross-fetch, which in turns depends on node-fetch. The dependency on node-fetch was only introduced in the tests because cross-fetch was mocked and it created unintended consequences that were not fully understood initially. Now, mocking of cross-fetch is done properly, which makes depending on node-fetch unecessary. cross-fetch comes with its own type definitions too.

Note that this supersedes https://github.com/inrupt/solid-client-authn-js/pull/1776, and it is a mystery to me why the latter has issues with `lerna bootstrap` while the current PR is fine with it.


packages/core/src/login/oidc/redirectHandler/IRedirectHandler.spec.ts
```diff
@@ -21,10 +21,15 @@
...
+import { Response as NodeResponse } from "cross-fetch";
 import { getWebidFromTokenPayload } from "./IRedirectHandler";
 
-jest.mock("cross-fetch");
+jest.mock("cross-fetch", () => {
+  const crossFetchModule = jest.requireActual("cross-fetch") as any;
+  crossFetchModule.default = jest.fn();
+  crossFetchModule.fetch = jest.fn();
+  return crossFetchModule;
+});
```
matthieubosquet, line 32: There is something weird in our jest config that requires us to import jest:
```javascript
import { jest, it, describe, expect } from "@jest/globals";
```

That messes with the types. Which means that the actual type of requireActual is not seen:
```typescript
    function requireActual<TModule extends {} = any>(moduleName: string): TModule;
```

If that was not the case, we should be able to mock the module as such, which I find to be nicer:
```suggestion
import type * as CrossFetch from "cross-fetch";

jest.mock("cross-fetch", () => {
  return {
    ...jest.requireActual("cross-fetch"),
    default: jest.fn(),
    fetch: jest.fn(),
  } as typeof CrossFetch;
});
```

matthieubosquet (approved):
It made me realise there is something a bit dodgy about our jest configuration (see comment), but the PR itself looks good to me.

packages/core/src/login/oidc/redirectHandler/IRedirectHandler.spec.ts
ajacksified: I would note that I think only solid-client and solid-client-auth import test directly. All of our other libraries use the implicit import.

matthieubosquet: Oh by the way, this actually works (one superfluous cast and of course, the dodgy impor { jest... } from "cross-fetch"; non standing):
```suggestion
import type * as CrossFetch from "cross-fetch";

jest.mock("cross-fetch", () => {
  return {
    ...jest.requireActual("cross-fetch") as typeof CrossFetch,
    default: jest.fn(),
    fetch: jest.fn(),
  } as typeof CrossFetch;
});
```

NSeydoux: > I would note that I think only solid-client and solid-client-auth import test directly. All of our other libraries use the implicit import.

I'm not sure the latest jest versions still support the implicit import, but I would need to double check to be sure.

NSeydoux: @matthieubosquet I consistently applied the change in https://github.com/inrupt/solid-client-authn-js/pull/1840/commits/5298208ef14105c03718864a84c7f3bc4a34a10a

PR #1856, opened 2021-12-15
Convert e2e tests to node-tap

NSeydoux:
The Jest runner has some subtles differences with the Node environment. It is fine for our unit tests, but we want our end-to-end tests to run in an environment as close as possible to the actual environment in which the code will actually run.

To this end, this experiments with node-tap as a test framework. There is no runner, the tests just run as regular node scripts.

There are a couple of points to consider before merging: 
- The test output isn't ideal, but it looks like it is pretty flexible. The current output gives us what we need to debug, but it could be nicer, e.g. the assertions comments could be indented under the test they relate to.
- There is native support for Typescript if `ts-node` is added to the dependencies. I haven't added it just yet though, because it is also possible to simply use `tsc` to transpile the tests and Node to run them, the same way we handle the rest of the TS codebase. I'm still pondering it, but I'm inclined to say that the overhead of additional dependencies outwheights the benefits of tap directly picking up the TS code. Note that emitting the e2e tests with the default TS compiler adds a directory in the `dist` folder, which is why https://github.com/inrupt/solid-client-authn-js/pull/1856/commits/8bd7c32921dc47518578033eb11b10b23ae1cf89 was necessary. It means the e2e tests are shipped with the code, which isn't too bad for node code (where size doesn't matter as much as it is in the browser), and which was already the case when the e2e tests were located under `src` anyway.


matthieubosquet (approved):
Seems ok to me, even though the e2e run fails on NSS:
https://github.com/inrupt/solid-client-authn-js/runs/4546669029?check_suite_focus=true#step:9:1366

Might be a matter of re-running it...

NSeydoux:
Yeah the browser-based tests are super flaky, in particular on NSS, but since we are looking to replace these soon I'm not even trying to fix it.

packages/node/e2e/authenticated.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+type AuthDetails = {
+  pod: string;
+  oidcIssuer: string;
+  clientId: string;
+  clientSecret: string;
+};
+// Instructions for obtaining these credentials can be found here:
+// https://github.com/inrupt/solid-client-authn-js/blob/1a97ef79057941d8ac4dc328fff18333eaaeb5d1/packages/node/example/bootstrappedApp/README.md
+const serversUnderTest: AuthDetails[] = [
```
ThisIsMissEm, line 57: _Makes note to also change this to a test matrix_

packages/node/e2e/authenticated.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      oidcIssuer,
+    });
+  });
+
+  afterEach(async () => {
+    await session.logout();
+  });
+
+  test("properly sets up session information", async () => {
+    strictSame(
```
ThisIsMissEm, line 115: There is also a mocha like DSL, which might make these easier? https://node-tap.org/docs/api/mochalike/

NSeydoux: I don't really have an opinion on which one reads better, as in any case it's different from the `expect(...).toStrictEqual(...)` from Jest. I'm happy to switch to the Mocha-like DSL if you think it improves readability.

ThisIsMissEm: I think it might be good, just so there's readable consistency between these tests and our other tests — less to learn, less context switching. Thoughts @matthieubosquet & @ajacksified ?

ajacksified: I personally prefer the mocha-like syntax, although it's going to be a context switch either way. I could go in either direction.

matthieubosquet: I prefer consistency and the mocha like syntax.

NSeydoux: That's settled then !

NSeydoux:
Overall, I'm not really satisfied with node-tap, and we agreed to not migrate the test to this framework.

PR #1867, opened 2021-12-21
Set up AVA for end-to-end tests

NSeydoux:
This is an experiment to see what our end-to-end tests would look like written with AVA. The main reason we would consider changing test frameworks for the end-to-end test is due to the discrepency between the Jest Node environment and the actual Node environment, leading to some subtle and painful-to-debug issues.

A couple of obesrvations:
- If running the tests in a loop, the test titles must change at each iteration. This will no longer be an issue when the environment will be set by the test matrix.
- AVA encourages to break tests into smaller files rather than in subtests with "describe" statements.
- The tests are recompiled when ran, to avoid running stale tests.
- TS support is quite good: the transliped code runs the tests, but errors are reported in the TS source.


matthieubosquet:
I have a vague idea that AVA is better for integration tests and that Jest's node environment has slight differences with node.

But do we have a more specific list of differences between AVA and Jest that would make the decision to move to AVA for e2e tests easier?

NSeydoux:
Not a list of differences between AVA and Jest, but here is a list of bugs related to Jest being slightly misaligned with the regular Node runtime: 
- https://github.com/facebook/jest/issues/11629
- https://github.com/facebook/jest/issues/7780
- https://github.com/facebook/jest/issues/2549

The latter contains the core of the issue really.

To summarize:
 - The main motivation to move from Jest to AVA would be to run in a "vanilla" Node runtime to avoid missing or hitting such issues. 
 - As far as end-to-end tests are concerned, AVA and Jest have all the required features, so there's no reason to use one over the other at this level.
 - The tests are as readable using one framework or the other
 - Our unit tests are written in Jest and would remain so in the foreseeable future, so moving the end-to-end tests to AVA introduces a new dependency and a different test framework, which is not a big deal but a (small) price to pay still.

matthieubosquet:
The issue with Jest globals seems pretty bad but I'm not sure how much it affects us.

I would tend to lean towards keeping the number of dependencies to a minimum and migrate if and when we hit a serious bottleneck with Jest (I'm not best placed to determine whether we have or not).

NSeydoux:
Migrating the end-to-end tests to AVA should work, but it is low-priority. Closing this for now, but keeping it around.

PR #1909, opened 2022-01-18
Fix outdated browser example lockfiles

ThisIsMissEm:
See the individual commits for details, this bumps a bunch of outdated dependencies in the examples, upgrades the lockfiles to v2, and fixes a bug in the demoSingleApp's functionality.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11262,12 +11262,12 @@
       "dev": true
     },
     "node_modules/marked": {
-      "version": "3.0.8",
-      "resolved": "https://registry.npmjs.org/marked/-/marked-3.0.8.tgz",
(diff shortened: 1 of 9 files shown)


matthieubosquet (approved):
LGTM

NSeydoux:
Do you think it'd be worth enabling dependabot for these while we're at it ?

ThisIsMissEm:
> Do you think it'd be worth enabling dependabot for these while we're at it ?
> 
> PS: some of the browser tests failing is fine, it comes from testcafé's falkiness, I'll move them to playwright soon to fix this)

Maybe, but I'm not sure the effect that'd have on our workload, so I'd perhaps be hesitant to enable it, given these are react apps and such that more frequently probably receive updates.

ThisIsMissEm:
Just a heads up: the demoClientApp is still broken due to it's implementation (the IdP is not persisted across page loads, which means you can't actually test the client works with different IdP's in e2e tests).

PR #1928, opened 2022-01-25
Refactor CI and CD workflows

NSeydoux:
This refactors the CI/CD workflows to align with https://github.com/inrupt/solid-client-js/commit/1683df65f6752f33b7dbe269edc451eb99d75cdc. The main improvement is that Node end-to-end test run on a matrix for environments, which enables defining the same secret names with environment-specific values.

Linting, audit and end-to-end tests run in their own workflows rather than in a single CI one (which still runs the unit tests). The new workflows are still gating, so there is no functional change compared to the previous setting, but each individual workflow fails faster if necessary, and reporting is more granular from the PR view. 

There are some differences with the solid-client setup that I'd like to draw a special attention to for discussion: 
- Running the audit workflow is disabled on dependabot PR
- In the CD workflow, only the first job is  skipped because of dependabot, which automatically skips the others
- Remove DEFAULT_NODE_VERSION from CI
- Use lint:check instead of lint:ci
- Change check-licenses and list-licenses into licenses:check and licenses:list
- Use exact version of github actions

Note that the browser-based tests are still flaky, and still need to be moved from testcafé to playwright. It will be done in a separate PR, which will also update them to use the same environment-driven setup as the Node tests.


e2e/setup/e2e-setup.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (typeof process.env.E2E_TEST_CLIENT_ID !== "string") {
+    throw new Error(
+      "The environment variable E2E_TEST_CLIENT_ID is undefined."
+    );
+  }
+  if (typeof process.env.E2E_TEST_CLIENT_SECRET !== "string") {
+    throw new Error(
+      "The environment variable E2E_TEST_CLIENT_SECRET is undefined."
+    );
+  }
```
matthieubosquet, line 87: Considering all this, I would probably extract the whole thing to an `asserts is TestingEnvironment` type guard.

e2e/setup/e2e-setup.ts
```diff
@@ -0,0 +1,103 @@
...
+    environment: process.env.E2E_TEST_ENVIRONMENT,
+    // feature: {}
+  };
+}
+
+export function formatForDisplay(iri: string): string {
+  // Trim `https://` from the start of these URLs,
+  // so that GitHub Actions doesn't replace them with *** in the logs.
+  return iri.replace(/^https?:\/\//, "");
+}
```
matthieubosquet, line 103: I don't think you want to use that. Those are secrets, they should be replaced with *** in the logs.

The logs will show which environment the tests are running against. That's enough.

packages/node/e2e/e2e-test.spec.ts
```diff
@@ -1,246 +1,157 @@
...
+      const response = await authenticatedSession.fetch(publicResourceUrl, {
+        headers: {
+          Accept: "text/turtle",
+        },
+      });
+      expect(response.status).toBe(200);
+      await expect(response.text()).resolves.toContain(
+        ":PersonalProfileDocument"
+      );
+    });
```
matthieubosquet, line 148:
```suggestion
    it("can fetch the WebID after logging out", async () => {
      const publicResourceUrl = authenticatedSession.info.webId as string;
      await authenticatedSession.logout();
      const response = await authenticatedSession.fetch(publicResourceUrl, {
        headers: {
          Accept: "text/turtle",
        },
      });
      expect(response.status).toBe(200);
      await expect(response.text()).resolves.toContain(
        ":PersonalProfileDocument"
      );
    });
```

This doesn't really test "a public resource" it tests the WebID (not necessarily hosted on a Solid server).

matthieubosquet (approved):
A few comments, but looks good to me.

The main one might be about the environments under test.

.github/workflows/cd-preview.yml
```diff
@@ -1,6 +1,13 @@
 name: CD-preview
 
-on: [push, workflow_dispatch]
+on: 
+  push:
+    branches:
+        - "**"
```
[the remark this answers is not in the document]
NSeydoux: The YAML schema disagrees :)

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
-        os: [ubuntu-20.04, windows-2019]
+        os: [ubuntu-20.04, windows-2019, macos-10.15]
```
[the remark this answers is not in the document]
NSeydoux: The rationale for not having them is for CI not to start breaking unexpectedly when the underlying change happens, but rather do the migration explicitly in our own time.

.github/workflows/e2e-node.yml
```diff
@@ -0,0 +1,45 @@
...
+    if: ${{ github.actor != 'dependabot[bot]' }}
+    runs-on: ${{ matrix.os }}
+    environment:
+      name: ${{ matrix.environment-name }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [16.x]
+        # NSS does not support static client registration, which we rely on for testing.
+        environment-name: ["ESS Production", "ESS Dev-Next"]
```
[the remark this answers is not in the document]
NSeydoux: Yup, even though currently 1.1 and Production would be similar, right ?

e2e/setup/e2e-setup.ts
```diff
@@ -0,0 +1,103 @@
...
+  "ESS Production" as const,
+  "ESS 1.1" as const,
+  "NSS" as const,
+];
+
+export type AvailableEnvironment = typeof availableEnvironment extends Array<
+  infer E
+>
+  ? E
+  : never;
```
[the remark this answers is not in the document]
NSeydoux: I knew you'd like it

e2e/setup/e2e-setup.ts
NSeydoux: Fair point, having the environment prevents from having to use this

packages/node/e2e/e2e-test.spec.ts
NSeydoux: Right, I have a bunch of FIXMEs for this, I'll add one here.

.github/workflows/cd-preview.yml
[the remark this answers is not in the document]
NSeydoux: `*` doesn't match `/`, but `**` does: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

e2e/setup/e2e-setup.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  if (missing) {
+    throw new Error(`Environment variable missing: ${missingMessage}`);
+  }
+}
+
+export function getTestingEnvironment(): TestingEnvironment {
+  if (!isAvailableEnvironment(process.env.E2E_TEST_ENVIRONMENT)) {
+    throw new Error(`Unknown environment: ${process.env.E2E_TEST_ENVIRONMENT}`);
+  }
```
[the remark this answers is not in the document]
NSeydoux: Interesting, I didn't know about that kind of type guard, nice 282b8eaa8b9c1087f8e482b96123cf808b6a0a6e

e2e/setup/e2e-setup.ts
NSeydoux: Done in 282b8eaa8b9c1087f8e482b96123cf808b6a0a6e

e2e/setup/e2e-setup.ts
NSeydoux: Done in 807b7318c71e16a7f36f1f20d1601dae8ac8a1bb

packages/node/e2e/e2e-test.spec.ts
NSeydoux: Done in 38864feb0ba578c8684e4caeccab899e09048b66. I want to have resource setup and teardown there, so it's definitely out of scope for this PR

Issue #1842, opened 2021-12-10
URL query parameters do not get removed in htu field


joachimvh:
### Search terms you've used
htu

### Impacted package

Which packages do you think might be impacted by the bug ?

- [x] solid-client-authn-core

### Bug description
The `htu` field of the access token is generated [here](https://github.com/inrupt/solid-client-authn-js/blob/c3864dc9d978176d682a2c90615d77c29f48a053/packages/core/src/authenticatedFetch/dpopUtils.ts#L61). Actual code of the [removeHashUsernameAndPassword](https://github.com/inrupt/solid-client-authn-js/blob/c3864dc9d978176d682a2c90615d77c29f48a053/packages/core/src/authenticatedFetch/dpopUtils.ts#L33-L39) function is as follows: 
```js
function removeHashUsernameAndPassword(audience: string): string {
  const cleanedAudience = new URL(audience);
  cleanedAudience.hash = "";
  cleanedAudience.username = "";
  cleanedAudience.password = "";
  return cleanedAudience.toString();
}
```

The issue is that the above code does not clear the query parameters of the URL. `http://example.com/?query` remains unchanged after going through that function, but the `htu` parameter is defined as
 > The HTTP URI used for the request, without query and fragment parts

### To Reproduce
See https://github.com/solid/community-server-recipes/issues/3

Or just run the code above with any URL containing a query parameter.

### Expected result
`htu: http://example.com/`

### Actual result
`htu: http://example.com/?query`

NSeydoux:
Thanks for reporting @joachimvh , we'll fix this soon.

diegoaraujo:
Any news?

ThisIsMissEm:
This was released in [1.11.4 and onwards](https://github.com/inrupt/solid-client-authn-js/blob/main/CHANGELOG.md#1114---2022-02-11), sorry we didn't follow up properly on this. We've also just done a fix for the `iss` parameter in #2103, which should be released soon-ish.

Closing as this is done.

PR #1932, opened 2022-01-26
Remove the URL query parameters from the HTU field.

NSeydoux:
This PR fixes bug #1842.

### DIFF
--- a/packages/core/src/authenticatedFetch/dpopUtils.ts
+++ b/packages/core/src/authenticatedFetch/dpopUtils.ts
@@ -30,11 +30,12 @@ import { PREFERRED_SIGNING_ALG } from "../constant";
  * @returns The normalized URL as a string.
  * @hidden
  */
-function removeHashUsernameAndPassword(audience: string): string {
+function normalizeHTU(audience: string): string {
   const cleanedAudience = new URL(audience);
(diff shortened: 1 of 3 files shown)


packages/core/src/authenticatedFetch/dpopUtils.ts
```diff
@@ -30,11 +30,12 @@ import { PREFERRED_SIGNING_ALG } from "../constant";
  * @returns The normalized URL as a string.
  * @hidden
  */
-function removeHashUsernameAndPassword(audience: string): string {
+function normalizeHTU(audience: string): string {
   const cleanedAudience = new URL(audience);
   cleanedAudience.hash = "";
   cleanedAudience.username = "";
   cleanedAudience.password = "";
+  cleanedAudience.search = "";
```
ajacksified, line 38: Would it be simpler to create a new URL with only protocol + pathname, rather than excluding fields?

NSeydoux: That would probably be cleaner, and prevent from future bugs like this one. What do you think about 9e1f1fec243379593be346900c0e8711fd4d78aa ?

ajacksified (approved):

PR #1942, opened 2022-01-27
Temporarily remove MacOS from CI matrix

NSeydoux:
The MacOS runners seem to be overrun at the moment, and the CI jobs aren't getting through. It is possible the large amount of Dependabot PRs in parallel are causing that (because they were just enabled on demo packages, and there was some catching up to do). I will try to re-enable them once the PR backlog is back to normal.


NSeydoux:
Looks like the issue is somewhat resolved. Will reopen if if acts up again.

Issue #1991, opened 2022-02-23
Code break in v1.11.5


IshaanOhri:
### Impacted package

Which packages do you think might be impacted by the bug ?

- [ ] solid-client-authn-browser
- [ ] solid-client-authn-core
- [ ] oidc-client-ext

### Bug description
The login process doesn't redirect to Inrupt's OAuth page. It throws an error.

### To Reproduce
1. Clone my git repo (`https://github.com/IshaanOhri/pod-nextjs.git`) on commit `ce198a88ffe3b1ee1c44c2da1145af68713e080d` where I have a basic code implementing the login functionality

### Expected result
The Inrupt's OAuth page should open up where I can enter my credentials

### Actual result
Following error occurs:

![image](https://user-images.githubusercontent.com/39885197/155282554-cf996724-2241-4e20-976d-79f2eb3ca8d6.png)

## Additional information
The code works perfectly fine on v1.11.3

jeff-zucker:
I am not sure this is the same issue but the dead-simple page below works against solidcommunity.net in 1.11.3 but produces this error in `1.11.5
```
solid-client-authn.bundle.js:2 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'includes')
    at solid-client-authn.bundle.js:2:47779
    at t.handleRegistration (solid-client-authn.bundle.js:2:47865)
    at t.default.handle (solid-client-authn.bundle.js:2:20696)
    at async t.default.login (solid-client-authn.bundle.js:2:6519)
    at async login (solid-client-authn.bundle.js:2:8394)
(anonymous) @ solid-client-authn.bundle.js:2
t.handleRegistration @ solid-client-authn.bundle.js:2
handle @ solid-client-authn.bundle.js:2
await in handle (async)
t.login @ solid-client-authn.bundle.js:2
loginButton.onclick @ a.html:17
```
Here's the page : 
```html
<span id="webId"></span>
<input id="oidc" value="https://solidcommunity.net" style="width:24em" />
<button id="loginButton">Login</button>
<button id="logoutButton">Logout</button>

<script src="https://cdn.jsdelivr.net/npm/@inrupt/solid-client-authn-browser@1.11.5/dist/solid-client-authn.bundle.js"></script>

<script>

const iscan = solidClientAuthentication;
var session = iscan.getDefaultSession();

const loginButton = document.querySelector("#loginButton");
const logoutButton = document.querySelector("#logoutButton");
[...]

PR #1968, opened 2022-02-10
Update Solid-OIDC discovery

NSeydoux:
The code implemented a legacy solid-oidc discovery. This updates the discovery to be aligned with https://solid.github.io/solid-oidc/#discovery.

This PR fixes bug #1927.

- [X] I've added a unit test to test for potential regressions of this bug.
- [X] The changelog has been updated, if applicable.
- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #1994, opened 2022-02-25
Ensure we check that scopesSupported actually exists

ThisIsMissEm:
This PR fixes bug #1991 where by `t.scopesSupport` is undefined — basically the code in https://github.com/inrupt/solid-client-authn-js/pull/1968 incorrectly assumed that `scopesSupported` was always present, but the marshalling code in https://github.com/inrupt/solid-client-authn-js/blob/57849d2ff0a37892650d19fdc6e6fa2924ee599b/packages/browser/src/login/oidc/IssuerConfigFetcher.ts doesn't actually ensure the shape of the IssuerConfig, and the [use of `as unknown as IIssuerConfig`](https://github.com/inrupt/solid-client-authn-js/blob/57849d2ff0a37892650d19fdc6e6fa2924ee599b/packages/browser/src/login/oidc/IssuerConfigFetcher.ts#L143) results in us accessing a key that doesn't actually exist.
 
A further fix would be to actually explicitly build the IssuerConfig structure from the values, without using `as`, such that we build an structure with defaults set (like we do in node.js). The issue we saw was that `scopesSupported` in node is always present due to how it builds IssuerConfig, but in browser, that's not true.

We could optionally make this PR actually implement a proper parsing of IssuerConfig with default values, and ensure that both browser and node truly do conform to the structures we expect, though I think that may be out of scope for a fix on `1.11.x`

### DIFF
--- a/packages/core/src/login/oidc/IClientRegistrar.ts
+++ b/packages/core/src/login/oidc/IClientRegistrar.ts
@@ -77,6 +77,7 @@ function determineClientType(
     return "static";
   }
   if (
+    issuerConfig.scopesSupported &&
     issuerConfig.scopesSupported.includes("webid") &&
     options.clientId !== undefined &&
     isValidUrl(options.clientId)
--- a/packages/core/src/login/oidc/IIssuerConfig.ts
+++ b/packages/core/src/login/oidc/IIssuerConfig.ts
@@ -38,7 +38,7 @@ export interface IIssuerConfig {
   userinfoEndpoint?: string;
   jwksUri: string;
   registrationEndpoint?: string;
-  scopesSupported: string[];
+  scopesSupported?: string[];
   responseTypesSupported?: string[];
   responseModesSupported?: string[];
   grantTypesSupported?: string[];


matthieubosquet (approved):

NSeydoux:
I think https://github.com/inrupt/solid-client-authn-js/pull/2000 supersedes this, but +1 on properly parsing the OpenID provider config, the legacy code currently used in the browser module is not providing the type safety it should.

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -77,6 +77,7 @@ function determineClientType(
     return "static";
   }
   if (
+    issuerConfig.scopesSupported &&
```
ThisIsMissEm, line 80:
```suggestion
    Array.isArray(issuerConfig.scopesSupported) &&
```

ThisIsMissEm:
> I think #2000 supersedes this, but +1 on properly parsing the OpenID provider config, the legacy code currently used in the browser module is not providing the type safety it should.

I think it'd still be a good idea to keep this code on the defensive side, i.e., have a second layer of checks beyond the initial parsing

PR #2002, opened 2022-03-01
Prevent silent auth cycle on authentication error

NSeydoux:
When an authentication error happens, local storage is now cleared in order to prevent further silent login attempts. This should fix the issue where failed login cycle endlessly.

In order to implement this, the Session object listens on authentication events bubbled up by the handlers. In particular, on login, the current session ID is stored in local storage to enable silent auth in the future, and it is cleared on logout/auth error.

### DIFF
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -922,6 +922,44 @@ describe("Session", () => {
       expect(window.localStorage.getItem(KEY_CURRENT_URL)).toBeNull();
       expect(clientAuthentication.login).not.toHaveBeenCalled();
     });
+
+    it("does not retry silent authentication if an authentication error happens", async () => {
+      const sessionId = "mySession";
+      // Pretend that we have a previously active session.
+      mockLocalStorage({
+        [KEY_CURRENT_SESSION]: sessionId,
+      });
+      mockLocation("https://mock.current/location");
+      const mockedStorage = new StorageUtility(
+        mockStorage({
+          [`${USER_SESSION_PREFIX}:${sessionId}`]: {
+            isLoggedIn: "true",
+          },
+        }),
+        mockStorage({})
+      );
+      const clientAuthentication = mockClientAuthentication({
+        sessionInfoManager: mockSessionInfoManager(mockedStorage),
+      });
+      clientAuthentication.handleIncomingRedirect = jest
+        .fn(clientAuthentication.handleIncomingRedirect)
+        .mockImplementationOnce(async (_url, emitter) => {
+          emitter.emit(EVENTS.ERROR, "Some authentication error");
+          return {
+            isLoggedIn: false,
+            sessionId,
+          };
+        });
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      // eslint-disable-next-line no-void
+      void mySession.handleIncomingRedirect("https://some.redirect/url");
+      // The local storage should have been cleared by the auth error
+      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBeNull();
+      // Silent authentication should not have been attempted
+      expect(clientAuthentication.login).not.toHaveBeenCalled();
+    });
   });
 
   describe("onLogin", () => {
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -270,6 +270,24 @@ describe("Session", () => {
       expect(mySession.info.webId).toBe("https://some.webid#them");
     });
 
+    it("updates the localStorage if the login is completed", async () => {
+      const clientAuthentication = mockClientAuthentication();
+      clientAuthentication.handleIncomingRedirect = jest.fn(
+        async (_url: string) => {
+          return {
+            isLoggedIn: true,
+            sessionId: "a session ID",
+            webId: "https://some.webid#them",
+          };
+        }
+      );
+      const mySession = new Session({ clientAuthentication });
+      await mySession.handleIncomingRedirect("https://some.url");
+      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBe(
+        mySession.info.sessionId
+      );
+    });
+
     it("directly returns the session's info if already logged in", async () => {
       const clientAuthentication = mockClientAuthentication();
       clientAuthentication.handleIncomingRedirect = jest.fn(
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -973,10 +973,36 @@ describe("Session", () => {
       const mySession = new Session({ clientAuthentication });
       // eslint-disable-next-line no-void
       void mySession.handleIncomingRedirect("https://some.redirect/url");
+      expect(mySession.info.isLoggedIn).toBe(false);
       // The local storage should have been cleared by the auth error
       expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBeNull();
       // Silent authentication should not have been attempted
       expect(clientAuthentication.login).not.toHaveBeenCalled();
+    });
+
+    it("clears the session ID from local storage if the session expired", async () => {
+      const sessionId = "mySession";
+      // Pretend that we have a previously active session.
+      mockLocalStorage({
+        [KEY_CURRENT_SESSION]: sessionId,
+      });
+      mockLocation("https://mock.current/location");
+      const mockedStorage = new StorageUtility(
+        mockStorage({
+          [`${USER_SESSION_PREFIX}:${sessionId}`]: {
+            isLoggedIn: "true",
+          },
+        }),
+        mockStorage({})
+      );
+      const clientAuthentication = mockClientAuthentication({
+        sessionInfoManager: mockSessionInfoManager(mockedStorage),
+      });
+
+      const mySession = new Session({ clientAuthentication });
+      mySession.emit(EVENTS.SESSION_EXPIRED);
+      // The local storage should have been cleared by the auth error
+      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBeNull();
     });
   });
 
--- a/packages/browser/src/Session.ts
+++ b/packages/browser/src/Session.ts
@@ -230,7 +230,7 @@ export class Session extends EventEmitter {
 
     // When a session is logged in, we want to track its ID in local storage to
     // enable silent refresh.
-    this.on("login", () => {
+    this.on(EVENTS.LOGIN, () => {
       // Store the current session ID specifically in 'localStorage' (i.e., not using
       // any other storage mechanism), as we don't deem this information to be
       // sensitive, and we want to ensure it survives a browser tab refresh.
@@ -238,7 +238,11 @@ export class Session extends EventEmitter {
     });
 
     // On logout, silent refresh should no longer be considered a viable option.
-    this.on("logout", () => {
+    this.on(EVENTS.LOGOUT, () => {
+      window.localStorage.removeItem(KEY_CURRENT_SESSION);
+    });
+
+    this.on(EVENTS.SESSION_EXPIRED, () => {
       window.localStorage.removeItem(KEY_CURRENT_SESSION);
     });
 
@@ -299,7 +303,7 @@ export class Session extends EventEmitter {
     await this.clientAuthentication.logout(this.info.sessionId);
     this.info.isLoggedIn = false;
     this.tmpFetchWithCookies = false;
-    this.emit("logout");
+    this.emit(EVENTS.LOGOUT);
   };
 
   /**
@@ -415,13 +419,13 @@ export class Session extends EventEmitter {
       if (currentUrl === null) {
         // The login event can only be triggered **after** the user has been
         // redirected from the IdP with access and ID tokens.
-        this.emit("login");
+        this.emit(EVENTS.LOGIN);
       } else {
         // If an URL is stored in local storage, we are being logged in after a
         // silent authentication, so remove our currently stored URL location
         // to clean up our state now that we are completing the re-login process.
         window.localStorage.removeItem(KEY_CURRENT_URL);
-        this.emit("sessionRestore", currentUrl);
+        this.emit(EVENTS.SESSION_RESTORED, currentUrl);
       }
     } else if (options.restorePreviousSession === true) {
       // Silent authentication happens after a refresh, which means there are no
@@ -463,7 +467,7 @@ export class Session extends EventEmitter {
    * @param callback The function called when a user completes login.
    */
   onLogin(callback: () => unknown): void {
-    this.on("login", callback);
+    this.on(EVENTS.LOGIN, callback);
   }
 
   /**
@@ -472,7 +476,7 @@ export class Session extends EventEmitter {
    * @param callback The function called when a user completes logout.
    */
   onLogout(callback: () => unknown): void {
-    this.on("logout", callback);
+    this.on(EVENTS.LOGOUT, callback);
   }
 
   /**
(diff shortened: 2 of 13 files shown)


packages/browser/src/Session.spec.ts
```diff
@@ -922,6 +922,44 @@ describe("Session", () => {
...
+        });
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      // eslint-disable-next-line no-void
+      void mySession.handleIncomingRedirect("https://some.redirect/url");
+      // The local storage should have been cleared by the auth error
+      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBeNull();
+      // Silent authentication should not have been attempted
+      expect(clientAuthentication.login).not.toHaveBeenCalled();
```
ThisIsMissEm, line 961: I think we also want to assert here that `mySession.info.isLoggedIn` is false, too

packages/browser/src/Session.ts
```diff
@@ -227,6 +227,27 @@ export class Session extends EventEmitter {
...
       )
     );
+
+    // When a session is logged in, we want to track its ID in local storage to
+    // enable silent refresh.
+    this.on("login", () => {
+      // Store the current session ID specifically in 'localStorage' (i.e., not using
+      // any other storage mechanism), as we don't deem this information to be
+      // sensitive, and we want to ensure it survives a browser tab refresh.
+      window.localStorage.setItem(KEY_CURRENT_SESSION, this.info.sessionId);
```
ThisIsMissEm, line 237: Is there a reason to not use the SessionInfoManager here? Or at least, to wrap the interaction with localstorage in a class?

packages/browser/src/Session.ts
```diff
@@ -227,6 +227,27 @@ export class Session extends EventEmitter {
...
+    });
+
+    // On logout, silent refresh should no longer be considered a viable option.
+    this.on("logout", () => {
```
ThisIsMissEm, line 241: Question: do we reliably have a "session expired" event? If so, we should also remove there which might assist in preventing the `client_invalid` errors

packages/browser/src/Session.ts
```diff
@@ -227,6 +227,27 @@ export class Session extends EventEmitter {
...
+      window.localStorage.removeItem(KEY_CURRENT_SESSION);
+    });
+
+    // If an error happens, silent refresh should no longer be considered a viable option
+    // to prevent endless redirection, and session information should be cleared.
+    this.on(EVENTS.ERROR, async () => {
+      window.localStorage.removeItem(KEY_CURRENT_SESSION);
+      await this.logout();
```
ThisIsMissEm, line 249: I wasn't sure in here if calling `logout()` was correct — I think we actually want a "logout" that doesn't come from user interaction or SDK API call, that way we can differentiate a logout from a session had an error?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -456,11 +455,6 @@ describe("AuthCodeRedirectHandler", () => {
       await authCodeRedirectHandler.handle(
         "https://coolsite.com/redirect?code=someCode&state=oauth2StateValue"
       );
-      // Check that the current session is stored correctly __specifically__ in
-      // 'localStorage'.
-      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBe(
```
ThisIsMissEm, line 461: Why isn't this check needed? Should the assertion be covered in a different unit test?

packages/browser/src/Session.spec.ts
```diff
@@ -922,6 +922,44 @@ describe("Session", () => {
...
+            sessionId,
+          };
+        });
+      clientAuthentication.login = jest.fn();
+
+      const mySession = new Session({ clientAuthentication });
+      // eslint-disable-next-line no-void
+      void mySession.handleIncomingRedirect("https://some.redirect/url");
+      // The local storage should have been cleared by the auth error
+      expect(window.localStorage.getItem(KEY_CURRENT_SESSION)).toBeNull();
```
ThisIsMissEm, line 959: I can't see a test that actually checks this value transitioning state, that might be a good unit test to have?

packages/browser/src/Session.ts
NSeydoux: The reason for it being separate from the `SessionInfoManager` is because this is a very browser-specific mechanism, which doesn't make sense in a browser context: it is only useful for silent login on refresh. `SessionInfoManager`, on the other hand, is the abstraction layer on top of storage which is shared between the browser and node codebase.

It could be wrapped in a separate class, but the intent is that it is basically scoped to the `Session` class, so I'm not sure the extra indirection is required.

packages/browser/src/Session.ts
NSeydoux: That's a good point, and we do. It may save a round trip in some cases, even though the event being thrown is quite unlikely: either the browser tab where the refresh happens is in focus, and refresh happens automatically in the background, or it is out of focus, the refresh doesn't happen in the background, but sending the session expired signal doesn't happen either (or at least this is my current understanding). 

Regardless, session expiration is a possible event, and it should be handled here.

packages/browser/src/Session.ts
ThisIsMissEm: Right, it's just that we're not able to mock it as a clear interface (we're mocking localstorage instead) and we're duplicating keys, plus there's a few other localStorage interactions that happen directly, and they're not all in one file.

packages/browser/src/Session.ts
ThisIsMissEm: hmm, is there maybe a way of tracking when a refresh should've happened, and if it didn't on refocus of the tab (https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) we can detect that the session has potentially expired and we trigger a "refresh the session now", and then clear if it's returning `interaction_required`?

packages/browser/src/Session.ts
NSeydoux: I think a consuming app may already distinguish between the two, by registering different callbacks to `onError` vs `onLogout`. However, it is true that calling `logout` here will also send the logout signal. I'm not sure if it is a wrong thing or not...

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
NSeydoux: That's right, it has been moved from `AuthCodeRedirectHandler` to `Session`. I'll cover this in an appropriate unit test.

packages/browser/src/Session.ts
ThisIsMissEm: I feel like it may be?

packages/browser/src/Session.spec.ts
NSeydoux: Fixed in 3be5eb8b80dd725d59a031b4f9fa76ea5f76c048

packages/browser/src/Session.ts
NSeydoux: Fixed in cadb21e7095c3d020f29d6146c7cd7e5ae1561fa and 124c004311e95d3de470bc408732cfc469cb1d80

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
NSeydoux: Fixed in 124c004311e95d3de470bc408732cfc469cb1d80

packages/browser/src/Session.spec.ts
NSeydoux: Fixed in 124c004311e95d3de470bc408732cfc469cb1d80

packages/browser/src/Session.ts
```diff
@@ -227,6 +227,31 @@ export class Session extends EventEmitter {
...
+      window.localStorage.setItem(KEY_CURRENT_SESSION, this.info.sessionId);
+    });
+
+    // On logout, silent refresh should no longer be considered a viable option.
+    this.on(EVENTS.LOGOUT, () => {
+      window.localStorage.removeItem(KEY_CURRENT_SESSION);
+    });
+
+    this.on(EVENTS.SESSION_EXPIRED, () => {
+      window.localStorage.removeItem(KEY_CURRENT_SESSION);
```
ThisIsMissEm, line 246: Does this also need to `await this.logout()` or is error emitted afterwards?

packages/browser/src/Session.spec.ts
```diff
@@ -922,6 +940,70 @@ describe("Session", () => {
...
+          },
+        }),
+        mockStorage({})
+      );
+      const clientAuthentication = mockClientAuthentication({
+        sessionInfoManager: mockSessionInfoManager(mockedStorage),
+      });
+
+      const mySession = new Session({ clientAuthentication });
+      mySession.emit(EVENTS.SESSION_EXPIRED);
```
ThisIsMissEm, line 1003: Can we test this by actually inducing the expiry in mySession, instead of injecting the event? Just thinking if the code that produces this event may emit other events or change state in another way?

packages/core/src/constant.ts
```diff
@@ -32,10 +32,13 @@ export const SOLID_CLIENT_AUTHN_KEY_PREFIX = "solidClientAuthn:";
 export const PREFERRED_SIGNING_ALG = ["ES256", "RS256"];
 
 export const EVENTS = {
-  NEW_REFRESH_TOKEN: "newRefreshToken",
   ERROR: "onError",
```
ThisIsMissEm, line 35: this ones a tad odd? Can we fix it to just be `"error"` ?

ThisIsMissEm (commented):
This is looking good, I've several more questions though!

packages/browser/src/Session.ts
NSeydoux: Yeah, some of these localStorage interaction are only required for a deprected workaround, and I'm currently investigating to see if anyone is still relying on it (I don't think so, and if not I'd want to remove it altogether). Otherwise, I think the only significant localStorage interaction is in `ClientAuthentication`, and I'll move it to the `Session` file so that it is all located in one place.

I don't know if there would be clear benefits in mocking it as a different interface, as this is explicitly relying on the browser local storage, as opposed to the `StorageUtility` we use otherwise, which is injected a storage instead.

packages/browser/src/Session.spec.ts
NSeydoux: I think it would be possible to mock this by actually expiring the session, by mocking a token response with zero time to live, but that would be more involved. Also, the portion of the code emitting this signal is meant to be isolated: it's the closure where the tokens/keys are kept. The signal is meant to be the only way for this closure to change state. Other signals could be emitted, but they should not impact the session state.

packages/core/src/constant.ts
NSeydoux: Yeah I was thinking that too, and I was initially reluctant because that's technically a public API, but it's not documented anywhere (people are expected to use the `onError` callback instead), so I'll go ahead and rename this.

packages/browser/src/Session.ts
NSeydoux: a66d20282c4d4abab963a65a6449e6e364ede7ba should clean this up. The logout signal is now only sent in case of user-initiated logout, and all the logout cleanup is located in one place.

packages/browser/src/Session.ts
NSeydoux: Fixed in a66d20282c4d4abab963a65a6449e6e364ede7ba, the error wasn't related to an actual logout indeed.

ThisIsMissEm:
The problem with onError is that there's no way to remove the listener, which is actually needed in the context of UI frameworks.

I think we'd be better exposing the on/off interface directly.

> On 3. Mar 2022, at 11:20, Zwifi ***@***.***> wrote:
> 
> ﻿
> @NSeydoux commented on this pull request.
> 
> In packages/core/src/constant.ts:
> 
> > @@ -32,10 +32,13 @@ export const SOLID_CLIENT_AUTHN_KEY_PREFIX = "solidClientAuthn:";
>  export const PREFERRED_SIGNING_ALG = ["ES256", "RS256"];
>  
>  export const EVENTS = {
> -  NEW_REFRESH_TOKEN: "newRefreshToken",
>    ERROR: "onError",
> Yeah I was thinking that too, and I was initially reluctant because that's technically a public API, but it's not documented anywhere (people are expected to use the onError callback instead), so I'll go ahead and rename this.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> You are receiving this because your review was requested.

-- 
This e-mail, and any attachments thereto, is intended only for use by the 
addressee(s) named herein and may contain legally privileged, confidential 
and/or proprietary information. If you are not the intended recipient of 
this e-mail (or the person responsible for delivering this document to the 
intended recipient), please do not disseminate, distribute, print or copy 
this e-mail, or any attachment thereto. If you have received this e-mail in 
error, please respond to the individual sending the message, and 
permanently delete the email.

ThisIsMissEm:
Wait, what? How so?

> On 3. Mar 2022, at 11:42, Zwifi ***@***.***> wrote:
> 
> ﻿
> @NSeydoux commented on this pull request.
> 
> In packages/core/src/constant.ts:
> 
> > @@ -32,10 +32,13 @@ export const SOLID_CLIENT_AUTHN_KEY_PREFIX = "solidClientAuthn:";
>  export const PREFERRED_SIGNING_ALG = ["ES256", "RS256"];
>  
>  export const EVENTS = {
> -  NEW_REFRESH_TOKEN: "newRefreshToken",
>    ERROR: "onError",
> Aha actually error is an invalid event name, which is probably why it was named as such in the first place. I totally forgot about this, I'll add a comment for the ever forgetful future me.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> You are receiving this because your review was requested.

-- 
This e-mail, and any attachments thereto, is intended only for use by the 
addressee(s) named herein and may contain legally privileged, confidential 
and/or proprietary information. If you are not the intended recipient of 
this e-mail (or the person responsible for delivering this document to the 
intended recipient), please do not disseminate, distribute, print or copy 
this e-mail, or any attachment thereto. If you have received this e-mail in 
error, please respond to the individual sending the message, and 
permanently delete the email.

NSeydoux:
If you run the tests from https://github.com/inrupt/solid-client-authn-js/pull/2002/commits/4a1566b2d12ab55732e7591a9069c9af1a0a8fb6, an exception is thrown complaining about an "invalid event name"

ThisIsMissEm (approved):

packages/core/src/constant.ts
```diff
@@ -32,10 +32,14 @@ export const SOLID_CLIENT_AUTHN_KEY_PREFIX = "solidClientAuthn:";
 export const PREFERRED_SIGNING_ALG = ["ES256", "RS256"];
 
 export const EVENTS = {
-  NEW_REFRESH_TOKEN: "newRefreshToken",
+  // `error` is not a valid event name, hence the `onError`.
```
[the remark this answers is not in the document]
NSeydoux: I updated the comment to reference this docs page, because future me will need a reminder at some point.

packages/core/src/constant.ts
```diff
@@ -32,10 +32,14 @@ export const SOLID_CLIENT_AUTHN_KEY_PREFIX = "solidClientAuthn:";
 export const PREFERRED_SIGNING_ALG = ["ES256", "RS256"];
 
 export const EVENTS = {
+  // Note that an `error` events MUST be listened to: https://nodejs.org/dist/latest-v16.x/docs/api/events.html#error-events.
+  ERROR: "error",
```
ThisIsMissEm, line 36: Where we emit error events, you can also use `emitter.listenerCount("error")` before actually emitting

PR #2009, opened 2022-03-03
Clear pkce verifier

NSeydoux:
This clears the PKCE verifier from storage after it has been retrieved to perform the token exchange. 

 Note that the PKCE verifier was stored in two distinct instances:
- As part of the `oidc-client` storage, which was not cleared when binding the exchanged token to a DPoP key because this exchange relies on our custom extension `oidc-client-ext`, rather than on the main library.
-   As part of the session  storage from our library.

A lot of the PR is just test refactoring.

-  [X] I've added a unit test to test for potential regressions of this bug.

### DIFF
--- a/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -359,7 +359,6 @@ describe("AuthCodeRedirectHandler", () => {
 
     it("returns an authenticated bearer fetch if requested", async () => {
       mockOidcClient();
-      mockLocalStorage({});
 
       const mockedFetch = mockFetch(
         new Response("", {
@@ -398,7 +397,6 @@ describe("AuthCodeRedirectHandler", () => {
 
     it("returns an authenticated DPoP fetch if requested", async () => {
       mockOidcClient();
-      mockLocalStorage({});
       const mockedFetch = jest.fn(global.fetch).mockReturnValue(
         new Promise((resolve) => {
           resolve(
@@ -436,7 +434,6 @@ describe("AuthCodeRedirectHandler", () => {
 
     it("saves session information in storage on successful login", async () => {
       mockOidcClient();
-      mockLocalStorage({});
       window.fetch = jest.fn(global.fetch).mockReturnValue(
         new Promise((resolve) => {
           resolve(
@@ -479,7 +476,6 @@ describe("AuthCodeRedirectHandler", () => {
 
     it("preserves any query strings from the redirect URI", async () => {
       mockOidcClient();
-      mockLocalStorage({});
 
       window.fetch = (jest.fn() as any).mockReturnValue(
         new Promise((resolve) => {
@@ -626,7 +622,6 @@ describe("AuthCodeRedirectHandler", () => {
 
   it("stores information about the resource server cookie in local storage on successful authentication", async () => {
     mockOidcClient();
-    mockLocalStorage({});
 
     // This mocks the fetch to the Resource Server session endpoint
     // Note: Currently, the endpoint only returns the webid in plain/text, it could
@@ -818,7 +813,9 @@ describe("AuthCodeRedirectHandler", () => {
   });
 
   it("returns a fetch that supports the refresh flow", async () => {
-    window.localStorage.setItem("tmp-resource-server-session-enabled", "false");
+    mockLocalStorage({
+      "tmp-resource-server-session-enabled": "false",
+    });
     mockOidcClient();
     const mockedStorage = mockDefaultStorageUtility();
     const coreModule = jest.requireActual(
--- a/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
+++ b/packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
@@ -31,7 +31,7 @@ import {
 } from "@inrupt/solid-client-authn-core";
 import type * as SolidClientAuthnCore from "@inrupt/solid-client-authn-core";
 import { jest, it, describe, expect } from "@jest/globals";
-import { CodeExchangeResult } from "@inrupt/oidc-client-ext";
+import { CodeExchangeResult, getBearerToken } from "@inrupt/oidc-client-ext";
 import { Response } from "cross-fetch";
 import { JWK, importJWK } from "jose";
 import { KeyObject } from "crypto";
@@ -250,10 +250,10 @@ const mockDefaultStorageUtility = (
   }
 ) =>
   mockStorageUtility({
-    "solidClientAuthenticationUser:oauth2StateValue": {
+    [`${USER_SESSION_PREFIX}:${DEFAULT_OAUTH_STATE}`]: {
       sessionId: "mySession",
     },
-    "solidClientAuthenticationUser:mySession": {
+    [`${USER_SESSION_PREFIX}:mySession`]: {
       dpop: options.dpop ? "true" : "false",
       issuer: mockIssuer().issuer.toString(),
       codeVerifier: options.codeVerifier ?? DEFAULT_CODE_VERIFIER,
@@ -342,6 +342,42 @@ describe("AuthCodeRedirectHandler", () => {
       );
     });
 
+    it("throws if the code verifier is missing from storage", async () => {
+      mockOidcClient();
+      mockFetch(
+        new Response("", {
+          status: 200,
+        })
+      );
+      const storageUtility = mockDefaultStorageUtility();
+      await storageUtility.deleteForUser("mySession", "codeVerifier");
+      const authCodeRedirectHandler = getAuthCodeRedirectHandler({
+        storageUtility,
+      });
+
+      await expect(
+        authCodeRedirectHandler.handle(mockRedirectUrl())
+      ).rejects.toThrow();
+    });
+
+    it("throws if the redirect URL is missing from storage", async () => {
+      mockOidcClient();
+      mockFetch(
+        new Response("", {
+          status: 200,
+        })
+      );
+      const storageUtility = mockDefaultStorageUtility();
+      await storageUtility.deleteForUser("mySession", "redirectUrl");
+      const authCodeRedirectHandler = getAuthCodeRedirectHandler({
+        storageUtility,
+      });
+
+      await expect(
+        authCodeRedirectHandler.handle(mockRedirectUrl())
+      ).rejects.toThrow();
+    });
+
     it("Fails if a session was not retrieved", async () => {
       mockOidcClient();
       mockFetch(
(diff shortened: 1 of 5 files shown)


packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -241,15 +231,44 @@ function getAuthCodeRedirectHandler(
...
 describe("AuthCodeRedirectHandler", () => {
   describe("canHandle", () => {
     it("Accepts a valid url with the correct query", async () => {
       const authCodeRedirectHandler = getAuthCodeRedirectHandler();
-      expect(
-        await authCodeRedirectHandler.canHandle(
-          "https://coolparty.com/?code=someCode&state=oauth2_state_value"
-        )
-      ).toBe(true);
+      expect(await authCodeRedirectHandler.canHandle(mockRedirectUrl())).toBe(
```
ThisIsMissEm, line 269: it may be better to use:
```suggestion
      expect(authCodeRedirectHandler.canHandle(mockRedirectUrl())).resolves.toBe(true)
```

In case the handler throws, that way you'll not bubble the exception / rethrow it via `await`.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -553,42 +545,60 @@ describe("AuthCodeRedirectHandler", () => {
...
       expect(sessionInfo.expirationDate).toBeNull();
     });
+
+    it("clears the oidc-client specific session information", async () => {
+      mockOidcClient();
+      mockLocalStorage({
+        // This mocks how oidc-client stores session information
+        "oidc.oauth2StateValue": "some arbitrary value",
+      });
+      const mockedFetch = jest.fn(global.fetch).mockReturnValue(
```
ThisIsMissEm, line 581: Can you explain the mocked fetch here, is this due to solid-oidc-ext?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -146,44 +147,44 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+      this.storageUtility,
+      this.issuerConfigFetcher
+    );
 
-    // Store the current session ID specifically in 'localStorage' (i.e., not using
-    // any other storage mechanism), as we don't deem this information to be
-    // sensitive, and we want to ensure it survives a browser tab refresh.
-    window.localStorage.setItem(KEY_CURRENT_SESSION, storedSessionId);
+    if (codeVerifier === undefined) {
+      throw new Error(
```
ThisIsMissEm, line 163: Will these be caught by the SDK?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -146,44 +147,44 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
+    if (storedRedirectIri === undefined) {
+      throw new Error(
+        `The redirect URL for session ${storedSessionId} is missing from storage.`
+      );
+    }
 
-    const issuerConfig = await this.issuerConfigFetcher.fetchConfig(issuer);
     const client: IClient = await this.clientRegistrar.getClient(
       { sessionId: storedSessionId },
       issuerConfig
```
ThisIsMissEm, line 176: Can we absolutely trust the `issuerConfig` in local storage not to have been tampered with / could this open a security attack vector? (tamper with that value, user gets taken somewhere that's not their OIDC provider, or something similar?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -146,44 +147,44 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
     );
 
+    // Store the current session ID specifically in 'localStorage' (i.e., not using
+    // any other storage mechanism), as we don't deem this information to be
+    // sensitive, and we want to ensure it survives a browser tab refresh.
+    window.localStorage.setItem(KEY_CURRENT_SESSION, storedSessionId);
+
     let tokens: CodeExchangeResult;
     const referenceTime = Date.now();
```
ThisIsMissEm, line 185: was a little confused reading this as "referenceTime", perhaps "tokenCreatedAt" would be better?

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -146,44 +147,44 @@ export class AuthCodeRedirectHandler implements IRedirectHandler {
...
-    // any other storage mechanism), as we don't deem this information to be
-    // sensitive, and we want to ensure it survives a browser tab refresh.
-    window.localStorage.setItem(KEY_CURRENT_SESSION, storedSessionId);
+    if (codeVerifier === undefined) {
+      throw new Error(
+        `The code verifier for session ${storedSessionId} is missing from storage.`
+      );
+    }
+
+    if (storedRedirectIri === undefined) {
```
ThisIsMissEm, line 168: I think we may also need to assert that this is a relative IRI to the current host, such that the redirect IRI can't be tampered with and take you off to `some.evil.host`

ThisIsMissEm (commented):
Looking good so far, a few minor things & questions otherwise

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -241,15 +231,44 @@ function getAuthCodeRedirectHandler(
...
+const DEFAULT_CODE_VERIFIER = "some code verifier";
+const DEFAULT_REDIRECT_URL = "https://some.redirect.uri";
+const DEFAULT_OAUTH_STATE = "oauth2StateValue";
+
+const mockRedirectUrl = () => {
+  const redirectUrl = new URL(DEFAULT_REDIRECT_URL);
+  redirectUrl.searchParams.append("state", DEFAULT_OAUTH_STATE);
+  redirectUrl.searchParams.append("code", DEFAULT_AUTHORIZATION_CODE);
+  return redirectUrl.href;
+};
```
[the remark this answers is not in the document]
NSeydoux: The behaviour on the error redirect flow is going to be different: it's not only the PKCE verifier that should be cleared, but all of the session information, which is done through logout. The cleanup in https://github.com/inrupt/solid-client-authn-js/pull/2002 should cover this.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
NSeydoux: If `window.fetch` isn't mocked, ```
 const authFetch = await buildAuthenticatedFetch(fetch, tokens.accessToken, {
      dpopKey: tokens.dpopKey,
      refreshOptions,
      eventEmitter,
      expiresIn: tokens.expiresIn,
    });
```
throws. I cleaned up the mock itself though.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: They should, but I think it would be cleaner to dedicate a complete PR to error management, because that's not the only error that should be caught at a higher level.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: If it has been tampered with, the subsequent token exchange will fail, because the redirect URL provided as part of the token request must match the redirect URL initially provided during the authorization request, so I don't think we need to add anything specific here.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Technically no, we cannot trust blindly the values in local storage. What we could do (I believe we talked about this recently) is to hash the session information in storage before the redirect, send it in the `state`, and use it to verify non-tampering when redirected back. I'll create a dedicated ticket.

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: 7c08f14da0c494092d1e1424b3fb0b9d9c13bf8d

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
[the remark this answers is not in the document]
ThisIsMissEm: When I initially read this I misread as testing all of the auth handler, but still a test that verifies that a `redirectUrl` without `state` and `code` does not trigger this handler should be there (I'm not sure if it exists)

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.spec.ts
ThisIsMissEm: 👍

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
ThisIsMissEm: Yeah, agree there, I just wanted to check as I wasn't sure

packages/browser/src/login/oidc/redirectHandler/AuthCodeRedirectHandler.ts
ThisIsMissEm: Right, but couldn't the malicious actor pretend to be the AS? i.e., client -> malicious -> AS such that the malicious actor gets back the token?

NSeydoux: The attack vector in this case would be modifying the `issuerConfig` as you rightly suggested in https://github.com/inrupt/solid-client-authn-js/pull/2009#discussion_r819024793, or modifying the issuer referenced in the user session information. The `storedRedirectIri`is the IRI to which the user should have been redirected already: it's the client-controlled URL where the auth code is returned. It is only verified by the OIDC provider, but at this point the user is not redirected to this URL.

ThisIsMissEm (approved):
I think this is fine, though it's quite a context switch for me, so I'd suggest having someone else review too if critical

matthieubosquet (approved):
Ditto to @ThisIsMissEm, there's a lot of context to this PR I don't know and to give it a full in-depth review I would need to familiarize myself with the library.

But deleting the `codeVerifier` from `StorageUtility` and loading the whole OIDC context from storage seems to make sense.

PR #2084, opened 2022-04-19
Replace legacy client profile

NSeydoux:
The latest Solid-OIDC version requires the document to be JSON-LD, rather than the legacy turtle serialization. Note that the client ID will be valid only after this PR gets merged.


ThisIsMissEm (approved):
LGTM

PR #2089, opened 2022-04-20
Refresh client identifier

NSeydoux:
This fixes a bug affecting client-side apps using a Solid-OIDC client identifier: the client identifier not being included in the refresh request, the session expired at the end of the lifetime of the initial access token, so somewhere around 10 minutes. Fixing this will allow the background refresh to happen, and the session to be maintained alive.

### DIFF
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -31,6 +31,9 @@ import {
 } from "@inrupt/solid-client-authn-browser";
 
 const REDIRECT_URL = "http://localhost:3113/";
+// This is the IRI where the Client identifier document (i.e. client-app-profile.jsonld)
+// is available to the OIDC issuer. See https://solid.github.io/solid-oidc/#clientids-document
+// for more information.
 const CLIENT_IDENTIFIER =
   "https://raw.githubusercontent.com/inrupt/solid-client-authn-js/main/packages/browser/examples/single/bundle/client-app-profile.jsonld";
 
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -31,7 +31,8 @@ import {
 } from "@inrupt/solid-client-authn-browser";
 
 const REDIRECT_URL = "http://localhost:3113/";
-const CLIENT_IDENTIFIER = "https://zwifi.eu/client-id.json";
+const CLIENT_IDENTIFIER =
+  "https://raw.githubusercontent.com/inrupt/solid-client-authn-js/main/packages/browser/examples/single/bundle/client-app-profile.jsonld";
 
 export default function App() {
   const [webId, setWebId] = useState(getDefaultSession().info.webId);
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -28,11 +28,10 @@ import {
   handleIncomingRedirect,
   fetch,
   getDefaultSession,
-} from "../../../../dist/index";
+} from "@inrupt/solid-client-authn-browser";
 
 const REDIRECT_URL = "http://localhost:3113/";
-const CLIENT_IDENTIFIER =
-  "https://raw.githubusercontent.com/inrupt/solid-client-authn-js/main/packages/browser/examples/single/bundle/client-app-profile.jsonld";
+const CLIENT_IDENTIFIER = "https://zwifi.eu/client-id.json";
 
 export default function App() {
   const [webId, setWebId] = useState(getDefaultSession().info.webId);
--- a/packages/oidc/src/__mocks__/issuer.mocks.ts
+++ b/packages/oidc/src/__mocks__/issuer.mocks.ts
@@ -151,7 +151,7 @@ export const mockDpopTokens = (): TokenEndpointRawResponse => {
   };
 };
 
-export const mockClient = (clientId: string = "some client"): IClient => {
+export const mockClient = (clientId = "some client"): IClient => {
   return {
     clientId,
     clientType: "dynamic",
--- a/packages/oidc/src/__mocks__/issuer.mocks.ts
+++ b/packages/oidc/src/__mocks__/issuer.mocks.ts
@@ -151,9 +151,9 @@ export const mockDpopTokens = (): TokenEndpointRawResponse => {
   };
 };
 
-export const mockClient = (): IClient => {
+export const mockClient = (clientId: string = "some client"): IClient => {
   return {
-    clientId: "some client",
+    clientId,
     clientType: "dynamic",
   };
 };
--- a/packages/oidc/src/refresh/refreshGrant.spec.ts
+++ b/packages/oidc/src/refresh/refreshGrant.spec.ts
@@ -19,6 +19,7 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import { IClient } from "@inrupt/solid-client-authn-core";
 import { jest, it, describe, expect } from "@jest/globals";
 import { jwtVerify, importJWK } from "jose";
 import {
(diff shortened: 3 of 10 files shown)


packages/browser/examples/single/bundle/src/App.js
```diff
@@ -28,11 +28,11 @@ import {
   handleIncomingRedirect,
   fetch,
   getDefaultSession,
-} from "../../../../dist/index";
+} from "@inrupt/solid-client-authn-browser";
 
 const REDIRECT_URL = "http://localhost:3113/";
-const CLIENT_APP_WEBID =
-  "https://raw.githubusercontent.com/inrupt/solid-client-authn-js/main/packages/browser/examples/single/bundle/client-app-profile.ttl#app";
+const CLIENT_IDENTIFIER =
```
ThisIsMissEm, line 34: Might be worth documenting this a bit

packages/oidc/src/refresh/refreshGrant.spec.ts
```diff
@@ -63,16 +63,20 @@ describe("refreshGrant", () => {
     );
   });
 
-  it("does not use basic auth if no client secret is available", async () => {
+  it("include the client id in the body if no client secret is available", async () => {
```
ThisIsMissEm, line 66: Make sure you also test when client_id is not an IRI, e.g., someone configured using client credentials but forgot to set client_secret's value; in that case I don't think client_id should be sent in the body.

packages/oidc/src/refresh/refreshGrant.spec.ts
```diff
@@ -63,16 +63,20 @@ describe("refreshGrant", () => {
...
     const myFetch = mockFetch(JSON.stringify(mockBearerTokens()), 200);
     const client = mockClient();
 
     await refresh("some refresh token", mockIssuer(), client);
     expect(myFetch.mock.calls[0][0]).toBe(
```
ThisIsMissEm, line 71: It would be a good idea to store `myFetch.mock.calls[0]` in a constant, refreshFetchArgs, and then reuse that in subsequent assertions; when I initially read this I thought lines 75 & 79 tested on a different fetch call, but then I remembered how mock captures args

packages/oidc/src/refresh/refreshGrant.ts
```diff
@@ -26,11 +26,22 @@ import {
...
 import formUrlEncoded from "form-urlencoded";
 import { validateTokenEndpointResponse } from "../dpop/tokenExchange";
 
+// Camelcase identifiers are required in the OIDC specification.
+/* eslint-disable camelcase */
+
+interface IRefreshRequestBody {
+  grant_type: "refresh_token";
+  refresh_token: string;
+  scope: typeof DEFAULT_SCOPES;
```
ThisIsMissEm, line 41: I think this is too strict? i.e., should it be possible to refresh with explicitly different scopes?

packages/oidc/src/refresh/refreshGrant.ts
```diff
@@ -62,6 +73,11 @@ export async function refresh(
         `${client.clientId}:${client.clientSecret}`
       )}`,
     };
+  } else {
```
ThisIsMissEm, line 76: As mentioned earlier, I think we need to check if the client_id is an IRI here

ThisIsMissEm (changes requested):
Looks good, but have a concern about not validating client_id as a client identifier (IRI) before using it as one

packages/oidc/src/refresh/refreshGrant.ts
NSeydoux: Currently we're not giving any flexibility to control the scopes, so I don't think we need to be less strict here. Since this is only internal, it will be easy to change if required later.

packages/browser/examples/single/bundle/src/App.js
NSeydoux: 14f6fc32dfd73e4562f67811e7e42011b051305f

packages/oidc/src/refresh/refreshGrant.spec.ts
NSeydoux: d218289054cd254c39a7c8c7d67cc568d2b5ecd1

packages/oidc/src/refresh/refreshGrant.spec.ts
NSeydoux: d218289054cd254c39a7c8c7d67cc568d2b5ecd1

packages/oidc/src/__mocks__/issuer.mocks.ts
```diff
@@ -151,9 +151,9 @@ export const mockDpopTokens = (): TokenEndpointRawResponse => {
   };
 };
 
-export const mockClient = (): IClient => {
+export const mockClient = (clientId = "some client"): IClient => {
   return {
-    clientId: "some client",
+    clientId,
```
ThisIsMissEm, line 156: I thought both client_id and client_secret were generated for dynamic client registration? So would this really be client_name instead?

packages/oidc/src/refresh/refreshGrant.ts
```diff
@@ -62,6 +85,11 @@ export async function refresh(
         `${client.clientId}:${client.clientSecret}`
       )}`,
     };
+  } else if (client.clientId !== undefined && isValidUrl(client.clientId)) {
```
ThisIsMissEm, line 88: Should line 80 also do:
```
if (client.clientId !== undefined && client.clientSecret !== undefined) {
```

ThisIsMissEm (approved):

packages/oidc/src/__mocks__/issuer.mocks.ts
NSeydoux: That's correct, the client credentials are generated as part of the dynamic registration flow. However, in the case of the refresh grant, the credentials would have been generated previously during the initial authorization code flow, so here these credentials are just passed from storage. This mock default is just an arbitrary value for an identifier that could have been generated, but could also have been statically registered. Basically at this point in the grant that's irrelevant: either you have both an ID and a secret and you use regular OIDC client authentication (the `Authorization <client id>:<client secret>` header, or you just have an IRI-shaped ID and then you do Solid-OIDC auth.

packages/oidc/src/refresh/refreshGrant.ts
NSeydoux: I approached this somewhat differently with f57f1282: the absence of a client_id is not something we support, so we might as well throw an error. The issue should actually be caught when looking up the credentials from storage, which is why there was initially no check to that intent.

packages/oidc/src/__mocks__/issuer.mocks.ts
ThisIsMissEm: Ooooh, right! Perhaps labelling client_id as "previously issued client_id" or something would make it clearer?

packages/oidc/src/refresh/refreshGrant.spec.ts
```diff
@@ -114,6 +115,17 @@ describe("refreshGrant", () => {
     expect(dpopProof.payload.htu).toBe(mockIssuer().tokenEndpoint.toString());
   });
 
+  it("throws if the client identifier is undefined", async () => {
+    const client = mockClient();
```
ThisIsMissEm: @NSeydoux linting issue here

packages/oidc/src/refresh/refreshGrant.ts
```diff
@@ -76,6 +76,12 @@ export async function refresh(
     };
   }
 
+  if (client.clientId === undefined) {
```
ThisIsMissEm, line 79: Let's move this up to line 71-ish, no need to generate DPoP header if we're just gonna throw (though do note that may change from an async throw to a sync throw potentially)

packages/oidc/src/refresh/refreshGrant.spec.ts
NSeydoux: Yeah I really have to setup my IDE to catch these in the absense of precommit hooks :)

packages/oidc/src/refresh/refreshGrant.ts
NSeydoux: Good point, I actually moved it all the way to the top of the function for clarity

PR #2096, opened 2022-04-22
chore: remove unnecessary form-urlencoded dependency

ThisIsMissEm:
# New feature description

Since node.js v10 `URLSearchParams` is a [global](https://nodejs.org/dist/latest-v16.x/docs/api/url.html#class-urlsearchparams), and it is also a global in all [browser environments](https://caniuse.com/urlsearchparams) that we support, therefore, we don't need to depend on an external package for this.

# Checklist

# Questions

- Is this changelog worthy?

### DIFF
--- a/packages/oidc/src/dpop/tokenExchange.ts
+++ b/packages/oidc/src/dpop/tokenExchange.ts
@@ -19,7 +19,6 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import formurlencoded from "form-urlencoded";
 import { OidcClient } from "@inrupt/oidc-client";
 import {
   IClient,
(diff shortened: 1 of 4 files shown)


packages/oidc/src/refresh/refreshGrant.ts
```diff
@@ -27,8 +27,8 @@ import {
   KeyPair,
   TokenEndpointResponse,
 } from "@inrupt/solid-client-authn-core";
+
 // NB: once this is rebased on #1560, change dependency to core package.
```
ThisIsMissEm, line 31: @NSeydoux you left this comment, but I'm not sure what this was about?

packages/oidc/src/dpop/tokenExchange.ts
```diff
@@ -222,20 +221,22 @@ export async function getTokens(
     )}`;
   }
 
+  const requestBody = {
```
ThisIsMissEm, line 224: This follows the same style as in `refreshGrant.ts` where the body is constructed separately before being passed to `new URLSearchParams`

packages/oidc/src/refresh/refreshGrant.ts
NSeydoux: I'm not sure either to be honest, this probably can be removed

PR #2097, opened 2022-04-23
WIP: hoist devDependencies and migrate to workspaces + rollup

ThisIsMissEm:
This is a highly experimental branch, where in I'm attempting to address some long standing issues with how we build the code in this repo. The commits are not minimal, there's a lot of cruft that needs to be cleaned up, but it's a first working version.

One note for Next.js: If you're using `@inrupt/solid-ui-react` and encounter an error when loading the SessionProvider (or other components), and the error is about window being undefined, well, it turns out that's kinda a next.js quirk that webpack was silently allowing to work somehow. We do need to "fix" the SDK to not make direct use of the `window` global object, but that's additional work that I'm trying to keep out of scope.

If you wanna test in next.js you'll need to do the following instead of just `import { SessionProvider } from "@inrupt/solid-ui-react"`:

```
const SessionProvider = dynamic(
  () => import("@inrupt/solid-ui-react").then((mod) => mod.SessionProvider),
  {
    ssr: false,
  }
);
```

Additionally, there's some package json changes needed to properly tie it all together:

```
    // New dependency versions
    "@inrupt/solid-client-authn-browser": "1.11.9-alpha.5",
    "@inrupt/solid-ui-react": "^2.8.0-0795c00b",

  // Overriding the version of authn-browser in solid-ui-react:
"overrides": {
    "@inrupt/solid-ui-react@^2.8.0-0795c00b": {
      "@inrupt/solid-client-authn-core": "^1.11.9-alpha.5",
      "@inrupt/solid-client-authn-browser": "^1.11.9-alpha.5"
    }
  }
```


package.json
```diff
@@ -1,73 +1,85 @@
...
-    "webpack-cli": "^4.1.0",
-    "webpack-merge": "^5.2.0"
-  },
-  "lint-staged": {
-    "*.{ts,js,css,json}": [
-      "prettier --write"
-    ]
+    "webpack": "^5.11.0",
+    "webpack-cli": "^4.2.0",
+    "webpack-merge": "^5.7.2"
```
ThisIsMissEm, line 75: Webpack is only used for the example apps now, but we can and probably should migrate to something else there (swc, next, rollup, etc)

package.json
```diff
@@ -1,73 +1,85 @@
...
   },
   "optionalDependencies": {
     "fsevents": "^2.3.2"
+  },
+  "overrides": {
+    "@inrupt/eslint-config-base@0.4.0": {
+      "eslint-plugin-jest": "^26.0.0"
```
ThisIsMissEm, line 82: this is needed until we publish `@inrupt/eslint-config-base@1.0.0`, as `v0.4.0` / `v0.4.1` do not use Jest 27

packages/browser/package.json
```diff
@@ -1,59 +1,45 @@
 {
   "name": "@inrupt/solid-client-authn-browser",
-  "version": "1.11.7",
+  "version": "1.11.9-alpha.5",
   "license": "MIT",
-  "types": "dist/index",
```
ThisIsMissEm, line 5: pretty sure this was a bug

ThisIsMissEm:
@NSeydoux let's figure out a plan for the work that's been done here, most is already done elsewhere now, so really there's just the webpack -> rollup migration left.

PR #2119, opened 2022-04-29
refactor: cleanup handlers and mocks

ThisIsMissEm:
# Description

This cleans up the internals of the packages a bit, making it cleaner when we're using the Redirector vs IncomingRedirectHandler, simplifies our mock usage, removes unnecessary handlers (WebID-TLS, AggregateLoginHandler), and starts adding a `@inrupt/solid-client-authn-core/mocks` module (though it's unused right now until we upgrade to Jest 28, as 27 doesn't support `exports` from package.json)

# Checklist


package.json
```diff
@@ -17,7 +17,7 @@
     "prepublishOnly": "npm run build",
     "publish": "lerna publish",
     "publish-preview": "lerna publish from-package",
-    "test": "jest --coverage --verbose",
+    "test": "lerna run --no-bail --ignore e2e-browser test",
```
ThisIsMissEm, line 20: just running `jest --coverage --verbose` like this means that the mocks from browser clobber those from node and vice-versa; Instead we're using lerna to run the all tests in isolation, if one package fails in it's tests, it'll keep going, but report the correct error exit code.

packages/browser/src/ClientAuthentication.spec.ts
```diff
@@ -21,20 +21,26 @@
...
 import { jest, it, describe, expect } from "@jest/globals";
 import {
-  mockStorageUtility,
   StorageUtility,
   USER_SESSION_PREFIX,
-  mockStorage,
 } from "@inrupt/solid-client-authn-core";
+
+// FIXME: For some reason jest crashes on trying to handle a subpath import
+// this should import from @inrupt/solid-client-authn-core/mocks
```
ThisIsMissEm, line 29: These will be resolved when we upgrade to Jest 28.

PR #2121, opened 2022-04-30
Improve handlers & ensure oauth params are always stripped from redirect URLs in browser

ThisIsMissEm:
# Description

This improves our tests where we were using the MockRedirector (now mockRedirector) to be like our other tests, adds test coverage to the `console.error` in `AuthorizationCodeWithPkceOidcHandler` for browsers, fixes a minor imports issue in the StorageUtility mock, and ensures that OAuth Parameters are always stripped from redirect URLs.

Previously in error conditions, it was possible that the user would be left with bad OAuth redirect parameters to be left in the URL, which usually resulted in the user getting stuck at an error if they reloaded the page.

# Checklist

### DIFF
--- a/packages/browser/src/ClientAuthentication.spec.ts
+++ b/packages/browser/src/ClientAuthentication.spec.ts
@@ -19,13 +19,14 @@
  * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-import { jest, it, describe, expect } from "@jest/globals";
+import { jest, it, describe, expect, beforeEach } from "@jest/globals";
 import { EventEmitter } from "events";
 import * as SolidClientAuthnCore from "@inrupt/solid-client-authn-core";
 
 import {
   StorageUtility,
   USER_SESSION_PREFIX,
+  EVENTS,
 } from "@inrupt/solid-client-authn-core";
 
 // FIXME: For some reason jest crashes on trying to handle a subpath import
@@ -34,6 +35,7 @@ import {
   mockStorageUtility,
   mockStorage,
   mockIncomingRedirectHandler,
+  mockHandleIncomingRedirect,
 } from "../../core/src/mocks";
 
 import { mockLoginHandler } from "./login/__mocks__/LoginHandler";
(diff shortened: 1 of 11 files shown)


packages/browser/src/ClientAuthentication.ts
```diff
@@ -129,13 +130,39 @@ export default class ClientAuthentication {
...
+      // I'm not sure if "redirect" is a good error code, and in theory `err`
+      // maybe an Error object and not a string; Maybe we want to just hardcode
+      // a description instead?
+      eventEmitter.emit(EVENTS.ERROR, "redirect", err);
+
+      return undefined;
+    }
+  };
 
+  private cleanUrlAfterRedirect(url: string): void {
```
ThisIsMissEm, line 165: This might be better moved to `removeOidcQueryParam` from oidc, but currently that's not how it worked.

packages/browser/src/ClientAuthentication.spec.ts
```diff
@@ -97,8 +107,44 @@ describe("ClientAuthentication", () => {
...
+      // redirected back to, since we don't pass redirectUrl:
+      window.location.href = "https://coolapp.test/some/redirect";
+
+      const clientAuthn = getClientAuthentication();
+      await clientAuthn.login(
+        {
+          sessionId: "mySession",
+          tokenType: "DPoP",
+          clientId: "coolApp",
+          oidcIssuer: "https://idp.com",
```
ajacksified, line 134: As this is a real url, let's stick with `test.url` here and elsewhere.

ThisIsMissEm: This is probably a much wider issue to resolve? I think we should add a ticket for this tech debt across the SDKs

NSeydoux (approved):

packages/browser/src/ClientAuthentication.ts
```diff
@@ -129,13 +130,39 @@ export default class ClientAuthentication {
...
+      };
+    } catch (err) {
+      // Strip the oauth params:
+      this.cleanUrlAfterRedirect(url);
+
+      // FIXME: EVENTS.ERROR should be errorCode, errorDescription
+      //
+      // I'm not sure if "redirect" is a good error code, and in theory `err`
+      // maybe an Error object and not a string; Maybe we want to just hardcode
+      // a description instead?
```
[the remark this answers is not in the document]
ThisIsMissEm: I think that's something to approach in a general refactor of this code, not something that's for right now; see also removeOidcParams

PR #2122, opened 2022-04-30
[WIP] fix: correctly respect client expiry

ThisIsMissEm:
This PR fixes the bug where dynamically registered clients did not expire, despite the server indicating that they should expiry at a given time.


packages/browser/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -193,6 +195,94 @@ describe("ClientRegistrar", () => {
...
+        {
+          ...IssuerConfigFetcherFetchConfigResponse,
+          registrationEndpoint: "https://some.issuer/register",
+        }
+      );
+
+      expect(mockFetch).toHaveBeenCalled();
+    });
+
+    it("registers a new client if the one in storage has no expiry stored", async () => {
```
ThisIsMissEm, line 244: @NSeydoux I think this is what we'll want, because otherwise someone may have a valid client, but we hadn't stored the expiry, in which case registering a new client resolves that missing expiry.

packages/node/src/login/oidc/ClientRegistrar.ts
```diff
@@ -143,18 +149,30 @@ export default class ClientRegistrar implements IClientRegistrar {
...
+    }
+
     const infoToSave: Record<string, string> = {
       clientId: registeredClient.metadata.client_id,
       idTokenSignedResponseAlg:
         registeredClient.metadata.id_token_signed_response_alg ?? signingAlg,
     };
     if (registeredClient.metadata.client_secret) {
       infoToSave.clientSecret = registeredClient.metadata.client_secret;
+      infoToSave.clientExpiresAt = clientExpiresAt.toString();
```
ThisIsMissEm, line 168: I think we may always want to store this, though technically `client_secret_expires_at` is only returned if a `client_secret` is returned: https://openid.net/specs/openid-connect-registration-1_0.html#RegistrationResponse

packages/node/src/login/oidc/ClientRegistrar.ts
```diff
@@ -143,18 +149,30 @@ export default class ClientRegistrar implements IClientRegistrar {
       }
     );
 
+    let clientExpiresAt = 0;
+    if (
+      typeof registeredClient.metadata.client_secret_expires_at === "number"
```
ThisIsMissEm, line 154: Note: `openid-client` doesn't actually verify that `client_secret_expires_at` is present and that it is in the future, I've added a set of checks in oidc-client-ext to verify these for the browser, but for server we don't currently have these checks.

ThisIsMissEm: In theory here we could receive back a client with a `client_secret_expires_at` in the past, if the users' clock time on their server is ahead of the IdP's clock time.

Can we assume that users and IdP's can configure their servers to have a synchronised clock time?

packages/browser/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -276,12 +368,18 @@ describe("ClientRegistrar", () => {
...
+      jest.setSystemTime(currentTime);
+
       const mockFetch = (jest.fn() as any).mockResolvedValueOnce(
         /* eslint-disable camelcase */
         new NodeResponse(
           JSON.stringify({
             client_id: "some id",
             client_secret: "some secret",
+            // Expire this client in 5 seconds:
+            client_secret_expires_at: currentTime / 1000 + 5,
```
ThisIsMissEm, line 382: We may want a test for the behaviour when `client_secret_expires_at` is in the past, as the `registerClient` would fail in this case, which would likely blow up the code for ClientRegistrar in browser.

ThisIsMissEm:
E2E tests are currently failing because of how the `silentlyAuthenticate` function works, which results in the existing client being wiped out when we try to authenticate again.

packages/browser/src/ClientAuthentication.ts
```diff
@@ -112,17 +113,21 @@ export default class ClientAuthentication {
...
     if (
       sessionInfo === undefined ||
-      sessionInfo.clientAppId === undefined ||
+      // sessionInfo.clientAppId === undefined ||
       sessionInfo.issuer === undefined
     ) {
       return null;
     }
+
+    // FIXME: verify client is still valid
```
ThisIsMissEm, line 129: This is actually taken care of by ClientRegistrar — if the client isn't valid, then a new one will be requested.

Valid Client: has clientId & clientExpiresAt == 0 or clientExpiresAt > now

packages/browser/src/ClientAuthentication.ts
```diff
@@ -112,17 +113,16 @@ export default class ClientAuthentication {
...
     currentSessionId: string
   ): Promise<(ISessionInfo & ISessionInternalInfo) | null> => {
     const sessionInfo = await this.sessionInfoManager.get(currentSessionId);
-    if (
-      sessionInfo === undefined ||
-      sessionInfo.clientAppId === undefined ||
-      sessionInfo.issuer === undefined
-    ) {
+
+    if (sessionInfo === undefined) {
```
ThisIsMissEm, line 122: sessionInfo is undefined if the issuer is undefined, and clientAppId is no longer used here (see splitting sessionInfoManager and ClientRegistrar)

packages/browser/src/login/oidc/ClientRegistrar.ts
```diff
@@ -44,67 +61,99 @@ export default class ClientRegistrar implements IClientRegistrar {
...
+      ]);
+
+    // Handle public client identifiers:
+    if (typeof storedClientId === "string" && isValidUrl(storedClientId)) {
       return {
         clientId: storedClientId,
-        clientSecret: storedClientSecret,
-        clientType: "dynamic",
+        clientExpiresAt: 0,
+        clientType: "solid-oidc",
```
ThisIsMissEm, line 82: No more converting a solid-oidc client to a dynamic client on page reload; now it always stays as a solid-oidc client.

packages/browser/src/login/oidc/ClientRegistrar.ts
```diff
@@ -44,67 +61,99 @@ export default class ClientRegistrar implements IClientRegistrar {
...
       };
     }
+
+    // Handle dynamically registered clients & client credentials:
```
ThisIsMissEm, line 86: theoretically "client credentials" though we haven't implemented client credentials flow in browser.

packages/browser/src/sessionInfo/SessionInfoManager.ts
```diff
@@ -159,35 +190,63 @@ export class SessionInfoManager implements ISessionInfoManager {
...
       throw new Error(`Tokens of type [${tokenType}] are not supported.`);
     }
 
-    if (
-      clientId === undefined &&
-      isLoggedIn === undefined &&
-      webId === undefined &&
-      refreshToken === undefined
-    ) {
+    if (issuer === undefined) {
```
ThisIsMissEm, line 209: Not entirely sure this logic is correct here / below

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -111,26 +120,26 @@ export async function handleRegistration(
   // or it is not compliant but the client_id isn't an IRI (we assume it has already
   // been registered with the IdP), then the client registration information needs
   // to be stored so that it can be retrieved later after redirect.
-  await storageUtility.setForUser(options.sessionId, {
+  const clientDetails: IClientDetails = {
```
ThisIsMissEm, line 123: We should probably revisit this, because I don't think it necessarily makes sense given this is only for solid-oidc or static clients both of which don't expire, as far as I'm aware. We also don't use the stored `clientName` value anywhere later, iirc.

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -98,7 +107,9 @@ export async function handleRegistration(
 ): Promise<IClient> {
   const clientType = determineClientType(options, issuerConfig);
   if (clientType === "dynamic") {
-    return clientRegistrar.getClient(
+    // Added because I was missing stacktrace frames otherwise:
+    // eslint-disable-next-line @typescript-eslint/return-await
+    return await clientRegistrar.getClient(
```
ThisIsMissEm, line 112: Rename to `getDynamicClient`

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -98,7 +107,9 @@ export async function handleRegistration(
 ): Promise<IClient> {
   const clientType = determineClientType(options, issuerConfig);
   if (clientType === "dynamic") {
-    return clientRegistrar.getClient(
+    // Added because I was missing stacktrace frames otherwise:
```
ThisIsMissEm, line 110: link article about return await from slack

packages/core/src/login/oidc/IClientRegistrar.ts
ThisIsMissEm: or rename clientRegistrar to dynamicClientRegistrar

NSeydoux:
This has been fixed with https://github.com/inrupt/solid-client-authn-js/commit/1d718fa8ccd97d9b348af18aea06001c34fd7bc8.

PR #2159, opened 2022-05-25
Remove reference to window.localstorage from Session constructor

ajacksified:
A temporary fix was put into place in 3fd7bb44de656bf044e9b09bb3f29fcada5e8de4 which cleared a localstorage value set by a temporary ESS session restoration workflow. This caused window to be accessed in the session constructor, which caused issues with server-side rendering.

The localstorage value no longer needs to be cleared as it is unused, so this line has been removed.

This PR fixes bug #.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The following changes have been implemented but not released yet:
 
 ### Bugfixes
 
--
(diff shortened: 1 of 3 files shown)


matthieubosquet (approved):
LGTM

PR #2182, opened 2022-06-17
Integrate Sonar analysis with test coverage on PRs into CI

edwardsph:
This extends the CI workflow to run a SonarCloud Scan on each PR, including unit tests and the coverage report. The results are added to SonarCloud. Since this is a monorepo there are 4 separate projects registered in SonarCloud and there are separate scans for each of the packages.


packages/core/sonar-project.properties
```diff
@@ -0,0 +1,16 @@
+sonar.projectKey=inrupt_solid-client-authn-core
+sonar.organization=inrupt
```
ThisIsMissEm, line 2: Do we need projectName here too?

packages/node/sonar-project.properties
```diff
@@ -0,0 +1,16 @@
+sonar.projectKey=inrupt_solid-client-authn-node
+sonar.organization=inrupt
```
ThisIsMissEm, line 2: Ditto re projectName

.github/workflows/ci.yml
```diff
@@ -18,3 +18,42 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run test
+
+      - run: git fetch --unshallow
+        if: ${{ matrix.node-version == '16' }}
+
+      - name: SonarCloud Scan - Browser
+        if: ${{ matrix.node-version == '16' }}
+        uses: SonarSource/sonarcloud-github-action@v1.6
```
ThisIsMissEm, line 27: It might be an idea to see if we can contribute tooling to this action to handle multiple projects at once, I'm not a huge fan of this style of one per package, as if we add additional packages it can create drift.

.github/workflows/ci.yml
```diff
@@ -18,3 +18,42 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run test
+
+      - run: git fetch --unshallow
```
ThisIsMissEm, line 22: Could you add a comment explaining why this is necessary?

ThisIsMissEm (changes requested):
Could do with some additional comments explaining why certain things are done. Also, it's preferable to have comments on the line above instead of at the end of line, as to not run afoul of line length limits.

packages/core/sonar-project.properties
edwardsph: Not necessary but I will add it for consistency in this project. It was only required in one place to fix the incorrect name that was used when the project was first imported.

.github/workflows/ci.yml
edwardsph: Will do

ThisIsMissEm (approved):

PR #2186, opened 2022-06-24
Chore: upgrade eslint

ThisIsMissEm:
Just to kick the wheels on the new style configs; but it may make merging #2122 hard, so let's keep this as a draft for now, I think


ThisIsMissEm:
I'm confident that this has been covered by other pull requests

PR #2193, opened 2022-06-28
Prevent Sonar from running on dependabot PRs

NSeydoux:
Since Sonar requires access to environment secrets, it cannot run on dependabot PRs


ThisIsMissEm (approved):
I'm okay with this for now, but let's align with @edwardsph on this, I think using a dependabot secret could be the solution here

PR #2216, opened 2022-07-06
Fix parse-path vulnerability

NSeydoux:
Fix https://github.com/advisories/GHSA-3j8f-xvm3-ffx4


ThisIsMissEm (changes requested):
Let's just upgrade lerna instead: https://github.com/lerna/lerna/blob/main/CHANGELOG.md

ThisIsMissEm (approved):

PR #2239, opened 2022-07-18
Temporarily disable production e2e tests

NSeydoux:
Since the ESS-production environment is migrating from ESS1.1 to ESS2.0, the tests need to be upgraded accordingly. They are disabled in the meantime because they will start to break when the migration happens. They will be re-enabled when secret sharing is properly configured to reduce the manual work as much as possible.


ThisIsMissEm (approved):
A suggestion, but otherwise LGTM

PR #2242, opened 2022-07-20
chore: improve strictness of the engines supported

jeswr:


NSeydoux (commented):
Why has a package-lock been updated as part of this change?

ThisIsMissEm:
@NSeydoux it updates because the engines field id copied to the lockfile. I'll rebase tomorrow

packages/browser/package.json
```diff
@@ -27,10 +27,7 @@
     "build-docs-preview-site": "npm run build-api-docs; cd docs/api; make html"
   },
   "devDependencies": {
-    "cross-fetch": "^3.1.5",
-    "webpack": "^5.11.0",
-    "webpack-cli": "^4.2.0",
-    "webpack-merge": "^5.7.2"
```
ThisIsMissEm, line 33: @NSeydoux the duplication between here and the root `package.json` file was causing additional audit failures, despite your earlier fix. Removing these from here and upgrading the parent package's versions resolves the issue.

NSeydoux (approved):

PR #2248, opened 2022-07-21
Fix GHSA-4wf5-vphf-c2xc

NSeydoux:
Use npm audit fix to resolve https://github.com/advisories/GHSA-4wf5-vphf-c2xc


ThisIsMissEm (approved):
I'd seen an audit issue in webpack / terser too, and npm audit didn't fix (double check this didn't downgrade anything)

PR #2346, opened 2022-09-22
Update bootstrapped app to static registration

NSeydoux:
The bootstrapped app was still relying on the now deprecated oidc-token-helper. It now instead uses static client registration, pointing to Inrupt's documentation for details.

Also, this uses a file dependency instead of using a path in the import.


packages/node/examples/bootstrappedApp/package.json
```diff
@@ -18,7 +18,8 @@
   "homepage": "https://github.com/inrupt/solid-client-authn-js#readme",
   "dependencies": {
     "express": "^4.17.1",
-    "yargs": "^16.1.1"
+    "yargs": "^16.1.1",
+    "@inrupt/solid-client-authn-node": "file:../../dist/"
```
ThisIsMissEm, line 22: Not sure this should point to dist? Normally we just go parent directory?

ThisIsMissEm (approved):
LGTM

packages/node/examples/bootstrappedApp/package.json
NSeydoux: Oh right, I tried a bunch of stuff before realizing I was just off one `..` and didn't fix this.

PR #2365, opened 2022-10-04
Typedoc api docs fix

kay-kim:
Noticed that the API docs for authn wasn't building. Added patch to fix + additional patches in prep for dependency updates:

1. Patch 1: [add back typedoc-plugin-markdown dependency](https://github.com/inrupt/solid-client-authn-js/commit/0b2cdd372afa71bf8519394b479b29d2ea9f7296): This fixes the build. I think the dependency accidentally got dropped during https://github.com/inrupt/solid-client-authn-js/pull/2118/files.  
2. The other patches are in prep for dependency updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,6 +39,7 @@
         "ts-node": "^10.9.1",
         "tsc-watch": "^5.0.2",
         "typedoc": "^0.23.1",
+        "typedoc-plugin-markdown": "^3.13.3",
         "typescript": "^4.9.4",
         "webpack": "^5.73.0",
         "webpack-cli": "^4.10.0",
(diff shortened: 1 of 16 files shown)


NSeydoux (approved):

ThisIsMissEm:
@kay-kim yeah, this was known — we never migrated it to the new tooling as we wanted to merge the docs into one instead of multiple; because we never adopted the new tooling, they'd become broken (it's the same in the VC library)

ThisIsMissEm (commented):
The changes here look good, but it'd be best to make sure we're aligned on the overall direction of these docs & how they should build.

kay-kim:
Yup.  You all can decide when you all want to migrate to the new tooling and such, as these doc belong to you all.  I figured that until you all  schedule the new tooling work, it might be good not to have the docs be broken, in case we need to publish some updates and whatnot.

PR #2386, opened 2022-10-17
Use jsdom polyfill and update jest

NSeydoux:
Initially, the intent of this PR was just to use jsdom polyfills we provide to simplify the environment setup. However, it showed that we are currently not running our unit test using the browser jose bundle in the browser package, because we are using an older version of `jest`. So this PR also looks at updating jest to the latest version.

EDIT: a bunch of upgrades tacked on this PR: 
- Update to the latest linting rules (because of the polyfill update coming from the same repo)
- Migrated jest configs to TS to reuse explicit typing provided by `ts-jest`
- A bunch of tests required some changes because of the jest major update
- The polyfill still don't entirely cut it, so for now the custom environment is still used. Changes will be pushed to the polyfills, and when upgrading the custom environment will be removed.


jest.config.js
```diff
@@ -2,9 +2,8 @@ module.exports = {
   roots: ["<rootDir>"],
   clearMocks: true,
   testMatch: ["**/(__tests__|src)/**/?(*.)+(spec|test).+(ts|tsx|js)"],
-  transform: {
-    "^.+\\.(ts|tsx)?$": "ts-jest",
-  },
+  preset: "ts-jest/presets/default-esm",
+  extensionsToTreatAsEsm: ['.ts'],
```
ThisIsMissEm, line 6: Isn't this default?

package.json
```diff
@@ -42,17 +43,18 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-header": "^3.1.1",
     "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jest": "^26.0.0",
+    "eslint-plugin-jest": "^27.1.1",
     "eslint-plugin-prettier": "^4.0.0",
```
ThisIsMissEm, line 47: You can remove most, if not all of these, they now come from the shared config package.

packages/browser/jest.config.js
```diff
@@ -1,7 +1,8 @@
 module.exports = {
   ...require("../../jest.config"),
   // Allows running tests from packages/browser/
-  testEnvironment: "<rootDir>/../../tests/environment/customEnvironment.js",
+  setupFilesAfterEnv: ['<rootDir>/../../jest.setup.js'],
```
ThisIsMissEm, line 4: You shouldn't need this again as it's in the root config.

packages/oidc/jest.config.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   ...require("../../jest.config"),
+  testEnvironment: "jsdom",
```
ThisIsMissEm, line 3: Might be worth leaving a comment that this package is only used by browser code so we only test in jsdom

packages/oidc/package.json
```diff
@@ -20,7 +20,7 @@
     "lint:prettier": "prettier \"src/**/*.{ts,tsx,js,jsx,css}\" \"**/*.{md,mdx,yml}\"",
     "format": "prettier --write \"src/**\" \"__tests__/**\"",
     "licenses:check": "license-checker --production --out license.csv --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
-    "test": "jest --coverage --verbose",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --coverage --verbose",
```
ThisIsMissEm, line 23: Be sure to check which node version introduced this

ThisIsMissEm (commented):
Looks good overall, a few minor comments

jest.config.js
NSeydoux: The default is `ts-jest`, which outputs CJS modules, while `ts-jest/presets/default-esm` outputs ESM.

packages/browser/jest.config.js
NSeydoux: `rootDir` doesn't resolve to the same directory whether the `jest` command runs from the root or from a sub-package of the monorepo. Overriding this here allows to run `npm run test` from `packages/browser/`, which is useful when you don't want to run the whole test suite. If this isn't overriden there, `setupFilesAfterEnv` points to `packages/browser/jest.setup.js`.

ThisIsMissEm: it'd be better to set `rootDir` option, I think, and specify that as `rootDir: "../../"` I think?

NSeydoux: Doing so runs all the test suites, and not only the target one, so it doesn't quite solve the issue.

package.json
NSeydoux: Done in 8e4102341a4f85e56e788421c4e7d5491754bb78 (also wow a lot of linting things overdue)

packages/oidc/package.json
NSeydoux: No longer required

packages/browser/jest.config.js
ThisIsMissEm: Ah, alrighty!

jest.setup.ts
```diff
@@ -0,0 +1 @@
+import "@inrupt/jest-jsdom-polyfills";
```
ThisIsMissEm, line 1:
```suggestion
import "@inrupt/jest-jsdom-polyfills";

```

missing trailing new line, and eslint may complain about license header.

jest.config.ts
```diff
@@ -1,10 +1,11 @@
...
+
+export default {
   roots: ["<rootDir>"],
   clearMocks: true,
   testMatch: ["**/(__tests__|src)/**/?(*.)+(spec|test).+(ts|tsx|js)"],
-  transform: {
-    "^.+\\.(ts|tsx)?$": "ts-jest",
-  },
+  preset: "ts-jest/presets/js-with-ts",
+  transformIgnorePatterns: [],
```
ThisIsMissEm, line 8: let's add a comment as to why.

package.json
```diff
@@ -30,30 +30,28 @@
     "lerna": "^6.0.0"
   },
   "devDependencies": {
-    "@inrupt/eslint-config-base": "0.4.0",
-    "@inrupt/eslint-config-lib": "0.4.1",
+    "@inrupt/eslint-config-base": "^1.3.0",
+    "@inrupt/eslint-config-lib": "^1.3.0",
+    "@inrupt/jest-jsdom-polyfills": "^1.3.0",
+    "@peculiar/webcrypto": "^1.4.0",
```
ThisIsMissEm, line 36: This should be done in `@inrupt/jest-jsdom-polyfills`

tests/environment/customEnvironment.ts
```diff
@@ -0,0 +1,25 @@
...
+      this.global.Uint8Array = Uint8Array;
+    }
+
+    // The following can be moved to jest-jsdom-polyfills.
+    // jsdom doesn't implement the Web Crypto API
+    // For some reason,  this.global.crypto = new Crypto() leaves .subtle undefined
+    (this.global.crypto.subtle as any) = (new Crypto()).subtle;
+    this.global.CryptoKey = CryptoKey;
+  }
+};
```
ThisIsMissEm, line 25: Let's fix this in jest-jsdom-polyfills instead of using custom environments.

tsconfig.build.json
```diff
@@ -14,7 +14,9 @@
     "sourceMap": true,
     "lib": ["es2018", "dom"],
     "moduleResolution": "node",
```
ThisIsMissEm, line 16: might be worth mentioning this kinda changes in a way soon-ish: https://www.typescriptlang.org/tsconfig#moduleResolution

ThisIsMissEm (changes requested):
Let's get the changes needed into jest-jsdom-polyfills instead of using a custom environment here.

tests/environment/customEnvironment.ts
NSeydoux: This works in the polyfill (I have tested it locally, I was planning on merging this and revisiting after releasing the updated polyfills), but I couldn't get the `Uint8Array` thing to work in there, so for now I'm afraid the custom environment has to stay. We can revisit why, but I'd like to move on because this has been a time sink :)

package.json
NSeydoux: Waiting on https://github.com/inrupt/javascript-style-configs/pull/62

tests/environment/customEnvironment.ts
NSeydoux: See https://github.com/inrupt/javascript-style-configs/pull/62

jest.config.ts
NSeydoux: 53b454b1bf7188d14e2e793db709c96a74a37cd4

package.json
ThisIsMissEm: Approved

NSeydoux: Now for the release :) https://github.com/inrupt/javascript-style-configs/pull/63

jest.config.ts
```diff
@@ -27,15 +28,48 @@ export default {
...
+
+export default {
+  projects: [{
+    ...baseConfig,
+    displayName: "core",
+    roots: ["<rootDir>/packages/core"],
+  }, {
+    ...baseConfig,
+    displayName: "oidc",
+    roots: ["<rootDir>/packages/oidc"],
```
ThisIsMissEm, line 51: For clarity, would it be worth renaming this folder to `oidc-browser` (not changing the package name, just the folder name)

packages/browser/package.json
```diff
@@ -22,7 +22,7 @@
     "lint:eslint": "eslint --config .eslintrc.js \"src/\"",
     "lint:prettier": "prettier \"src/*.{ts,tsx,js,jsx,css}\" \"**/*.{md,mdx,yml}\"",
     "licenses:check": "license-checker --production --out license.csv --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
-    "test": "OPENSSL_CONF=/dev/null jest --coverage --verbose",
+    "test": "OPENSSL_CONF=/dev/null jest --coverage --verbose --selectProjects browser --config ../../jest.config.ts",
```
ThisIsMissEm, line 25: Can this be done with a setupFiles setting that mutates process.env? (That is, before the test code is loaded)

jest.config.ts
NSeydoux: No strong opinion, but no harm in being more explicit: 1a94ef691cf16b8a33eb8530c406e7e7584979b0

packages/browser/package.json
NSeydoux: Interestingly, it can be done in jest's config file, but not as part of the environment setup for some reason: d18fbcb4cf8662ae90410c16b7674b0931e8dc39.

jest.config.ts
NSeydoux: Actually this breaks sonar integration, which would create a bit of delay to sort out where this config is. I'd like this PR to get merged now, as it has already lingered quite a while, so let's do the renaming later when I have figured out what needs to be adjusted on sonar.

ThisIsMissEm: It should just be updating here: https://github.com/inrupt/solid-client-authn-js/blob/main/.github/workflows/ci.yml#L63

ThisIsMissEm: But I am of course okay doing it separatey

NSeydoux: Ah cool, it was late yesterday when I was looking into it so I didn't want to investigate, but if that's just that's good.

jest.config.ts
```diff
@@ -0,0 +1,78 @@
+import { JestConfigWithTsJest } from "ts-jest";
+
+const baseConfig: JestConfigWithTsJest = {
+  roots: ["<rootDir>"],
+  clearMocks: true,
+  testMatch: ["**/(__tests__|src)/**/?(*.)+(spec|test).+(ts|tsx|js)"],
```
ThisIsMissEm, line 6: We should really standardise around either `.test.ts` or `.spec.ts`, but not both. Maybe we can already change this to: 
```suggestion
  testMatch: ["src/**/(.+).spec.ts"],
```

Elsewhere we use: 
```
testRegex: "/src/.*\\.test\\.ts$",
```

This is definitely a future problem to tackle though. But lets standardise across all our projects.

NSeydoux: This repo should consistently be using `.spec.ts`, and I have no problem changing to `.test.ts`, just in another PR :)

packages/browser/src/login/oidc/incomingRedirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -149,12 +149,16 @@ export class AuthCodeRedirectHandler implements IIncomingRedirectHandler {
...
     }
 
-    const authFetch = await buildAuthenticatedFetch(fetch, tokens.accessToken, {
-      dpopKey: tokens.dpopKey,
-      refreshOptions,
-      eventEmitter,
-      expiresIn: tokens.expiresIn,
-    });
+    const authFetch = await buildAuthenticatedFetch(
+      window.fetch,
```
ThisIsMissEm, line 153: should this be `window.fetch` or should it be `import { fetch } from "cross-fetch";`? (I mean, yeah, arguably we are in the browser, so we have `window.fetch` but I think maybe we want to standardise around the import, especially with our Node.js 18+ support work.

packages/core/package.json
```diff
@@ -30,29 +31,23 @@
     "prepublishOnly": "npm run build",
     "clean": "npm run clean-module",
     "clean-module": "rimraf ./dist",
-    "build": "tsc -p tsconfig.json",
+    "build": "rollup --config rollup.config.mjs",
```
ThisIsMissEm, line 34: ooohh! aaahh! You've done rollup here too! Nice!!

packages/core/package.json
```diff
@@ -30,29 +31,23 @@
...
   },
-  "devDependencies": {
-    "@inrupt/vocab-inrupt-common": "^0.7.2",
-    "@types/n3": "^1.4.4",
-    "n3": "^1.6.4"
-  },
   "dependencies": {
-    "@inrupt/solid-common-vocab": "^1.0.0",
     "@types/lodash.clonedeep": "^4.5.6",
     "@types/uuid": "^8.3.0",
```
ThisIsMissEm, line 47: These need to be moved to `devDependencies`

packages/core/rollup.config.mjs
```diff
@@ -0,0 +1,62 @@
...
+      typescript: require("typescript"),
+      tsconfigOverride: {
+        compilerOptions: {
+          module: "esnext",
+        },
+      },
+    })
+  ],
+  // The following option is useful because symlinks are used in monorepos
+  preserveSymlinks: true,
```
ThisIsMissEm, line 36: Let's move this into a `common` const, as it's used in both exports (index & mocks)

packages/core/src/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -98,6 +98,17 @@ const mockFetch = (response: Response, url: string) => {
...
+  const spyTimeout = jest.spyOn(global, "setTimeout");
+
+  afterEach(() => {
+    // Clear the latest timeout to avoid dangling open handles.
+    // FIXME: Should just use fake timers, but that chokes on recursive calls.
+    const handle = spyTimeout.mock.results[spyTimeout.mock.results.length - 1];
+    if (handle !== undefined) {
+      (handle.value as ReturnType<typeof setTimeout>).unref();
+    }
+  });
```
ThisIsMissEm, line 110: Can this be changed with the usage of mock timers? https://jestjs.io/docs/timer-mocks — err, totally blanked on the comment; Let's open a Jira ticket to fix this.

ThisIsMissEm: I wonder if the issue you faces was due to https://jestjs.io/docs/timer-mocks#run-pending-timers

i.e., `jest.runOnlyPendingTimers();` vs `jest.runAllTimers();`

packages/core/src/authenticatedFetch/fetchFactory.spec.ts
```diff
@@ -314,7 +330,9 @@ describe("buildAuthenticatedFetch", () => {
 
   // jest.useFakeTimers();
   function sleep(ms: number) {
-    return new Promise((resolve) => setTimeout(resolve, ms));
+    return new Promise((resolve) => {
+      setTimeout(resolve, ms);
+    });
```
ThisIsMissEm, line 335: When we use fake timers, we'll need to change these calls to `jest.advanceTimersByTime(ms);`

packages/core/src/login/oidc/__mocks__/IncomingRedirectHandler.ts
```diff
@@ -22,23 +22,17 @@
...
-      sessionId: "global",
-      isLoggedIn: true,
-      webId: "https://pod.com/profile/card#me",
-      fetch: jest.fn(),
-    })
+const handle = jest.fn((_url: string, _emitter: EventEmitter | undefined) =>
+  Promise.resolve({
+    sessionId: "global",
+    isLoggedIn: true,
+    webId: "https://pod.com/profile/card#me",
```
ThisIsMissEm, line 33:
```suggestion
    webId: "https://webid.test/profile/card#me",
```

We shouldn't use real URLs in tests, `pod.com` is potentially real

packages/oidc-browser/package.json
```diff
@@ -20,20 +20,21 @@
...
   "devDependencies": {
-    "cross-fetch": "^3.1.5"
+    "cross-fetch": "^3.1.5",
+    "ts-node": "^10.9.1"
   },
   "dependencies": {
     "@inrupt/oidc-client": "^1.11.6",
     "@inrupt/solid-client-authn-core": "^1.12.2",
     "@types/jest": "^27.0.3",
     "@types/uuid": "^8.3.0",
```
ThisIsMissEm, line 36: `@types` need to be moved to `devDependencies`

packages/oidc-browser/src/refresh/refreshGrant.spec.ts
```diff
@@ -48,6 +48,15 @@ jest.mock("@inrupt/solid-client-authn-core", () => {
...
 
+// jest.unstable_mockModule("@inrupt/solid-client-authn-core", async () => {
+//   const actualCoreModule = await import("@inrupt/solid-client-authn-core");
+//   return {
+//     ...actualCoreModule,
+//     // This works around the network lookup to the JWKS in order to validate the ID token.
+//     getWebidFromTokenPayload: jest.fn<() => Promise<string>>().mockResolvedValue("https://my.webid/")
+//   };
+// });
+
```
ThisIsMissEm, line 59: Can this be deleted entirely?

tests/environment/customEnvironment.ts
ThisIsMissEm: I think this should've been released now, so this shouldn't be necessary anymore.

ThisIsMissEm (changes requested):
Looks good! Just a few more things left to resolve — Feel free to comment on them if you split them into separate work tickets.

packages/browser/src/Session.ts
```diff
@@ -159,7 +159,7 @@ export class Session extends EventEmitter {
    */
   constructor(
     sessionOptions: Partial<ISessionOptions> = {},
-    sessionId?: string
+    sessionId: string | undefined = undefined
```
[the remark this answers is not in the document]
NSeydoux: Deprecating the session id being nullable on construction? I don't think so.

packages/browser/src/login/oidc/incomingRedirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -143,37 +146,22 @@ jest.mock("@inrupt/oidc-client-ext");
...
-  );
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  mockedOidcClient.createDpopHeader = (jest.fn() as any).mockResolvedValue(
-    "someToken"
+function mockOidcClient(): typeof OidcClientExt {
+  const mockedOidcClient = jest.requireMock<typeof OidcClientExt>(
+    "@inrupt/oidc-client-ext"
   );
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  mockedOidcClient.decodeJwt = (jest.fn() as any).mockResolvedValue(jwt);
```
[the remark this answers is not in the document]
NSeydoux: Yeah I'm not sure how that got there in the first place ^

packages/core/package.json
NSeydoux: It's just for the core though, I think the browser relies on WebPack a bit more so that will be more of a pickle. Here is allows to output ESM as well as CJS (which I don't think was actually needed in retrospective ^^)

packages/core/src/authenticatedFetch/fetchFactory.spec.ts
NSeydoux: Unfortunately no it didn't appear to be the issue. Basically it looks like Jest handles the timers being set by `setTimeout`, but not the ones set in the callback passed to `setTimeout`, resulting in the second round of timers not to be mocked.

jest.config.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+  },
+  collectCoverageFrom: [
+    "<rootDir>/src/**/*.ts",
+  ],
+  coveragePathIgnorePatterns: [
+    "node_modules/",
+    "dist/",
+    ".*.spec.ts"
+  ],
```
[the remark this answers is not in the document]
NSeydoux: `dist` appears to be required there, but not node_modules.

packages/browser/src/login/oidc/incomingRedirectHandler/AuthCodeRedirectHandler.ts
NSeydoux: Fixed in 535276759bd27391d0c211ccca72d2531de128f8, but there's a lot of that remaining :)

packages/core/src/login/oidc/__mocks__/IncomingRedirectHandler.ts
NSeydoux: I'll add a tech debt ticket to do this all at once, it applies across all codebase

packages/browser/src/Session.ts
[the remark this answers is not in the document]
ThisIsMissEm: I thought it was moved into sessionOptions

PR #2445, opened 2022-11-16
Use shared env setup in e2e tests

NSeydoux:


jest.e2e.config.ts
```diff
@@ -1,11 +1,14 @@
...
   testTimeout: 30000,
-  testMatch: ["**/e2e/?(*.)+(spec|test).+(ts|tsx|js)"],
+  testMatch: ["<rootDir>/e2e/node/*.spec.ts"],
   transform: {
     "^.+\\.(ts|tsx)?$": "ts-jest",
   },
   testEnvironment: "node",
-};
+  setupFiles: ["<rootDir>/jest.e2e.setup.ts"]
+} as Config;
```
ThisIsMissEm, line 14: Can this be merged with the main jest config?

e2e/browser/e2e.playwright.ts
```diff
@@ -25,7 +25,9 @@ import { LoginFlow } from "./pages/LoginFlow";
 // TODO: Redirected resource tests? I'm not sure what those actually show
 
 test.describe.parallel("Not Logged In", () => {
```
ThisIsMissEm, line 27: I think I've seen tests fail due to hitting the same pod at the same time, maybe it's worth removing the .parallel?

ThisIsMissEm (commented):
I've done a first pass review, look good, not too keen on leaving a disabled test case though.

e2e/browser/e2e.playwright.ts
```diff
@@ -25,7 +25,9 @@ import { LoginFlow } from "./pages/LoginFlow";
...
-  test("Public resource in my Pod", async ({ testContainer, app }) => {
+  // Skipping this for now, as it is currently failing. Will investigate separately.
```
chelseapinka, line 28: why are we using parallel here?

NSeydoux: I'm not sure, so I'll disable this for now.

e2e/browser/e2e.playwright.ts
NSeydoux: Probably so, I'm not sure why it got here in the first place.

NSeydoux:
Note that the added skipped test should si behind a feature flag, as not allowing unauthenticated fetches to public resources is a feature of ESS by design, the logic being that "anyone can access the resource" doesn't prevent from having audit capability to know who access it effectively for instance.

chelseapinka (approved):

PR #2472, opened 2022-11-30
chore: minor cleanups

ThisIsMissEm:
This PR cleans up a few things that we'd previously changed, but missed parts of:
- removes the custom license header (inherits from typescript-sdk-tools
- removes outdated overrides
- updates lockfiles

PR #2473, opened 2022-11-30
Chore: update demo client app

ThisIsMissEm:
Given #2464, we can refactor the demo client app to use more of the standard approach, as well as resolve some outdated dependency issues. This builds on #2472


packages/browser/examples/demoClientApp/src/DemoClientApp.js
```diff
@@ -132,6 +66,7 @@ class DemoClientApp extends Component {
 
     this.state = {
       status: "loading",
+      // FIXME: This is currently wrong on session restore, as we don't expose the Issuer from ISessionInfo
```
ThisIsMissEm, line 69: @NSeydoux this is where I hit into the issue with not having the issuer exposed.

packages/browser/tsconfig.eslint.json
```diff
@@ -2,5 +2,5 @@
   "extends": "./tsconfig.json",
 
   "include": ["src/**/*"],
-  "exclude": []
+  "exclude": ["examples/**/*"]
```
ThisIsMissEm, line 5: The examples are currently in JavaScript, not typescript, so should be excluded.

NSeydoux (approved):

PR #2484, opened 2022-12-08
Refactor ESLint to run from monorepo root

NSeydoux:
- Centralize eslint config: Run all linting commands from the root of the repository, and no longer rely on lerna to run linting commands at a package granularity.
- Delete now unused eslint parser configs
- Run linting on previously unlinted files

For some reason this ended up linting all the files to update the license header.


chelseapinka (approved):
LGTM except for the failing tests but it seems like some are rerunning successfully?

package.json
```diff
@@ -12,8 +12,29 @@
...
+    "lint:eslint:node": "eslint --config packages/node/.eslintrc.js \"packages/node/src/\"",
+    "lint:prettier:node": "prettier \"packages/node/**/*.{ts,tsx,js,jsx,css}\" \"packages/node/**/*.{md,mdx,yml}\"",
+    "lint:fix:core": "npm run lint:eslint:core -- --fix && npm run lint:prettier:core -- --write",
+    "lint:check:core": "npm run lint:eslint:core && npm run lint:prettier:core -- --check",
+    "lint:eslint:core": "eslint --config packages/core/.eslintrc.js \"packages/core/src/\"",
+    "lint:prettier:core": "prettier \"packages/core/**/*.{ts,tsx,js,jsx,css}\" \"packages/core/**/*.{md,mdx,yml}\"",
+    "lint:fix:oidc-browser": "npm run lint:eslint:oidc-browser -- --fix && npm run lint:prettier:oidc-browser -- --write",
+    "lint:check:oidc-browser": "npm run lint:eslint:oidc-browser && npm run lint:prettier:oidc-browser -- --check",
+    "lint:eslint:oidc-browser": "eslint --config packages/oidc-browser/.eslintrc.js \"packages/oidc-browser/src/\"",
+    "lint:prettier:oidc-browser": "prettier \"packages/oidc-browser/**/*.{ts,tsx,js,jsx,css}\" \"packages/oidc-browser/**/*.{md,mdx,yml}\"",
```
ThisIsMissEm, line 37: This is quite a configuration smell, we should probably look at migrating this repository to `nx` and `npm workspaces` instead, which should help here, especially with caching and executing with dependencies.

ThisIsMissEm (commented):
Let's review the custom eslint configuration rules, they may not be applicable anymore. Also, we've introduced a code smell with those `npm scripts` that have been added, we should find a better way that's more maintainable. Ideally, just using a root and having the overrides for each child package, if necessary, in the root, would be better.

package.json
NSeydoux: Yup it looks pretty bad, but I'm not sure to what extent this is worse than what we had before: it's the same amount of config, just all in the same place rather than scattered across the monorepo packages.

ThisIsMissEm: hmm... I feel like we should just be able to run from the root, ignoring lerna & workspaces, and it should just "work"

NSeydoux:
https://github.com/inrupt/solid-client-authn-js/pull/2526 follows up on the comments on this PR

PR #2492, opened 2022-12-09
Replaces events with tiny-emitter

NSeydoux:
`events` implements a bigger set of features than we need, and `tiny-emitter` is sufficient for our purpose. It is more lightweight, and has less bug surface because it has less code.


ThisIsMissEm (approved):
Code wise this LGTM, but let's make sure we ship this as an alpha release first, such that we can test it out in various applications as-compiled & packaged.

PR #2587, opened 2023-01-13
Reduce dependabot updates frequency

NSeydoux:
Updating dev dependencies doesn't need to happen as often as prod dependencies, and maybe that would help reducing the amount of in-flight PRs.


NSeydoux:
Closing this, as the proposed change isn't compatible with the dependabot config expectations (the <ecosystem, branch, directory> triplet must be unique)

PR #2588, opened 2023-01-13
Skip jobs at a matrix level for dependabot

NSeydoux:
Skipping jobs at the top-level prevents the e2e tests being gating in the non-dependabot PRs, while skipping them at a lower level in the matrix allow for them to be consistently present in all PRs. They can then be marked as gating.


ThisIsMissEm (approved):
This does feel weird, but if it's what's needed, then fine. Make sure you also do the e2e-browser tests.

PR #2601, opened 2023-01-17
Accelerate dependabot CI by skipping more steps

NSeydoux:
The end-to-end tests are eventually skipped in dependabot in any case, this just skips more steps, targeting the more time-consuming ones. This directly builds on b1aab125d732f530649a5b3b39d2f04c9a416c47.

### DIFF
--- a/.github/workflows/e2e-browser.yml
+++ b/.github/workflows/e2e-browser.yml
@@ -29,12 +29,16 @@ jobs:
           node-version: 16.x
           cache: npm
       - run: npm ci
+        if: github.actor != 'dependabot[bot]'
       - run: npm run bootstrap -- --ci
       - run: npm run build
+        if: github.actor != 'dependabot[bot]'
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-browser.yml
```diff
@@ -29,12 +29,16 @@ jobs:
           node-version: 16.x
           cache: npm
       - run: npm ci
+        if: github.actor != 'dependabot[bot]'
       - run: npm run bootstrap -- --ci
```
ThisIsMissEm, line 33: do we actually need this still in CI?

ThisIsMissEm (approved):

.github/workflows/e2e-browser.yml
NSeydoux: I don't think so, since this is in the `package.json`: "postinstall": "lerna bootstrap --ci && npm run build". I'll change this in a separate PR.

PR #2602, opened 2023-01-17
Update test app dependencies

NSeydoux:
This does the major update from Next 12 to 13 in the test browser app, alongside other updates.


ThisIsMissEm:
This may be risky, as next.js 13 messes with `fetch`: https://nextjs.org/blog/next-13#data-fetching

NSeydoux:
I manually ran the app without issues, and all the end-to-end tests are running, so I don't think it impacts us.

ThisIsMissEm (approved):

PR #2603, opened 2023-01-17
Remove redundant CI build steps

NSeydoux:
`npm ci` has a postinstall script running both lerna bootstrap and npm build, so having explicit step to do so is redundant.


ThisIsMissEm:
I'd actually like to remove the postinstall, tbh.

jeswr (approved):
LGTM

jeswr:
@NSeydoux Not sure why this wasn't merged when first opened - we can address post-install script in a follow up

PR #2643, opened 2023-02-13
Prevent redirect URL from being normalized

NSeydoux:
The redirect URL provided to the `login` function was being normalized, which
could result in misalignments with the redirect URLs declared in a Client Identifier
document, since the latter wasn't being normalized. The normalization step has
now been removed, and the redirect provided by the user is sent to the OpenID
Provider unchanged.


packages/browser/examples/single/script/index.html
```diff
@@ -4,8 +4,8 @@
     <h1>solid-client-authn Script Tag Tester</h1>
     <p>
       Logging in to
-      <a id="oidcIssuer" href="https://broker.pod.inrupt.com/"
-        >https://broker.pod.inrupt.com/</a
+      <a id="oidcIssuer" href="https://login.inrupt.com/"
```
solid-akb, line 7: 😅 great catch

e2e/browser/test-app/components/appContainer/index.tsx
```diff
@@ -28,7 +28,7 @@ import {
 } from "@inrupt/solid-client-authn-browser";
 import AuthenticatedFetch from "../authenticatedFetch";
 
-const REDIRECT_URL = window.location.href;
+const REDIRECT_URL = "http://localhost:3001";
```
solid-akb, line 31: I know this is a conversation we've been having, but I still believe that using `window.location.href` is a better pattern than hardcoding it. 

Two thoughts that have come from building apps and using both approaches:
- reusability: If the redirect url is being set dynamically, you can easily reuse components without having to hardcode the redirect URL. This is especially helpful in the case of routing and deeplinking where you can define which routes should be allowed to redirect in your client identifier document but you don't need to explicitly call it out in the code. 
- separating deployment details from code: when we hardcode the redirect url in the code, we have to know where our application is being hosted. This to me should be as separate from the code itself and should be handled in our CI/CD process or package file or env file and not in the source code itself.

We may run into issues if a `window.location.href` in our application is not in our client identifier document and the redirect will fail, but I think instead of calling this an issue or a failure point, what is more beneficial is to understand what happened and either add a route to the client identifier document or update how the application is deployed. 


*Note this is not a blocking comment as I don't want to block this PR before we wrap up the conversation*

VirginiaBalseiro (approved):

PR #2644, opened 2023-02-13
Use standard URL constructor for path appending

NSeydoux:
Remove a custom implementation of a function appending a path to a base URL, and replace it with the standard URL constructor, which implements the same logic.

Note that this doesn't include a CHANGELOG update, because it is a purely internal change that just removes some tech debt, and as such it is transparent to users.


solid-akb (approved):
🙌 love this PR. Makes the code easier to read. Thanks for taking the initiative here!

PR #2649, opened 2023-02-13
Fix GHSA-rc47-6667-2j5j

NSeydoux:


solid-akb (approved):
Small typo in changelog but otherwise approve!

PR #2658, opened 2023-02-15
Fix/token expiration

VirginiaBalseiro:
This PR is to ensure that the token expiration is passed on to `buildAuthenticatedFetch`, so that the proper expiration time is set and the logout signal is sent at the correct time.  


NSeydoux (changes requested):

NSeydoux (approved):

NSeydoux (changes requested):
A changelog update is necessary

NSeydoux (approved):

PR #2711, opened 2023-03-07
Append .well-known component to full issuer URL

NSeydoux:
This reverts the change in https://github.com/inrupt/solid-client-authn-js/commit/baac030d33163ba08dadebabdaf676450be7fa88. According to [the specification](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig), /.well-known/openid-configuration should be appended at the end of the full issuer URL, and the current code removed the last path component if it didn't have a trailing slash.


pmcb55 (approved):
LGTM

PR #2739, opened 2023-03-13
Add `EventEmitter` attribute to Session class

NSeydoux:
Instead of having the `Session` be an `EventEmitter`, this adds an `events` attribute instantiating an extension of `EventEmitter`. The new emitter class itself doesn't add anything to the base `EventEmitter`, except for type hints about the callbacks for each supported event.

In order to prevent a breaking change, the Session class should keep extending the `EventEmitter` API for the time being. To keep things simple, `session.events` will proxy to session, so that even during the transition there is a single `EventEmitter` instance. When we decide to make the breaking change, we can just remove the proxying, and have `session.events` be the `EventEmitter` instance instead of the session.

# Checklist

### DIFF
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -32,7 +32,7 @@ import { Session } from "./Session";
 import { LocalStorageMock } from "./storage/__mocks__/LocalStorage";
 import { mockSessionInfoManager } from "./sessionInfo/__mocks__/SessionInfoManager";
 import { KEY_CURRENT_SESSION, KEY_CURRENT_URL } from "./constant";
-import type ClientAuthentication from "./ClientAuthentication";
+import ClientAuthentication from "./ClientAuthentication";
 
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 
@@ -912,14 +912,6 @@ describe("Session", () => {
       const spiedOn = jest.spyOn(mySession, "on");
       mySession.events.on("login", jest.fn());
       expect(spiedOn).toHaveBeenCalled();
-    });
-
-    it("throws on calls from events which aren't part of the EventEmitter interface", () => {
-      const mySession = new Session();
-      // @ts-expect-error onLogin is a function on Session, and not SessionEventEmitter
-      expect(() => mySession.events.onLogin(jest.fn())).toThrow(
-        "[onLogin] is not supported"
-      );
     });
   });
 });
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -903,4 +903,23 @@ describe("Session", () => {
       expect(myCallback).toHaveBeenCalled();
     });
   });
+
+  describe("proxies events to the session", () => {
+    // This describe block is only required as long as Session extends the EventEmitter
+    // class.
+    it("proxies the EventEmitter calls from events to the session object", () => {
+      const mySession = new Session();
+      const spiedOn = jest.spyOn(mySession, "on");
+      mySession.events.on("login", jest.fn());
+      expect(spiedOn).toHaveBeenCalled();
+    });
+
+    it("throws on calls from events which aren't part of the EventEmitter interface", () => {
+      const mySession = new Session();
+      // @ts-expect-error onLogin is a function on Session, and not SessionEventEmitter
+      expect(() => mySession.events.onLogin(jest.fn())).toThrow(
+        "[onLogin] is not supported"
+      );
+    });
+  });
 });
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -1022,5 +1022,13 @@ describe("Session", () => {
       mySession.events.on("login", jest.fn());
       expect(spiedOn).toHaveBeenCalled();
     });
+
+    it("throws on calls from events which aren't part of the EventEmitter interface", () => {
+      const mySession = new Session();
+      // @ts-expect-error onLogin is a function on Session, and not SessionEventEmitter
+      expect(() => mySession.events.onLogin(jest.fn())).toThrow(
+        "[onLogin] is not supported"
+      );
+    });
   });
 });
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -1011,18 +1011,6 @@ describe("Session", () => {
         expect.assertions(1);
       });
     });
-
-    describe("sessionExpired", () => {
-      it("calls the provided callback when receiving the appropriate event", async () => {
-        const myCallback = jest.fn();
-        const mySession = new Session({
-          clientAuthentication: mockClientAuthentication(),
-        });
-        mySession.events.on(EVENTS.SESSION_EXPIRED, myCallback);
-        mySession.events.emit(EVENTS.SESSION_EXPIRED);
-        expect(myCallback).toHaveBeenCalled();
-      });
-    });
   });
 
   describe("proxies events to the session", () => {
--- a/packages/browser/src/Session.spec.ts
+++ b/packages/browser/src/Session.spec.ts
@@ -408,19 +408,6 @@ describe("Session", () => {
     });
   });
 
-  describe("onError", () => {
-    it("calls the registered callback on error", async () => {
-      const onErrorCallback = jest.fn();
-      const mySession = new Session();
-      mySession.onError(onErrorCallback);
-      mySession.emit(EVENTS.ERROR, "error", "error_description");
-      expect(onErrorCallback).toHaveBeenCalledWith(
-        "error",
-        "error_description"
-      );
-    });
-  });
-
   describe("silent authentication", () => {
     it("does nothing if no previous session is available", async () => {
       mockLocalStorage({});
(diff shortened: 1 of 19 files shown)


packages/browser/src/Session.spec.ts
```diff
@@ -902,4 +904,134 @@ describe("Session", () => {
...
+    });
+
+    describe("sessionExpired", () => {
+      it("calls the provided callback when receiving the appropriate event", async () => {
+        const myCallback = jest.fn();
+        const mySession = new Session({
+          clientAuthentication: mockClientAuthentication(),
+        });
+        mySession.events.on(EVENTS.SESSION_EXPIRED, myCallback);
+        mySession.events.emit(EVENTS.SESSION_EXPIRED);
```
NSeydoux, line 1021: This is a very shallow test (to say the least), which doesn't check that the event is sent at the appropriate moment. I'm questioning whether it is useful at all.

NSeydoux: I removed this and replaced it with an actual end-to-end test, because this unit test wasn't providing value.

packages/browser/src/Session.ts
```diff
@@ -132,16 +133,43 @@ function isLoggedIn(
...
+   * @hidden
+   */
+  private handler = {
+    // This proxy is only a temporary measure until Session no longer extends
+    // SessionEventEmitter, and the proxying is no longer necessary.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    get(target: any, prop: any, receiver: any) {
+      return Reflect.get(target, prop, receiver);
+    },
+  };
```
jeswr, line 171: Since this will proxy every property/method of the `Session` it might be worth throwing an error here if consumers try and access properties or methods that are not available in the `EventEmitter` API.

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,256 @@
...
+  ): this {
+    return super.on(eventName, listener);
+  }
+
+  /**
+   * Register a listener called on the next successful login with the logged in WebID.
+   * @param eventName The login event name.
+   * @param listener The callback called on the next successful login.
+   */
+  once(eventName: typeof EVENTS.LOGIN, listener: (webid: string) => void): this;
```
jeswr, line 113: This differs from the type signature of the `on` method for the login event which is just `() => void`.

To avoid these kinds of errors can I suggest defining the params as a type; e.g. `type LOGIN_ARGS = [typeof EVENTS.LOGIN, listener: (webid: string) => void)]` so each method could share the type and be consistent.

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,256 @@
...
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+import { EventEmitter } from "events";
+import { EVENTS } from "./constant";
+
+export interface ISessionEventEmitter {
+  events: SessionEventEmitter;
+}
+
+export class SessionEventEmitter extends EventEmitter {
```
jeswr, line 28: I'm not convinced this needs to be a class - shouldn't this just be an interface and in the long term the events property is created as `public events: SessionEventEmitter = new EventEmitter()`. Functionally the reflection of properties doesn't seem to do anything.

packages/core/src/constant.ts
```diff
@@ -41,7 +41,7 @@ export const EVENTS = {
   SESSION_EXTENDED: "sessionExtended",
   SESSION_RESTORED: "sessionRestore",
   TIMEOUT_SET: "timeoutSet",
-};
+} as const;
```
jeswr, line 44: Though unlikely, this may be a breaking change for consumers if they have any types expecting this to be a `Record<string, string>`.

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * Register a listener called when a timeout is set for a session event with
+   * the timeout handle.
+   * @param eventName The timeout set event name.
+   * @param listener The callback called when setting a timeout.
+   */
+  on(
+    eventName: typeof EVENTS.TIMEOUT_SET,
+    listener: (timeoutHandle: number) => void
+  ): this;
```
jeswr, line 91: Could the `timeoutHandle` name be more desriptive here? Perhaps `millisecondsToTimeout`?

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * Unegister a listener called on session restoration after a silent login.
+   * @param eventName The session restoration event name.
+   * @param listener The callback to unregister.
+   */
+  off(
+    eventName: typeof EVENTS.SESSION_RESTORED,
+    listener: (currentUrl: string) => unknown
+  ): this;
+  /**
```
jeswr, line 210: Most of the `off` handlers appear to have an incorrect description for the `listener`

packages/node/src/Session.ts
```diff
@@ -85,12 +86,39 @@ export const defaultStorage = new InMemoryStorage();
...
+   * When Session no longer implements SessionEventEmitter, this can be removed.
+   * @hidden
+   */
+  private handler = {
+    // This proxy is only a temporary measure until Session no longer extends
+    // SessionEventEmitter, and the proxying is no longer necessary.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    get(target: any, prop: any, receiver: any) {
+      return Reflect.get(target, prop, receiver);
+    },
```
jeswr, line 119: My comments about erroring on non `EventEmitter` properties also applies to the node API.

jeswr (changes requested):

jeswr:
I haven't thoroughly reviewed the tests yet - will do so once the above comments are addressed.

packages/browser/src/Session.ts
NSeydoux: I agree, and that was actually part of the initial commit in 5d6228c02a04d262bb2eb1d8b0611f46669d7383. However, I reverted this in 91c49d4388d7ac8734be4942c3b1cf7622445fd7, because these checks are necessary when applied on the public API, but they also apply to the internal calls, which means the whole prototype chain should be present, as internal calls go beyond the `EventEmitter`. However, maybe reversing the logic, and throwing if a called method is in the `Session` API and not in the `EventEmitter` one (rather than just the latter condition) would work.

packages/core/src/SessionEventEmitter.ts
NSeydoux: Do you mean in the type description or the API docs? Based on https://nodejs.org/api/events.html#emitterremovelistenereventname-listener the API docs sounds right, although I'm open to any change of phrasing. I'll double-check the type signatures.

packages/core/src/SessionEventEmitter.ts
NSeydoux: I'll change this to `timeoutId`, as per https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value

packages/core/src/SessionEventEmitter.ts
NSeydoux: That's an excellent point, I didn't think about verifying whether there was an `EventEmitter` interface along the existing class, and there is. Fixed with e5bed656.

packages/browser/src/Session.ts
NSeydoux: I think da7b5dfc fixes the issue: any call to `session.events` that is part of `Session` but not of `EventEmitter` will throw, which prevents using the proxied `session.events` as the main `session`.

packages/core/src/constant.ts
NSeydoux: That's correct, and the same applies to users using a session instance as an arbitrary event emitter: the type overloads we are specifying don't allow for non-session specific events. Since none of this is public API (core being an internal module to the public the browser and node packages), I think it is unlikely enough that downstream users are relying on this to make it okay for this to change. In particular, not making this a `const` prevents from getting useful type hints in the `SessionEventEmitter` interface, as `typeof EVENTS.LOGIN` would be `string` rather than `"login"`.

packages/core/src/SessionEventEmitter.ts
NSeydoux: Fixed with b054d7e5

packages/node/src/Session.ts
```diff
@@ -85,12 +87,45 @@ export const defaultStorage = new InMemoryStorage();
...
+   */
+  private handler = {
+    // This proxy is only a temporary measure until Session no longer extends
+    // SessionEventEmitter, and the proxying is no longer necessary.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    get(target: any, prop: any, receiver: any) {
+      // Reject any calls to the proxy that isn't specific to the EventEmitter API
+      if (
+        !Object.getOwnPropertyNames(EventEmitter).includes(prop) &&
+        Object.getOwnPropertyNames(Session.prototype).includes(prop)
```
jeswr, line 119: Why is the second condition here - don't we just want to error on everything not included in the `EventEmitter` API?

packages/browser/src/Session.ts
```diff
@@ -132,16 +134,49 @@ function isLoggedIn(
...
+   */
+  private handler = {
+    // This proxy is only a temporary measure until Session no longer extends
+    // SessionEventEmitter, and the proxying is no longer necessary.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    get(target: any, prop: any, receiver: any) {
+      // Reject any calls to the proxy that isn't specific to the EventEmitter API
+      if (
+        !Object.getOwnPropertyNames(EventEmitter).includes(prop) &&
+        Object.getOwnPropertyNames(Session.prototype).includes(prop)
```
jeswr, line 170: Why is the second condition here - don't we just want to error on everything not included in the `EventEmitter` API?

packages/browser/src/Session.ts
```diff
@@ -132,16 +134,49 @@ function isLoggedIn(
...
+      ) {
+        throw new Error(
+          `events only implements SessionEventEmitter, [${prop}] is not supported`
+        );
+      }
+      return Reflect.get(target, prop, receiver);
+    },
+  };
```
jeswr, line 178: This handler can just be an object exported from the authn-core packages and re-used by the browser + node packages.

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,309 @@
...
+  /* eslint-disable lines-between-class-members */
+  /**
+   * Register a listener called on successful login.
+   * @param eventName The login event name
+   * @param listener The callback called on successful login
+   */
+  on(
+    eventName: LOGIN_ARGS["eventName"],
+    listener: LOGIN_ARGS["listener"]
+  ): this;
```
jeswr, line 72: It might be worthwhile overloading the `emit` event as well to do one of two things:
 - Disallow it's use entirely
 - Ensure that when used with events used by on/once/off that the arguments match the types of the arguments for the listener function.

jeswr: It might also be worth overloading the `addListener` and `removeListener` parameters which correspond to the `on`, `off` events for developers that use the old API *or* marking `addListener` and `removeListener` as deprecated methods.

packages/browser/src/defaultSession.ts
```diff
@@ -119,8 +123,18 @@ export const onLogout: Session["onLogout"] = (...args) => {
...
 };
+
+/**
+ * {@link SessionEventEmitter} instance to subscribe to events by the default session.
+ *
+ * @since Unreleased
+ */
+export const events = (): Session["events"] => {
+  return getDefaultSession().events;
+};
```
jeswr, line 140: What is the rationale for making this a function of its own rather than having consumers just use the `.events` property of the default session?

packages/core/src/SessionEventEmitter.ts
```diff
@@ -0,0 +1,309 @@
...
+  eventName: typeof EVENTS.ERROR;
+  listener: (error: string | null, errorDescription?: string | null) => unknown;
+};
+type SESSION_EXTENDED_ARGS = {
+  eventName: typeof EVENTS.SESSION_EXTENDED;
+  listener: (expiresIn: number) => void;
+};
+type TIMEOUT_SET_ARGS = {
+  eventName: typeof EVENTS.TIMEOUT_SET;
+  listener: (timeoutHandle: number) => void;
```
jeswr, line 49: This variable name is still unclear - perhaps `millisecondsToTimeout`?

jeswr (changes requested):

packages/browser/src/Session.ts
NSeydoux: I don't think that would be possible: that's the initial approach I took, but calls to the proxy include some internals higher in the prototype chain we don't want to block.

packages/browser/src/Session.ts
NSeydoux: Good idea, done in 891c1b83

packages/core/src/SessionEventEmitter.ts
NSeydoux: Renamed in https://github.com/inrupt/solid-client-authn-js/pull/2739/commits/73e85304f3301c9e15665371b0ff14cad809e123

packages/core/src/SessionEventEmitter.ts
NSeydoux: Should be all fixed in 60e368ecfe51ba971e8fe72f7c0fee84f6e28aac

packages/browser/src/Session.ts
```diff
@@ -132,16 +135,37 @@ function isLoggedIn(
...
+   * the Session.
+   *
+   * When Session no longer implements SessionEventEmitter, this can be removed.
+   * @hidden
+   */
+  private handler = buildProxyHandler(
+    EventEmitter,
+    Session.prototype,
+    "events only implements SessionEventEmitter"
+  );
```
jeswr, line 167: I still wouldn't provide this handler as a property of this class as all, instead it can be created as a constant in the constructor.

jeswr (approved):
LGTM - one minor comment

PR #2742, opened 2023-03-14
Set session expiration date when using Client Credentials

NSeydoux:
The Node session didn't have the expiration time set appropriately on login. This fixes the issue and adds tests to prevent regression. Browser tests are also added, even though there wasn't any bugs there.

### DIFF
--- a/packages/node/src/login/oidc/refresh/__mocks__/TokenRefresher.ts
+++ b/packages/node/src/login/oidc/refresh/__mocks__/TokenRefresher.ts
@@ -50,6 +50,7 @@ export const mockDefaultTokenSet = (): TokenEndpointResponse => {
   return {
     accessToken: "some refreshed access token",
     idToken: JSON.stringify(mockIdTokenPayload()),
+    expiresAt: Date.now() + 3600,
   };
 };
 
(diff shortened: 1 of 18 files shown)


packages/node/src/login/oidc/refresh/__mocks__/TokenRefresher.ts
```diff
@@ -50,6 +50,7 @@ export const mockDefaultTokenSet = (): TokenEndpointResponse => {
   return {
     accessToken: "some refreshed access token",
     idToken: JSON.stringify(mockIdTokenPayload()),
+    expiresAt: Date.now() + 3600,
```
chelseapinka, line 53: so 3600 is just the "refresh_extension_time" in miliseconds?

chelseapinka (approved):
looks good :+1:

packages/node/src/login/oidc/refresh/__mocks__/TokenRefresher.ts
NSeydoux: Yup, that's just the token lifespan. On top of head, that's the value used by ESS (which isn't particularly important here ^^)

Issue #2738, opened 2023-03-13
mismatching redirect uri in solid-client-authn-browser@1.13.2


bourgeoa:
### Impacted package

Which packages do you think might be impacted by the bug ?

- [ x] solid-client-authn-browser@1.13.2

### Bug description

When logged in and reloading a page on NSS or CSS with mashlib, the browser URL displays : 
```
https://solidcommunity.net/authorize?
client_id=8f79c1de8d8dc9efb19bb1cf0e646254
&redirect_uri=https%3A%2F%2Fbourgeoa.solidcommunity.net%2Fpublic%2F%3Fstate%3D07d7855be91b426789df22bce49e66de
&response_type=code&scope=openid%20offline_access%20webid
&state=07d7855be91b426789df22bce49e66de
&code_challenge=JMn9nqpXHaSd8JY6VA1ioPJPTVc4N7SGXhzS3kLGseM
&code_challenge_method=S256
&prompt=none
&response_mode=query
```

and returns the following error : 
```
{"error":"invalid_request","error_description":"Mismatching redirect uri"}
```

### To Reproduce

1. login a NSS `https://solidcommunity.net` or CSS with mashlib pod `https://solidcommunity.net:3000/`
2. reload the page

### Expected result

A return to the initial URL. In the above example : `https://bourgeoa.solidcommunity.net/public/`

### Environment
```
# npm ls @inrupt/solid-client-authn-browser
solid-server@5.7.6 /mnt/d/github/source/workspaces/node-solid-server
└─┬ mashlib@1.8.7
  └─┬ solid-panes@3.5.31
    └─┬ activitystreams-pane@0.6.10
      └─┬ solid-logic@3.0.1
        └── @inrupt/solid-client-authn-browser@1.13.2

```

bourgeoa:
issue # https://github.com/SolidOS/solidos/issues/200

NSeydoux:
Hello @bourgeoa, and sorry for the delayed response! Thanks for reporting this. This issue is currently being worked on, the next release should patch it.

NSeydoux:
v1.13.4 was just released, it should fix this issue, and it also fixes the end-to-end test that should have caught it in the first place :)

bourgeoa:
Thanks for the quick reaction. I shall try it tomorrow.

I hope to have resolved the other issue : the removal of the redirectUrl normalization.
[...]

PR #2749, opened 2023-03-15
Fix silent refresh redirect URL normalization

NSeydoux:
This PR fixes bug #2738.

The redirect URL is stored on redirect to be reused in case of user refresh. However, it must be cleared from all OpenID params, and must not be altered compared to the originally registered redirect URL sent as part of the client's metadata. In particular, the client domain may not include a trailing slash, and this should not be normalized.

This regression went unnoticed because of another issue with the browser-based app we use for end-to-end testing. This is now fixed too.

- ~[ ] I've added a unit test to test for potential regressions of this bug~. N/A: There was a test already, but the app had a dependency issue


packages/oidc-browser/src/cleanup/cleanup.ts
```diff
@@ -34,7 +34,19 @@ export function removeOidcQueryParam(redirectUrl: string): string {
   // As per https://tools.ietf.org/html/rfc6749#section-3.1.2, the redirect URL
   // must not include a hash fragment.
   cleanedUrl.hash = "";
-  return cleanedUrl.toString();
+  // Do not normalize the trailing slash, and respect the original redirect URL.
```
chelseapinka, line 37: Thank you for theses helpful code comments

chelseapinka (approved):

PR #2875, opened 2023-05-08
Pin urllib3 to v1.x in API docs

NSeydoux:
The requests v2.30.0 release introduced support for urllib3 v2, which appears to be breaking for sphinx. This pins the version of urllib3 to a working version until a patch in sphinx. More context in https://github.com/psf/requests/releases/tag/v2.30.0 and urllib3/urllib3#2168.


NSeydoux:
:facepalm: this is from a fork, let me rebase on the main repo

PR #2986, opened 2023-06-19
chore: add next as a dev dependency

jeswr:
Resolves the following error when trying to run the e2e browser tests locally

```
[WebServer] sh: 1: next: not found
#0 0.894 Error: Process from config.webServer was not able to start. Exit code: 127
```


NSeydoux (changes requested):
`next` is a dependency of the app being tested, not of the test runner. It is already a dependency of the browser e2e app, so it should not be required as a dependency of the main app, otherwise the tests in CI would not be running in the first place.

NSeydoux:
Can you confirm that tests can run locally without this change, and if so close this PR?

chelseapinka:
@NSeydoux @jeswr can confirm, tests run and pass locally without `next` devDep, closing PR.

PR #3003, opened 2023-06-22
chore: upgrade to ts tools v2

jeswr:
This PR does 3 things:
 - Upgrades to v2 of our internal tools.
 - Supercedes #2987
 - Disables sonarcloud until it is configured correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
           path: coverage/
 
   sonar-scan:
+    if: false
     needs: [test]
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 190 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           path: coverage/
 
   sonar-scan:
+    if: false
```
chelseapinka, line 35: what was the configuration problem with sonarcloud?

chelseapinka (approved):

.github/workflows/ci.yml
jeswr: It's not working with the monrepo setup because it is not being given the coverage correctly. I thought it was doing nothing at present; but at a second look it does still seem to be doing static analysis so we should probably

PR #3005, opened 2023-06-26
Feat/oidc-rp-logout

jeswr:
Supercedes #2980

There are 3 main aspects to this PR:
 - Refactoring some shared logic into *-core (in particular the `GeneralLogoutHandler` and `SessionManager`)
 - Introducing RP initiated logout
 - Migrating to npm workspaces (required because the introduction of authn-* libraries as a dependency of our ts-tools meant that we were not resolving to the development version of *-core when doing *-browser tests).


chelseapinka (approved):
Looks good :+1: Don't we need to do any additional doc updates for `ILogoutOptions` ?

PR #3020, opened 2023-06-28
chore: migrate to npm workspaces

jeswr:
Migrates to use npm workspaces


packages/browser/examples/single/bundle/package.json
```diff
@@ -27,6 +28,6 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "regenerator-runtime": "^0.13.9",
-    "@inrupt/solid-client-authn-browser": "file:../../../"
+    "@inrupt/solid-client-authn-browser": "^1.16.0"
```
chelseapinka, line 31: were all these example apps meant to use the developed changes tho? For developers who might need to change something and test it out in the package?

chelseapinka (approved):

packages/browser/examples/single/bundle/package.json
jeswr: Previously this was outside of the lerna workspace.

What I have done is made it part of the npm workspace; so it will resolve to the development version; and then marked the package as private so it doesn't get published.

Issue #3001, opened 2023-06-19
ESM distro for @inrupt/solid-client-authn-browser


riovir:
**UPDATE 2**: Here's a small [reproduction on Codesandbox](https://codesandbox.io/p/sandbox/vite-forked-vs5l4c) make sure to interact with the preview on a separate browser tab to avoid a login-loop.
**UPDATE**: Changed workaround to use a more idiomatic way to force ESM transitive dependencies for Rollup / Vite.

### Search terms you've used

- esm
- rollup
- vite

### Impacted environment

In which environment would the proposed feature apply?

- [x] The browser (with a bundler)
- [ ] Node.js
- [ ] Other (please specify): ...
- [ ] I'm not sure.

### Feature suggestion

The @inrupt/solid-client-authn-browser/package.json declares a valid ESM export. (With either an mjs extension or also the "type": "module" set).

### Expected functionality/enhancement

It would be possible to use the package with modern bundlers, such as Vite.

### Actual functionality/enhancement

When Vite does a production build the app seemingly works, but fails at runtime. The rabbit hole looks like this:
- Vite uses Rollup for production builds (instead of ESBuild used during development)
- Rollup concludes that @inrupt/solid-client-authn-browser is in CommonJS
- Then, when processing the transitive dependencies, eventually imports the `@inrupt/solid-client-authn-core` however, picks the CommonJS edition. This edition, has been written with Node.js in mind, trying to use the `node-fetch` instead of the readily available `fetch` in the browser.

The ideal way to break that chain is to offer an ESM edition of the `@inrupt/solid-client-authn-browser` so that bundlers won't try to pick the CJS / Node version of transitives.

### Use Cases

Trying to build a basic Vue.js app with Vite instead of Webpack. The Vue community has been slowly backing away from Webpack, instead putting their effort into Vite (ESBuild for dev, Rollup for prod builds). Unfortunately, without hacking into node_modules with `patch-package` I was unable to create a working production build.

### Additional information

As a workaround I explicitly list most of the transitive dependencies of `@inrupt/solid-client-authn-browser` and use a `customResolver` that's [forced to pick ESM versions](https://gitlab.com/riovir/babalog/-/blob/27d0f8e35c3c9e0d28af04f342416dae398f9f47/vite.config.ts#L23-26)

[...]

PR #3022, opened 2023-06-28
feat: expose *-authn-node and *-authn-browser as ESM

jeswr:
adds ESM support closing #3001

Supercedes #3019


packages/browser/package.json
```diff
@@ -5,18 +5,27 @@
...
+      "types": "./dist/index.d.ts"
+    }
+  },
   "scripts": {
     "prepublishOnly": "npm run build",
     "clean": "npm run clean-module && npm run clean-browser && rimraf ./coverage",
     "clean-module": "rimraf ./dist",
     "clean-browser": "rimraf ./browserDist",
-    "build": "npm run build-module && npm run build-browser",
+    "build": "npm run build-module && npm run build-browser && npm run build-esm",
```
NSeydoux, line 25: Let me look into more details about what `build-module` does, but it sounds that it should build ESM too.

NSeydoux (changes requested):

packages/node/rollup.config.mjs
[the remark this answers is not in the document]
jeswr: I did consider doing it this way.

The reason I chose not to do this is because (at least the way I had configured it), rollup was bundling all of the js into a single index.js file; which meant that we could cause breaking changes if any consumers are importing a particular file path into their code.

If we consider the public API to be just the root then this is not a breaking change and I am happy to apply it.

packages/browser/package.json
jeswr: I think it is just badly named since it is just executing `tsc` with [this config](https://github.com/inrupt/solid-client-authn-js/blob/main/packages/browser/tsconfig.json).

packages/node/rollup.config.mjs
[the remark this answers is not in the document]
NSeydoux: I see. Since our `package.json` only includes `"main": "dist/index.js"` for CJS exports, that's what I consider our public API. If someone is relying on a specific import, we can add it on a case-by-case basis as a named export the same way it is done in https://github.com/inrupt/solid-client-js/blob/a87e7e9084950cd656e16d2b4d14e79c350211ce/package.json#L39, but I would say it is okay to go ahead and use rollup for bundling both ESM and CJS, even if that changes the content of the `dist` folder if these don't contradict the endpoints exposed in the `package.json`.

packages/browser/package.json
NSeydoux: Then I assume it would make sense to also collapse all bundling into rollup, and output both CJS and ESM?

packages/browser/rollup.config.mjs
```diff
@@ -0,0 +1,39 @@
...
+const pkg = require("./package.json");
+
+export default [
+  ...createConfig(pkg),
+  {
+    input: "./src/index.ts",
+    output: [
+      {
+        file: pkg.bundle,
+        format: "iife",
```
NSeydoux, line 22: Ah good, that's non-breaking for the scrip tags imports ? if we can dump webpack and focus on rollup, that's great.

packages/node/rollup.config.mjs
```diff
@@ -0,0 +1,11 @@
...
+// import pkg from "./package.json" assert { type: "json" };
+
+// Until we only support Node 18+, this should be used instead
+// (see https://rollupjs.org/guide/en/#importing-packagejson)
+import { createRequire } from "node:module";
+import createConfig from '../../rollup.common.mjs';
+const require = createRequire(import.meta.url);
+const pkg = require("./package.json");
+
+export default createConfig(pkg);
```
NSeydoux, line 11: All the dependencies of the package need to be added to the `external` array, so that they are not bundled with the source code, and rather resolved as transitive dependencies by the consuming app.

packages/browser/rollup.config.mjs
```diff
@@ -0,0 +1,39 @@
...
+    ...sharedConfig,
+    external: [],
+    plugins: [
+      ...sharedConfig.plugins,
+      nodeResolve({
+        browser: true,
+        preferBuiltins: false
+      }),
+      commonjs(),
+      terser(),
```
NSeydoux, line 36: Can you add a comment to explain why these modules are required here?

packages/node/rollup.config.mjs
jeswr: https://github.com/inrupt/solid-client-authn-js/pull/3022/commits/2cc1c08c6adc63272711070ca85da40a68343b12 now gets them directly from the package.json dependencies

packages/browser/rollup.config.mjs
jeswr: Can confirm that https://github.com/inrupt/solid-client-authn-js/tree/main/packages/browser/examples/single/script works. Would it be worth adding a playwright test at some point for that one?

NSeydoux: Absolutely, can you create a ticket to track this?

NSeydoux:
That looks all good, but since this touches how we export our library, and that's not well-covered by our tests, I'd be more comfortabe if we did a beta release of this, published it on npm using the `@next` tag, and check it doesn't break things before merging it in main.

chelseapinka (approved):
:+1:

PR #3025, opened 2023-06-30
chore: remove unused typescript dev-dependencies

jeswr:
We have a typescript version pinned at the root. Having it in the
dev dependencies of other packages is just creating issues with
versioning conflicts.

Hopefully this will also stop https://github.com/inrupt/solid-client-authn-js/pull/3024 complaining.


NSeydoux (dismissed):
These example apps are not subprojects of the root project, and so they should have their dependencies self-contained. I'll have a look into this to see if there is another way to resolve this.

NSeydoux (approved):

PR #3027, opened 2023-07-02
feat: rp initiated logout

jeswr:
Supercedes #3005


packages/core/src/logout/endSessionUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+export interface IEndSessionOptions {
+  endSessionEndpoint: string;
+  idTokenHint?: string;
+  postLogoutRedirectUri?: string;
+  state?: string;
+}
```
NSeydoux, line 26: I think you could do something like 
```
export interface IEndSessionOptions {
  endSessionEndpoint: string;
  idTokenHint?: string;
  postLogoutRedirectUri?: string;
  state?: string;
}

export interface IEndSessionOptionsPostLogout extends IEndSessionOptions {
  postLogoutRedirectUri: string;
  state?: string;
}

export interface IEndSessionOptionsNoCallback extends IEndSessionOptions {
  postLogoutRedirectUri?: undefined;
  state?: undefined;
}
```
which would result in a compile-time error if "state" is specified but not `postLogoutRedirectUri`:
```
const test = (a: IEndSessionOptionsPostLogout | IEndSessionOptionsNoCallback) : void => {};
test({endSessionEndpoint: "test", state: "test"})
```
The names of the interfaces here are a bit too long, but they show the intent. I would typically do this inline with a union type, but I don't think interfaces allow such thing. Approaching it the same way as `ILogoutOptions` sounds fine to me.

packages/core/src/logout/endSessionUrl.ts
```diff
@@ -0,0 +1,53 @@
...
+  postLogoutRedirectUri,
+  state,
+}: IEndSessionOptions) {
+  const url = new URL(endSessionEndpoint);
+
+  if (idTokenHint) url.searchParams.append("id_token_hint", idTokenHint);
+
+  if (postLogoutRedirectUri) {
+    url.searchParams.append("post_logout_redirect_uri", postLogoutRedirectUri);
+    if (state) url.searchParams.append("state", state);
```
NSeydoux, line 49:
```suggestion
    if (state !== undefined) url.searchParams.append("state", state);
```

NSeydoux (changes requested):

packages/core/src/logout/ILogoutHandler.ts
```diff
@@ -26,8 +26,25 @@
 
 import type IHandleable from "../util/handlerPattern/IHandleable";
 
+export interface IRPLogoutOptions {
+  logoutType: "idp";
+  postLogoutUrl?: string | undefined;
```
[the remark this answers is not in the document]
jeswr: Not if consumers are using `--exactOptionalProperties` enabled (see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#exact-optional-property-types---exactoptionalpropertytypes)

packages/core/src/logout/ILogoutHandler.ts
```diff
@@ -26,8 +26,25 @@
...
+  state?: string | undefined;
```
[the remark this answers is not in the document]
jeswr: See above comment

packages/core/src/logout/endSessionUrl.ts
jeswr: I find that these kinds of patterns are generally more of a hindrance than a help if consumers because the TS type inference system just doesn't work well with a lot of these types. To provide a minimal example as to why; the following code

```ts
interface T {
  a: true;
  b: true;
}

interface F {
  a: false;
  b: false;
}

function test1({ a, b }: T | F) {
  return test2({ a, b })
}

function test2(options: T | F) {
  throw new Error("Not Implemented")
}
```

 results in a typescript compilation error

```ts
Argument of type '{ a: boolean; b: boolean; }' is not assignable to parameter of type 'T | F'.
  Type '{ a: boolean; b: boolean; }' is not assignable to type 'F'.
    Types of property 'a' are incompatible.
      Type 'boolean' is not assignable to type 'false'.ts(2345)
```

jeswr: In fact we get exactly this kind of error internally if we do try and implement such changes.
 - Commit implementing these changes: https://github.com/inrupt/solid-client-authn-js/commit/1c8cec3c9cbe4143cdb7e0b750222081e19cafc1
 - Resulting TS Error: https://github.com/inrupt/solid-client-authn-js/actions/runs/5459532806/jobs/9935630385

jeswr: I've added some jsdoc in https://github.com/inrupt/solid-client-authn-js/pull/3027/commits/c72ee18497a41277af3c5cd75621b73fe268379f to try and address this.

packages/core/src/logout/ILogoutHandler.ts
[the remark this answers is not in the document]
NSeydoux: Oh, interesting!

packages/core/src/logout/endSessionUrl.ts
NSeydoux: Well that's annoying. The following actually compiles for 1c8cec3, which I find ridiculous ^^
```
if (options.postLogoutUrl === undefined) {
        this.boundLogout({
          logoutType: "idp",
          postLogoutUrl: options.postLogoutUrl,
          state: options.state,
        });
      } else {
        this.boundLogout({
          logoutType: "idp",
          postLogoutUrl: options.postLogoutUrl,
          state: options.state,
        });
      }
```
Adding the docs is a good alternative solution.

jeswr: > The following actually compiles for https://github.com/inrupt/solid-client-authn-js/commit/1c8cec3c9cbe4143cdb7e0b750222081e19cafc1, which I find ridiculous ^^

This is something to take up with the TS team one day ...

jeswr:
> Do you want to go ahead, merge this, and implement the node part in a subsequent PR?

Let's do that - I've enabled automerge so this should merge as soon as it is approved.

chelseapinka (approved):
Looks good, the Sonar Cloud scans seem to want more test coverage but you can also add them in the subsequent PR that Nic mentioned in the comments.

NSeydoux (approved):

PR #3044, opened 2023-07-08
feat: implement RP initiated logut for node

jeswr:


.github/workflows/ci.yml
```diff
@@ -9,7 +9,19 @@ concurrency:
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20"
+          cache: "npm"
+      # Install dependencies:
+      - run: npm ci --ignore-scripts
+      # Build the packages
+      - run: npm run build
+      # Run the linting command:
+      - run: npm run lint
```
jeswr, line 24: This is required to stop this linting error https://github.com/inrupt/solid-client-authn-js/actions/runs/5497045924/jobs/10017431247

e2e/browser/test/fixtures.ts
```diff
@@ -241,6 +241,91 @@ const createClientResource = async (
 const clientApplicationUrl =
   process.env.E2E_DEMO_CLIENT_APP_URL ?? "http://localhost:3001/";
 
+export interface ISeedPodResponse {
```
chelseapinka, line 244: Not sure if I understand the naming of this type. What is an `ISeedPodResponse` abstracting, just the session used to access the pod we  are using in the test?

packages/node/src/Session.ts
chelseapinka, line 1: :clap:

packages/node/src/Session.spec.ts
chelseapinka, line 1: should we add a postive test case here for "succeeds when performing a idp type logout?" or would that overlap with the `e2e/node/e2e-spect.ts`?

chelseapinka (approved):
Overall looks good, couple of small questions/clarifications. Could use one more test case potentially but :+1:

.github/workflows/ci.yml
jeswr: I've opened up https://github.com/inrupt/typescript-sdk-tools/pull/260 which I think is cleaner than my original solution and eslintignoring.

e2e/browser/test/fixtures.ts
jeswr: The `seedPod` function generates a client ID document and a resource which only that application can access. The URIs of those to resources are the main things used; the session that is returned is used to interact with those resources in some cases.

packages/core/src/logout/RpLogoutHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    if (options.toLogoutUrl === undefined) {
+      throw new Error(
+        "Cannot perform IDP logout. Did you log in using the OIDC authentication flow?"
+      );
+    }
+
+    return this.redirector.redirect(options.toLogoutUrl(options), {
+      handleRedirect: options.handleRedirect,
+    });
```
NSeydoux, line 53: Since the return type is `Promise<void>`, I would have expected an `await` here rather than a return.

packages/core/src/logout/endSessionUrl.ts
```diff
@@ -52,3 +54,40 @@ export function getEndSessionUrl({
...
+}
+
+/**
+ * @param options.endSessionEndpoint The end_session_endpoint advertised by the server
+ * @param options.idTokenHint The idToken supplied by the server after logging in
+ * Redirects the window to the location required to perform RP initiated logout
+ *
+ * @hidden
+ */
+export function maybeBuildRpInitiatedLogout({
```
NSeydoux, line 86: Why not collapse these two functions into one?

NSeydoux (dismissed):

e2e/browser/test/fixtures.ts
```diff
@@ -241,6 +241,91 @@ const createClientResource = async (
...
+  session: Session;
+}
+
+export async function seedPod(
+  setupEnvironment: TestingEnvironmentNode
+): Promise<ISeedPodResponse> {
+  if (
+    setupEnvironment.clientCredentials.owner.type !== "ESS Client Credentials"
+  ) {
+    throw new Error("Unsupported client credentials");
```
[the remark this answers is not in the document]
jeswr: This is for type safety - we added CSS client credentials to our typescript-sdk-tools [here](https://github.com/inrupt/typescript-sdk-tools/blob/72f676b1f56300649336499e55fb36099f80cf91/packages/internal-test-env/index.ts#L55-L74) when we were working on adding e2e tests against the CSS.

e2e/node/express.ts
```diff
@@ -0,0 +1,115 @@
...
+      return res.status(400).send("oidcIssuer is required").end();
+    }
+
+    return session.login({
+      redirectUrl: "http://localhost:3001/redirect",
+      oidcIssuer,
+      clientId: typeof clientId === "string" ? clientId : undefined,
+      handleRedirect: (url) => res.redirect(url),
+    });
+  });
```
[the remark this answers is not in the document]
jeswr: This alert is not relevant as it is only used in tests.

e2e/node/e2e-test.spec.ts
```diff
@@ -64,6 +72,131 @@ function getCredentials() {
   };
 }
 
+describe("handleIncomingRedirect", () => {
```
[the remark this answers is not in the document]
jeswr: > using jest as a node runner but a plain script (potentially using express) that implements the server-side, and have playwright drive the test against that.

Good idea - I've implemented this in https://github.com/inrupt/solid-client-authn-js/pull/3044/commits/41a7c75feeca46fc50309fe5e6ae0ae636b55d4e .

However, I think we should have this in addition to the original node test as they test two distinct use cases (the original testing scripts with browser interaction components to login; the second which I just added testing apps with server-side rendering).

Would moving these tests to a new file resolve your concerns?

e2e/node/e2e-test.spec.ts
```diff
@@ -30,12 +30,19 @@ import {
 } from "@jest/globals";
 import { custom } from "openid-client";
 import {
+  getBrowserTestingEnvironment,
   getNodeTestingEnvironment,
   getPodRoot,
 } from "@inrupt/internal-test-env";
+import { CognitoPage, OpenIdPage } from "@inrupt/internal-playwright-helpers";
+import type { Request } from "@playwright/test";
+import { firefox } from "@playwright/test";
```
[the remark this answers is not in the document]
jeswr: Would moving these tests to a new file resolve your concerns (see https://github.com/inrupt/solid-client-authn-js/pull/3044#discussion_r1261095324)?

e2e/node/e2e-test.spec.ts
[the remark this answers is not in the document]
NSeydoux: Having these tests separate would be an improvement, but I'm challenging whether we need the former at all. Is "scripts with browser interaction components to login" a use case that is important enough that we add the complexity of this test to cover it, while it will rely on the same code paths as a regular Node web server? Because essentially, what happens is:
- The Node code calls `login`, and redirects the user away to the OP via their browser.
- The OP eventually redirects the user back to the node code with the redirect URL containing the auth code

I don't think embedding playwright in the Node tests is worth it. Having an example of script that does that type of redirect, why not, but having a full-fledged browser testing framework on top of jest just for this niche use case is overkill in my opinion.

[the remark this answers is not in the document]
jeswr: > "scripts with browser interaction components to login" a use case that is important enough that we add the complexity of this test to cover it

I've encountered this several times in the wild; in particular:
 - A similar flow is used to log into desktop applications like SolidFS; and can be used for logging into slack-style apps in the future.
 - A similar flow is being used by SolidLab to log into browser extensions
 - I've seen multiple people doing experiments with Solid are logging in with scripts this way; especially if they want to test against other Solid Servers that don't support client-creds in addition to testing against the ESS.

> while it will rely on the same code paths as a regular Node web server

I agree with this; the main differences between the two is that the browser is closed and re-opened between login and logout; and the app is redirects to the login/logout pages rather than opening the browser directly on those pages.

> overkill in my opinion.

I was viewing this in terms of whether the ongoing maintenance cost of this test would be worth it; I was imagining this would be fairly low since it is re-using existing components from the browser tests.

For now I will move this to a separate file; and we can remove these tests once you're back from break if they are still overkill / not worth the maintenance cost.

packages/node/src/Session.spec.ts
jeswr: I've added unit tests for the `RpLogoutHandler` in https://github.com/inrupt/solid-client-authn-js/pull/3044/commits/fedf8de00fcb1d22c847b17f7f74ce5fd8624966; and the e2e tests should cover the rest.

PR #3053, opened 2023-07-14
v1.17.0

### LINKED ISSUE #1815: import fails with esbuild
### Search terms you've used

- esbuild
- vite
- does not provide an export named
- esm

### Impacted package

- [x] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

`solid-client-authn-browser` appears to be not working with Vite when it is running esbuild. The build mode which is using Webpack works.

### To Reproduce
1. create Vite project and run the dev server: `npm init vite@latest`, `npm install` and `npm run dev`
2. add the following code to the `<script>` section of the file `App.vue`
```js
import { ref } from 'Vue'
import { Session } from "@inrupt/solid-client-authn-browser"
const test = ref('')
console.log(new Session())
```

### Expected result

Session is imported and can be used.

### Actual result

Error in browser console
```
Uncaught SyntaxError: The requested module '/node_modules/@inrupt/solid-client-authn-browser/dist/index.js?t=1638734365925' does not provide an export named 'Session'
```

### Environment

```
System:
    OS: macOS 12.0.1
    CPU: x64 Intel(R) Core(TM)
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.2.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.1.4 - /usr/local/bin/npm
  npmPackages:
    @inrupt/solid-client-authn-browser: ^1.11.2 => 1.11.2 
    @vitejs/plugin-vue: ^1.9.3 => 1.10.1 
    vite: ^2.6.4 => 2.6.14 
    vue: ^3.2.23 => 3.2.23 
  npmGlobalPackages:
    npm: 8.1.4
```

## Additional information

I believe the problem is that no esm version is shipped by the `solid-client-authn-browser`.

### LINKED ISSUE #3001: ESM distro for @inrupt/solid-client-authn-browser

### LINKED ISSUE #3052: Cannot build with versions >1.13.4
### Search terms you've used

polyfill, esbuild, undici

### Impacted package

- [x] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

Since version 1.14 @inrupt/solid-client-authn-browser seems to depend on a lot of node internals. I am getting several errors when trying to build my app with esbuild, like these:

```
✘ [ERROR] Could not resolve "perf_hooks"

    ../node_modules/undici/lib/fetch/util.js:5:32:
      5 │ const { performance } = require('perf_hooks')
        ╵                                 ~~~~~~~~~~~~

  The package "perf_hooks" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

✘ [ERROR] Could not resolve "assert"

    ../node_modules/undici/lib/fetch/body.js:19:23:
      19 │ const assert = require('assert')
         ╵                        ~~~~~~~~

  The package "assert" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

✘ [ERROR] Could not resolve "util/types"

    ../node_modules/undici/lib/fetch/body.js:21:48:
      21 │ const { isUint8Array, isArrayBuffer } = require('util/types')

```

It worked fine with v1.13.4. Newer version seem to have introduced universal-fetch, which depends on undici, that causes the errors:

```
▶ npm ls undici   
  +-- @inrupt/solid-client-authn-browser@1.16.0
  | `-- @inrupt/universal-fetch@1.0.2
  |   `-- undici@5.22.1
```

The README only mentions to install the buffer polyfill, which I did. If other polyfills are required with the new version this should be documented.

### To Reproduce

1. Install esbuild and @inrupt/solid-client-authn-browser@1.16.0
2. Create file test.js:

```
import { login } from "@inrupt/solid-client-authn-browser";

function test() {
  login();
}

```

3. try to bundle it using esbuild: `esbuild --bundle test.js`

### Expected result

Successful build.

### Actual result

Errors shown above.

### Environment

Please run

npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers

in your project folder and paste the output here:

```
$ npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers
  System:
    OS: Linux 6.2 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 4.02 GB / 31.05 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 19.6.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.4.0 - /usr/local/bin/npm
  npmPackages:
    @inrupt/solid-client-authn-browser: 1.16.0 => 1.16.0 
    buffer: ^6.0.3 => 6.0.3 
    esbuild: ^0.18.11 => 0.18.11 
  npmGlobalPackages:
    corepack: 0.15.3
    npm: 9.4.0

```

jeswr:
Closes https://github.com/inrupt/solid-client-authn-js/issues/1815
Closes https://github.com/inrupt/solid-client-authn-js/issues/3001
Closes https://github.com/inrupt/solid-client-authn-js/issues/3052

### DIFF
--- a/e2e/browser/test-app/package.json
+++ b/e2e/browser/test-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "test-app",
-  "version": "0.1.0",
+  "version": "1.17.0",
   "private": true,
   "scripts": {
     "dev": "next dev -p 3001",
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@inrupt/internal-playwright-testids": "^2.0.2",
-    "@inrupt/solid-client-authn-browser": "^1.16.0",
+    "@inrupt/solid-client-authn-browser": "^1.17.0",
     "next": "^13.4.7",
     "react": "^18.2.0",
     "react-dom": "^18.2.0"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,13 @@ The following have been deprecated, and will be removed in future major releases
 
 The following changes have been implemented but not released yet:
 
+## [1.17.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v1.17.0) - 2023-07-14
+
 ### New Feature
 
 - Support for [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) in Node and Browser
   libraries.
+- ESM support for Node and Browser packages.
 
 ## [1.16.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v1.16.0) - 2023-05-09
 
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
   "useNx": true,
-  "version": "1.16.0"
+  "version": "1.17.0"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,10 +57,10 @@
       }
     },
     "e2e/browser/test-app": {
-      "version": "0.1.0",
+      "version": "1.17.0",
       "dependencies": {
         "@inrupt/internal-playwright-testids": "^2.0.2",
-        "@inrupt/solid-client-authn-browser": "^1.16.0",
+        "@inrupt/solid-client-authn-browser": "^1.17.0",
         "next": "^13.4.7",
         "react": "^18.2.0",
         "react-dom": "^18.2.0"
@@ -23275,11 +23275,11 @@
     },
     "packages/browser": {
       "name": "@inrupt/solid-client-authn-browser",
-      "version": "1.16.0",
+      "version": "1.17.0",
       "license": "MIT",
       "dependencies": {
-        "@inrupt/oidc-client-ext": "^1.16.0",
-        "@inrupt/solid-client-authn-core": "^1.16.0",
+        "@inrupt/oidc-client-ext": "^1.17.0",
+        "@inrupt/solid-client-authn-core": "^1.17.0",
         "@inrupt/universal-fetch": "^1.0.2",
         "events": "^3.3.0",
         "jose": "^4.3.7",
@@ -23292,7 +23292,7 @@
     },
     "packages/browser/examples/demoAuthentiationMonitor": {
       "name": "demo-authentiation-monitor",
-      "version": "1.0.0",
+      "version": "1.17.0",
       "license": "MIT",
       "dependencies": {
         "express": "^4.18.2"
@@ -23300,10 +23300,10 @@
     },
     "packages/browser/examples/demoClientApp": {
       "name": "@inrupt/demo-app-authn-browser",
-      "version": "1.0.0",
+      "version": "1.17.0",
       "license": "MIT",
       "dependencies": {
-        "@inrupt/solid-client-authn-browser": "^1.16.0",
+        "@inrupt/solid-client-authn-browser": "^1.17.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "regenerator-runtime": "^0.13.11"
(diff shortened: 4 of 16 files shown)


e2e/browser/test-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "test-app",
-  "version": "0.1.0",
+  "version": "1.17.0",
```
chelseapinka, line 3: was it intentional to version the app this time, when it looks like it hasnt been bumped with previous releases?

chelseapinka (approved):
:+1:

e2e/browser/test-app/package.json
jeswr: This is the first release since moving to npm workspaces and the app was added to the workspace as part of that; so lerna versioned it.

This doesn't really matter because the manifest sets it as a private package so a release of it will not be made.

PR #3077, opened 2023-08-04
chore: migrate to use shared rollup config

jeswr:
Uses https://www.npmjs.com/package/@inrupt/base-rollup-config which we just released


chelseapinka (approved):
It really seems like those "fetch fails" for test cases are something we should look into more.

PR #3088, opened 2023-08-09
chore: clarify no postLogoutUrl redirect on dynamic registration

jeswr:


chelseapinka (approved):
This is better but we could also link to the client id registration page for users to learn more if they are still confused. (Thinking about it from first time implementing logout perspective)

chelseapinka:
Okay it makes sense that all the Dev 2-2 tests are failing due to the login flow issues, but interesting that all the "Not Logged In" test cases are also flaky :thinking:

chelseapinka (approved):

PR #3089, opened 2023-08-10
chore: re-enable skipped dependency tests

jeswr:
These tests were being skipped because they left open handles in jest due to the `setTimeout` calls. This resolves those issues.

I am somewhat surprised that we don't see these openHandles errors when trying to use the fully fledged session class in e2e tests driven by jest.


chelseapinka (approved):

chelseapinka:
Are we missing the newest playwright helpers to assist with this failing test case issue in Dev-2-2?

jeswr:
> Are we missing the newest playwright helpers to assist with this failing test case issue in Dev-2-2?

Correct - waiting for those to land in main so I can rebase.

PR #3099, opened 2023-08-23
Remove nodemon dependency

NSeydoux:
`nodemon` was just use in one example, and the provided value wasn't worth the maintenance overhead.


NSeydoux:
Note: the dev-2-2 browser-based tests will pass again once the playwright helpers have been updated.

jeswr (approved):

PR #3100, opened 2023-08-23
Uprade linting

NSeydoux:


package.json
```diff
@@ -55,6 +55,7 @@
     "dotenv-flow": "^3.2.0",
     "eslint": "^8.25.0",
     "eslint-config-next": "^13.4.13",
+    "eslint-plugin-prettier": "^5.0.0",
```
jeswr, line 58: We shouldn't need to have that as a dependency here as it is already a dependency of `@inrupt/eslint-config-base`

package.json
```diff
@@ -38,8 +38,8 @@
     "build-demo-app": "npm run build; cd packages/browser/examples/demoClientApp/; npm ci; npm run build; mkdir -p ../../../../dist/website/demo/; cp -r dist/. ../../../../dist/website/demo/"
   },
   "devDependencies": {
-    "@inrupt/eslint-config-base": "^2.0.4",
-    "@inrupt/eslint-config-lib": "^2.1.2",
+    "@inrupt/eslint-config-base": "^2.1.3",
```
jeswr, line 41: We should be able to get rid of this `@inrupt/eslint-config-base` dependency entirely as it is already a dependency of `@inrupt/eslint-config-lib`

package.json
NSeydoux: That's what I thought too, but ESLint complains about this missing if not installed, and the same setup is present in `solid-client`. Let me see if that could be related to your next comment.

NSeydoux: Well removing both dependencies doesn't cause the issue to come back. I also opened https://github.com/inrupt/solid-client-js/pull/2149 to do the same in `solid-client`

NSeydoux: See https://github.com/inrupt/solid-client-authn-js/pull/3100/commits/cdf7b4252cd52a74d5bcb3f2948f6939f931f119

package.json
NSeydoux: Fixed in https://github.com/inrupt/solid-client-authn-js/pull/3100/commits/cdf7b4252cd52a74d5bcb3f2948f6939f931f119

chelseapinka (approved):
So many commas

PR #3102, opened 2023-08-25
Feat/prompt

jaxoncreed:
# New feature description

Added the support to pass a `prompt` to the underlying oidc library. This was once a feature, but I think it might have been accidentally removed.

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. (No. I don't have access to update Inrupt docs)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The following have been deprecated, and will be removed in future major releases
 ## Unreleased
 
 The following changes have been implemented but not released yet:
+ - Added `prompt` as a login option.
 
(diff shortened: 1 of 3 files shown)


NSeydoux:
Hi @jaxoncreed , thanks for opening this, and i'm sorry it took me so long to respond. The `prompt` parameter is set by the library internally to `consent` for requests including the `offline_access` scope (which is requested by default so that the library can use the refresh token to maintain authentication), and to `none` for silent authentication on page reload. What is the use case for exposing this parameter through the public API?

jaxoncreed:
This would allow you to use the `create` prompt for servers that support it (https://openid.net/specs/openid-connect-prompt-create-1_0.html). Alternatively, we could have a separate function for `create`, but that might be confusing because not all Solid Servers support the `create` prompt.

PR #3104, opened 2023-08-28
Upgrade demo app dependencies

NSeydoux:


packages/browser/examples/single/bundle/client-app-profile.jsonld
```diff
@@ -4,5 +4,6 @@
   "client_name": "solid-client-authn-browser demo",
   "redirect_uris": ["http://localhost:3113/"],
   "scope" : "openid offline_access webid",
-  "grant_types" : ["refresh_token","authorization_code"]
+  "grant_types" : ["refresh_token","authorization_code"],
+  "token_endpoint_auth_method": "none"
 }
```
jeswr, line 9: Can you add the missing EOL char while you're at it?

jeswr (approved):

packages/browser/examples/single/bundle/client-app-profile.jsonld
NSeydoux: https://github.com/inrupt/solid-client-authn-js/pull/3104/commits/373fa973d97ee32b45c6fb331b200d2a48e5ba07

PR #3105, opened 2023-08-28
Upgrade playwright helpers to fix browser tests

NSeydoux:
The browser-based tests helpers have a bug preventing them to run against dev-2-2 that is patched in 2.1.3.

### DIFF
--- a/e2e/browser/test/fixtures.ts
+++ b/e2e/browser/test/fixtures.ts
@@ -23,6 +23,7 @@
 /* eslint-disable camelcase */
 
 import { test as base } from "@inrupt/internal-playwright-helpers";
+import { File } from "node:buffer";
 
 import { randomUUID } from "crypto";
 import type {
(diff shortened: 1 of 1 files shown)


e2e/browser/test/fixtures.ts
```diff
@@ -83,23 +84,48 @@ export type TestContainer = {
...
+    } catch (e: unknown) {
+      tries += 1;
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise((resolve) => {
+        setTimeout(resolve, interval);
+      });
+    }
+  }
+  throw new Error(
+    `An async callback is still failing after ${maxRetries} retries.`,
```
jeswr, line 108: Can you also include the 5 error messages in this so that we can see what goes wrong when it fails after 5 retries.

jeswr (approved):

e2e/browser/test/fixtures.ts
NSeydoux: Good idea: https://github.com/inrupt/solid-client-authn-js/pull/3105/commits/ddb4d57c65aef58d25bb64594080f328cede001e

PR #3114, opened 2023-08-28
Convert solid-ui-ract smoke tests to nextjs/playwright

NSeydoux:
This started as a fix to GHSA-rp65-9cf3-cjxr, but one thing led to another, `overrides` didn't work, I got carried away and rewrote the tests using the same setup as out other apps for consistency (and also removing the audit issue, which was most likely a false positive but hey).


e2e/browser/solid-client-authn-browser/test/fixtures.ts
jeswr, line 1: For retries on file creation, try and use a different file name each time so that the retries are not at risk of creating conflicts.

e2e/browser/solid-client-authn-browser/test/fixtures.ts
```diff
@@ -139,45 +137,36 @@ const createClientIdDoc = async (
...
-      new File([JSON.stringify(clientIdDoc)], "clientId", {
-        type: "application/json",
-      }),
-      {
-        fetch: session.fetch,
-      },
-    ),
+    overwriteFile(clientId, new Blob([JSON.stringify(clientIdDoc)]), {
+      fetch: session.fetch,
+    }),
```
jeswr, line 169: I would recommend specifying content type here.

e2e/browser/solid-client-authn-browser/test-app/components/appContainer/index.tsx
```diff
@@ -36,7 +36,8 @@ import {
 } from "@inrupt/solid-client-authn-browser";
 import AuthenticatedFetch from "../authenticatedFetch";
 
-const REDIRECT_URL = new URL("http://localhost:3001/").href;
+// Importing the playwright config file crashes the Next Web server.
```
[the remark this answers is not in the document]
jeswr: Can you make a shared constants file that both this and the playwright config use?

e2e/browser/solid-ui-react/test/fixtures.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    const pods = await getPodUrlAll(session.info.webId, {
+      fetch: session.fetch,
+    });
+
+    const resourceContent = "This is a plain piece of text";
+    // Create the container:
+    const testResource = await saveFileInContainer(
+      // Usually there's only a single Pod URL on the test accounts, so this *should* be fine:
```
jeswr, line 96: I'd suggest just having an `if (pods.length !== 1) throw new Error("This test suite can only handle pods with exactly one storage")` somewhere above.

playwright.shared.config.ts
```diff
@@ -78,6 +64,4 @@ const config: PlaywrightTestConfig = {
     //   },
     // },
   ],
-};
-
-export default config;
+};
```
jeswr, line 67: Can we add the newline char at the end of this file.

e2e/browser/solid-ui-react/test/solidui.playwright.ts
```diff
@@ -0,0 +1,37 @@
...
+
+test("solid-ui-react smoke test", async ({ app, testResource, auth }) => {
+  await app.home();
+  await expect(app.sampleRequest()).resolves.toEqual("Sample response");
+
+  await auth.login();
+  await expect(app.getWebid()).resolves.not.toBeNull();
+  await expect(app.fetchResource(testResource.url)).resolves.toEqual(
+    testResource.expectedContent,
+  );
```
jeswr, line 36: Can we also run this before login so we are testing the unauthenticated fetch for solid-ui-react.

jeswr (approved):
Overall looks pretty good - before merging could you please test to make sure the test suite *fails* when on version [v1.17.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v1.17.0) of the library. (Ensuring that it breaks when removing some of the bound fetches introduced [between 1.17.0 and 1.17.1](https://github.com/inrupt/solid-client-authn-js/compare/v1.17.0...v1.17.1) is sufficient).

.github/workflows/e2e-browser.yml
```diff
@@ -33,31 +33,29 @@ jobs:
       - name: Install e2e tests dependencies
         run: npx playwright install --with-deps
         if: github.actor != 'dependabot[bot]'
-      - name: Run browser-based end-to-end tests
```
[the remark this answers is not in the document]
NSeydoux: I find it more informative to actually display the command ran, in particular when the following step has the same name :)

e2e/browser/solid-client-authn-browser/test/fixtures.ts
NSeydoux: Agreed, I would actually take this one step further and remove all file names from tests. Let the server give a guuid to everything, these names are only a minor convenience when debugging a test, but even then they aren't game changer.

NSeydoux: 654f08b633ad762c47bed8fa48d4ba92fb3df396

e2e/browser/solid-ui-react/test/fixtures.ts
NSeydoux: 3f7bb3a302f2dafb49939c8dcdf7f9419af1255f

Issue #2985, opened 2023-06-15
Strip `iss` from incoming redirect URL callback in Node.js version


joachimvh:
### Impacted package

Which packages do you think might be impacted by the bug ?

- [ ] solid-client-authn-browser
- [x] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

This is the same issue as https://github.com/inrupt/solid-client-authn-js/issues/2103, which was fixed in https://github.com/inrupt/solid-client-authn-js/pull/2584. But the fix only solved this for the browser version of the package. The Node.js version of the source runs into the same problem.

I ran into this when upgrading the `oidc-provider` library for CSS to the latest version. The internal tests started failing as those use the Node.js version of the library.

jeswr:
Thanks for reporting @joachimvh - we will look into this!

NSeydoux:
The browser module manages the URL the browser is redirected to though the browser APIs, which is why we can clean up the OpenID query params from the redirect URL. This isn't the case in the Node library, where the URL management is left to the Web framework used by the app. Restoring a previous session is implemented differently in Node, since there is a secure storage where the refresh token can be stored. Therefore, there is no front-channel redirect relying on a redirect URL, which was causing the issue in https://github.com/inrupt/solid-client-authn-js/issues/2103.

In other words, the application code must have already been managing the redirect URL and cleaning up the OpenID parameters, so maybe this is out of scope for `solid-client-authn-node` ?

Could you elaborate a bit on steps to reproduce the issue, so that I understand if I'm on the right track or completely mistaken?

joachimvh:
The problem occurs since updating the `oidc-provider` library to v8. This is done on the branch in this PR, so you can use that to test it out: https://github.com/CommunitySolidServer/CommunitySolidServer/pull/1662

I've just tested it out, the problem can be reproduced by the `demoClientApp` in `packages/node/examples/demoClientApp`.

[...]

PR #3225, opened 2023-11-08
SDK-3246: full support for RFC9207 in node (`iss` parameter)

NSeydoux:
This PR fixes bug #2985 .

The main issue is that `openid-client` will compare the provided callback URL to the URL provided in the Authorization request, and we weren't cleaning up the `iss` parameter from the effective callback URL, causing a mismatch.

This include some refactoring, moving the openid cleanup util from the browser module to the code module. I've also removed a related test for the openid params specific to the implicit flow, which we don't support, and code to implement the actual check of RFC 9207, which is supported by `openid-client`.

### DIFF
--- a/packages/core/src/login/oidc/redirectIriUtils.ts
+++ b/packages/core/src/login/oidc/redirectIriUtils.ts
@@ -34,3 +34,16 @@ export function isValidRedirectUrl(redirectUrl: string): boolean {
     return false;
   }
 }
+
+export function removeOpenIdParams(redirectUrl: string): string {
+  const cleanedUpUrl = new URL(redirectUrl);
+  // For auth code flow
+  cleanedUpUrl.searchParams.delete("state");
+  cleanedUpUrl.searchParams.delete("code");
+  // For login error
+  cleanedUpUrl.searchParams.delete("error");
+  cleanedUpUrl.searchParams.delete("error_description");
+  // For RFC9207
+  cleanedUpUrl.searchParams.delete("iss");
+  return cleanedUpUrl.href;
+}
--- a/packages/core/src/login/oidc/redirectIriUtils.ts
+++ b/packages/core/src/login/oidc/redirectIriUtils.ts
@@ -35,7 +35,7 @@ export function isValidRedirectUrl(redirectUrl: string): boolean {
   }
 }
 
-export function removeOpenIdParams(redirectUrl: string): string {
+export function removeOpenIdParams(redirectUrl: string): URL {
   const cleanedUpUrl = new URL(redirectUrl);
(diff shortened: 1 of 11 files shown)


packages/node/src/login/oidc/incomingRedirectHandler/AuthCodeRedirectHandler.spec.ts
```diff
@@ -471,7 +471,10 @@ describe("AuthCodeRedirectHandler", () => {
     it("throws if the iss parameter does not match stored issuer", async () => {
       const mockedStorage = mockDefaultRedirectStorage();
       const mockedTokens = mockDpopTokens();
-      setupOidcClientMock(mockedTokens);
+      // oidc-client will throw if the iss parameter mismatches.
+      setupOidcClientMock(mockedTokens, () => {
```
NSeydoux, line 475: It is a bit unfortunate to have to do this, rather than mocking out the responses from the OpenID Provider to `oidc-client`

packages/core/src/login/oidc/redirectIriUtils.ts
```diff
@@ -34,3 +34,16 @@ export function isValidRedirectUrl(redirectUrl: string): boolean {
...
+  const cleanedUpUrl = new URL(redirectUrl);
+  // For auth code flow
+  cleanedUpUrl.searchParams.delete("state");
+  cleanedUpUrl.searchParams.delete("code");
+  // For login error
+  cleanedUpUrl.searchParams.delete("error");
+  cleanedUpUrl.searchParams.delete("error_description");
+  // For RFC9207
+  cleanedUpUrl.searchParams.delete("iss");
+  return cleanedUpUrl.href;
```
jeswr, line 48: Is there a reason that the `id_token` and `access_token` are no longer being stripped? It appears that the test related to these parameters has been removed.

jeswr: Also, why not make this function just return a `URL` since you seem to just be wrapping most invocations of this function back in a `new URL`

jeswr (approved):

packages/core/src/login/oidc/redirectIriUtils.ts
NSeydoux: > Is there a reason that the id_token and access_token are no longer being stripped? It appears that the test related to these parameters has been removed.

These are only relevant for the implicit flow, which we don't support in the first place. Since we don't do the implicit flow authorization request, we don't need to handle the response from it.

> Also, why not make this function just return a URL since you seem to just be wrapping most invocations of this function back in a new URL

Good point, will do.

PR #3273, opened 2023-12-15
Node 16 EoL: Upgrade dependencies

NSeydoux:
- Upgrade jose to v5
- Upgrade dotenv-flow to v4
- Upgrade Next to v14


jeswr (approved):
Would you also be able to remove unused imports (in particular the rollup ones).

Running `npx depcheck` should also give you a good idea if there are any other extraneous dependencies.

PR #3276, opened 2023-12-18
Node16 EoL: Upgrade rollup

NSeydoux:
This uses the latest rollup shared configuration.

Note that these changes have been locally tested by building the code, publishing it to a local verdaccio registry, and running a demo app against the local dependency to verify that we aren't breaking things.


jeswr (approved):

jeswr:
Might be worth adding a changelog entry for this.

chelseapinka (approved):

PR #3290, opened 2023-12-18
Remove deprecated APIs

NSeydoux:
This removes APIs marked for deprecation: 
- `EventEmitter` extended by `Session`, and related Session constructor parameter.
- Remove `useEssSession` from type signature
- Make `getClientAuthenticationWithDependencies` private

### DIFF
--- a/packages/node/src/dependencies.ts
+++ b/packages/node/src/dependencies.ts
@@ -102,7 +102,7 @@ export const buildRedirectHandler = (
 /**
  *
  * @param dependencies
- * @deprecated This function will be removed from the external API in an upcoming release.
+ * @hidden
  */
 export function getClientAuthenticationWithDependencies(dependencies: {
   secureStorage?: IStorage;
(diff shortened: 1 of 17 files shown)


packages/node/src/dependencies.ts
```diff
@@ -102,7 +102,7 @@ export const buildRedirectHandler = (
 /**
  *
  * @param dependencies
- * @deprecated This function will be removed from the external API in an upcoming release.
+ * @hidden
```
jeswr, line 105: I would suggest keeping the `@deprecated` jsdoc as well as hidden will only hide it from our documentation - not the exports iirc.

jeswr (approved):

packages/node/src/dependencies.ts
NSeydoux: Correct me if I'm wrong, but given this is no longer exported from either `index.ts`, nor is `dependencies.ts` mapped in the `package.json` exports, then in order to import this you'd need to import it directly from `@inrupt/solid-client-authn-node/src/dependencies`, which you can technically do but isn't public API, right?

PR #3380, opened 2024-02-29
Support auth code flow for statically registered clients

NSeydoux:
# New feature description

This adds support for the authorization code flow in statically registered clients. Which grant is applied to a given request is determined via a Handler pattern: a handful of OpenID handlers are registered (one for each flow, i.e. authorization code and client credentials in our case), and each is applied by priority order to figure out if they should process the response. The Client Credentials handler is higher priority than the Authorization code flow one, so the issue was that it was handling more requests than it should have. The fix was to prevent the Client Credentials handler to process requests coming in with a redirect URL: static clients without redirect URL will keep relying on the Client Credentials flow, while static clients with a redirect URL are now handled by the Authorization Code handler. Public clients (for which the Client ID and Client Secret are not specified in the request, and rely on Dynamic Client Registration) are not impacted by this change. 

# Checklist

### DIFF
--- a/packages/core/src/login/oidc/IClient.ts
+++ b/packages/core/src/login/oidc/IClient.ts
@@ -24,17 +24,21 @@
  * @packageDocumentation
  */
 
-export type ClientType = "static" | "dynamic" | "solid-oidc";
-
 type ISolidOidcClient = {
   clientId: string;
   clientType: "solid-oidc";
 };
 
-type IOpenIdClient = {
+type IOpenIdConfidentialClient = {
   clientId: string;
   clientSecret: string;
-  clientType: "static" | "dynamic";
+  clientType: "static";
+};
+
+type IOpenIdPublicClient = {
+  clientId: string;
+  clientSecret?: string;
+  clientType: "dynamic";
 };
 
 /**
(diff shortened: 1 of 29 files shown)


packages/core/src/login/oidc/IClient.ts
```diff
@@ -24,15 +24,27 @@
...
+  clientId: string;
+  clientSecret: string;
+  clientType: "static";
+};
+
+type IOpenIdPublicClient = {
+  clientId: string;
+  clientSecret?: string;
+  clientType: "dynamic";
+};
```
NSeydoux, line 42: This cleans up the supported client types and the associated assumptions.

packages/core/src/login/oidc/IClientRegistrar.ts
```diff
@@ -95,8 +101,22 @@ export async function handleRegistration(
...
+    };
+  } else if (isStaticClient(options)) {
+    clientInfo = {
+      clientId: options.clientId,
+      clientSecret: options.clientSecret,
+      clientName: options.clientName,
+      clientType: "static",
+    };
+  } else {
+    // Case of a dynamically registered client.
```
NSeydoux, line 119: Changing to type guards here allows to remove type assertions below.

chelseapinka (approved):

edwardsph (approved):
LGTM

PR #3405, opened 2024-03-15
apidocs: use common requirements instead of maintaining individual ones

kay-kim:
Use common requirements.txt from docs-assets repo instead of maintaining individual ones in the client lib repos.
I left the `dependabot.yml` alone as it seems, from the solid-client repo, we put the pip stuff back.


NSeydoux (approved):
Tested locally (since preview is broken), all seems well.

Issue #2348, opened 2022-09-23
[node] Using an authenticated session across multiple processes eg. Service + MQ Worker


elf-pavlik:
We are planning to add a message queue with a worker to https://github.com/janeirodigital/sai-impl-service/
It will handle things like subscribing to [Solid Notifications - Webhook](https://github.com/solid/notifications/blob/main/webhook-subscription-2021.md) and other background jobs that will require an authenticated session.

I wanted to ask if we should expect any issues when the authenticated session gets used by more than one node process. I haven't dug into how refreshing of the tokens is being scheduled but I can imagine some potential issues if two processes would try refreshing tokens without coordinating it via storage. Especially when the refresh token is rotated.

NSeydoux:
This library has not been designed for multiple processes being authenticated using the same session concurrently. In order for this to be successful, you'll have to be careful about how the `Storage` implementation that gets passed to the library handles the refresh token. One approach that would probably work (but that's just me thinking out loud) is if each process gets the session from storage before performing an authenticated request, which will enforce the latest refresh token being used (since the refresh token may be rotated on each usage). Concurrency will need to be managed very carefully.

elf-pavlik:
I think each process still needs an instance of `Session` which gets created from serialized data in `Storage`. Probably there might need to be a way to disable refreshing tokens by each session and instead enable re-hydrating the session from the storage instead. In that case, there would be one instance of the `Session` which would still be refreshing the token, would it keep serialized tokens in the `Storage` valid? (also thinking out loud)

NSeydoux:
Re-hydrating the session from storage is actually based on the refresh token being present: the access token is short-lived, and it is a sensitive piece of data, so it isn't persisted in storage. This means re-hydrating the session necessarily refreshes the token (and rotates the refresh token).

elf-pavlik:
If each process has to use refresh-token (which results in it being rotated). I don't really see how the session could be used across multiple processes. At least not without some pesky locking.

[Looking at CSS' OP config](https://github.com/CommunitySolidServer/CommunitySolidServer/blob/17d0413923cee27ee017b19d372be7e1da8fd365/config/identity/handler/provider-factory/identity.json#L46-L55)

It sets the Access Token expiry to 1h and the Refresh token to 24h. Having an option that opts-in to store the Access Token in the storage would make it easier to share the session across processes. Only one process would be responsible for refreshing it, possibly using a scheduler.

[...]

PR #3473, opened 2024-04-28
Allow to disable session background refresh

NSeydoux:
# New feature description

In a server-side scenario, one may not want sessions to proactively refresh to keep their access token active. Instead, the session si logged in using the refresh token when loaded from storage, but it will only be active during the lifetime of its Access Token.

When building a Session, a new option "keepAlive" is supported. It defaults to 'true' for backwards compatibility. If set to 'false', the refresh token of the session will not be used to proactively refresh the access token. `keepAlive` is persisted in storage with the session state so that it is used after the authorization request during the token request in the authorization code flow.

The default value is set in multiple places, which is a bad thing. This can be fixed later.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ The following changes have been implemented but not released yet:
   parameter is added to the constructor: `Session({ keepAlive: false })`. This prevents
   the `Session` setting a callback to refresh the Access Token before it expires, which
   could cause a memory leak in the case of a server-side application with many users, in
-  addition to unnecessary requests sent to the OpenID Provider. 
+  addition to unnecessary requests sent to the OpenID Provider.
 
 ## [2.1.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v2.1.0) - 2024-03-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,11 @@ The following changes have been implemented but not released yet:
 
 #### node
 
-- It is now possible to prevent a `Session` to self-refresh in NodeJS. To do so, a new
+- It is now possible to prevent a `Session` self-refreshing in NodeJS. To do so, a new
   parameter is added to the constructor: `Session({ keepAlive: false })`. This prevents
   the `Session` setting a callback to refresh the Access Token before it expires, which
-  could cause a memory leak in the case of a server-side application with many users, in
-  addition to unnecessary requests sent to the OpenID Provider.
+  could cause a memory leak in the case of a server-side application with many users.
+  It also avoids unnecessary requests being sent to the OpenID Provider.
 
 ## [2.1.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v2.1.0) - 2024-03-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes have been implemented but not released yet:
 
+### New Feature
+
+#### node
+
+- It is now possible to prevent a `Session` to self-refresh in NodeJS. To do so, a new
+  parameter is added to the constructor: `Session({ keepAlive: false })`. This prevents
+  the `Session` setting a callback to refresh the Access Token before it expires, which
+  could cause a memory leak in the case of a server-side application with many users, in
+  addition to unnecessary requests sent to the OpenID Provider. 
+
 ## [2.1.0](https://github.com/inrupt/solid-client-authn-js/releases/tag/v2.1.0) - 2024-03-13
 
 ### New Feature
--- a/packages/core/src/Session.ts
+++ b/packages/core/src/Session.ts
@@ -1,0 +1,24 @@
+//
+// Copyright Inrupt Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal in
+// the Software without restriction, including without limitation the rights to use,
+// copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+// Software, and to permit persons to whom the Software is furnished to do so,
+// subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+export type SessionConfig = {
+  keepAlive: boolean;
+};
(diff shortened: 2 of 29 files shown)


packages/core/src/Session.ts
```diff
@@ -0,0 +1,24 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+export type SessionConfig = {
```
NSeydoux, line 22: This may be used to add additional config entries in the future.

packages/core/src/login/ILoginOptions.ts
```diff
@@ -53,4 +53,6 @@ export default interface ILoginOptions extends ILoginInputOptions {
    * Event emitter enabling calling user-specified callbacks.
    */
   eventEmitter?: EventEmitter;
+
+  keepAlive?: boolean;
```
NSeydoux, line 57: `ILoginOptions` is an internal type, this is not exposed on the `Session::login` function. `keepAlive` is only set when constructing the `Session` object.

packages/node/src/login/oidc/incomingRedirectHandler/AuthCodeRedirectHandler.ts
```diff
@@ -169,7 +170,7 @@ export class AuthCodeRedirectHandler implements IIncomingRedirectHandler {
     }
     const authFetch = await buildAuthenticatedFetch(tokenSet.access_token, {
       dpopKey,
-      refreshOptions,
+      refreshOptions: oidcContext.keepAlive ? refreshOptions : undefined,
```
NSeydoux, line 173: This is a key part of this PR: is `keepAlive` is set to false, we initialize the authenticated fetch as if there was no refresh token, so that it doesn't set a timeout to refresh the access token before it expires.

This part of the code runs at the point of the token request in the authorization code flow (i.e. after the resource owner has been redirected back to the relying party)

packages/node/src/login/oidc/oidcHandlers/RefreshTokenOidcHandler.ts
```diff
@@ -93,7 +94,7 @@ async function refreshAccess(
     };
     const authFetch = await buildAuthenticatedFetch(tokens.accessToken, {
       dpopKey,
-      refreshOptions: rotatedRefreshOptions,
+      refreshOptions: keepAlive ? rotatedRefreshOptions : undefined,
```
NSeydoux, line 97: This is the same as above, but it runs when logging a session in that has a refresh token. This is typically the flow used to reactivate a session pulled from storage: the user may not be present, so the auth code flow is not appropriate. If a refresh token is present, we go through the refresh flow, which results in a new access token to be issued.

packages/node/src/Session.ts
```diff
@@ -70,6 +71,10 @@ export interface ISessionOptions {
    * An instance of the library core. Typically obtained using `getClientAuthenticationWithDependencies`.
    */
   clientAuthentication: ClientAuthentication;
+  /**
+   * A boolean flag indicating whether a session should be constantly kept alive in the background.
+   */
+  keepAlive: boolean;
```
NSeydoux, line 77: Note: this should be added to our prose docs.

elf-pavlik:
Will this change to some degree address #2348 ?

packages/core/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
```diff
@@ -94,7 +104,11 @@ export default abstract class AuthorizationCodeWithPkceOidcHandlerBase {
         issuer: oidcLoginOptions.issuer.toString(),
         // The redirect URL is read after redirect, so it must be stored now.
         redirectUrl: oidcLoginOptions.redirectUrl,
-        dpop: oidcLoginOptions.dpop ? "true" : "false",
+        dpop: Boolean(oidcLoginOptions.dpop).toString(),
+        keepAlive: booleanWithFallback(
+          oidcLoginOptions.keepAlive,
+          true,
```
jholleran, line 110: The default `true` seems to be set in multiple places. Its also set [here](https://github.com/inrupt/solid-client-authn-js/pull/3473/files#diff-659ac87dfd3711f4cfcea3c7bf6970980f4740fd59df45f04c7977bffaa23e98R86), [here](https://github.com/inrupt/solid-client-authn-js/pull/3473/files#diff-457298190cdad0e611f3ced3efa5d7dbcc9c199d3425ef0803ecfa1fc0c7216bR164) and [here](https://github.com/inrupt/solid-client-authn-js/pull/3473/files#diff-12a917a86414ec22d88e79dbb0dbc0c7a1f578d55c0686b36a582eb78fe9e083R183). Consider having it set in one place.

packages/node/src/ClientAuthentication.ts
```diff
@@ -31,6 +31,7 @@ import {
 import type {
   ILoginInputOptions,
   ISessionInfo,
+  SessionConfig
```
langsamu, line 34:
```suggestion
  SessionConfig,
```

NSeydoux:
> Will this change to some degree address https://github.com/inrupt/solid-client-authn-js/issues/2348 ?

That's a good question, I didn't think about it this way, but it might help. You'll definitely want to set `keepAlive` to false if you want to use a given session in two different processes. There may still be some synchronization issues depending on how the storage is accessed and the token rotated, but I think it might indeed support what you are describing. Let me know if it does!

jholleran:
> Will this change to some degree address #2348 ?

Hi @elf-pavlik, Nic would be best placed to answer your question as he would have more context on this. He is on vacation right now but I'll ask him to respond to your question when he is back. Is that ok for you?

elf-pavlik:
Hi @jholleran, Nic already responded in https://github.com/inrupt/solid-client-authn-js/pull/3473#issuecomment-2085218535
Whenever I get a chance to get back to that issue I will follow up in #2348 

Thanks!

packages/core/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
jholleran: This can be left as is for now and it can be a future tidy up.

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 
+### New Feature
+
+#### node
+
+- It is now possible to prevent a `Session` to self-refresh in NodeJS. To do so, a new
+  parameter is added to the constructor: `Session({ keepAlive: false })`. This prevents
+  the `Session` setting a callback to refresh the Access Token before it expires, which
+  could cause a memory leak in the case of a server-side application with many users, in
+  addition to unnecessary requests sent to the OpenID Provider.
```
edwardsph, line 20:
```suggestion
- It is now possible to prevent a `Session` self-refreshing in NodeJS. To do so, a new
  parameter is added to the constructor: `Session({ keepAlive: false })`. This prevents
  the `Session` setting a callback to refresh the Access Token before it expires, which
  could cause a memory leak in the case of a server-side application with many users.
  It also avoids unnecessary requests being sent to the OpenID Provider.
```

packages/core/src/login/oidc/IOidcOptions.ts
```diff
@@ -65,6 +65,11 @@ export interface IOidcOptions {
   redirectUrl?: string;
   handleRedirect?: (url: string) => unknown;
   eventEmitter?: EventEmitter;
+  /**
+   * Should the resulting session be refreshed in the background? This is persisted prior redirection.
```
edwardsph, line 69:
```suggestion
   * Should the resulting session be refreshed in the background? This is persisted prior to redirection.
```

edwardsph (commented):
Couple of text change suggestions

edwardsph (approved):
LGTM

Issue #3518, opened 2024-05-31
refresh tokens being stored in local storage


VirginiaBalseiro:
### Search terms you've used

* refresh token
* local storage

### Impacted package

Which packages do you think might be impacted by the bug ?

- [x] solid-client-authn-browser
- [ ] solid-client-authn-node
- [ ] solid-client-authn-core
- [ ] oidc-client-ext
- [ ] Other (please specify): ...

### Bug description

Refresh tokens are being stored in local storage, posing a significant security risk.

### To Reproduce

1. Log in to any app using this library (I tested with two different apps from different developers and could reproduce this issue, one of them is amc.inrupt.com which I assume is on v2.0 of this library).
2. Open local storage.
3. Wait approximately 5 minutes.
4. Observe the refresh token stored in local storage.

### Expected result

Refresh tokens should generally not be stored in local storage. This is also outlined in the security policy of this repository.

### Actual result

Refresh tokens are being stored in local storage.

### Environment

Please run

npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers

in your project folder and paste the output here:

```
$ npx envinfo --system --npmPackages --binaries --npmGlobalPackages --browsers
```

## Additional information

The presence of refresh tokens in local storage can lead to severe security breaches, including unauthorized access to user accounts. Immediate action is required to mitigate this risk and ensure the safety and security of users' data.

### Screenshot

![screenshot](https://github.com/inrupt/solid-client-authn-js/assets/28412960/a088198b-55ef-4aff-b32b-6c444b9711a8)

NSeydoux:
Hi @VirginiaBalseiro , 

Thanks for reporting this. Indeed, AMC is on v2.0.0 of the library: https://github.com/inrupt/authorization-management-component/blob/9b8976a9bd153b6cb4622b04d31d569fe16d1b05/package.json#L33.

The refresh token should no be stored in the local storage. However, by default, the session is DPoP-bound, which means without the DPoP key, the refresh token is unusable. The DPoP key not being stored in local storage, this mitigates the severity of the issue (which should be fixed nonetheless).

NSeydoux:
v2.2.1 should resolve this issue. Let us know if you see it persisting!

PR #3519, opened 2024-06-03
3518: Do not store refresh token in local storage

NSeydoux:
The local storage is not a secure storage, and as such the refresh token should not be stored there. This issue is mitigated by the token being DPoP-bound, and the DPoP key not being available in storage.

This PR fixes bug #3518 .


edwardsph (approved):
LGTM

PR #3540, opened 2024-06-11
SDK-3337: parse clientId from ID Token `azp` claim

garciafdezpatricia:
This PR fixes `clientId` not being parsed from the token payload.


packages/oidc-browser/src/refresh/refreshGrant.ts
```diff
@@ -122,7 +122,7 @@ export async function refresh(
     response,
     dpopKey !== undefined,
   );
-  const webId = await getWebidFromTokenPayload(
+  const { webId } = await getWebidFromTokenPayload(
```
garciafdezpatricia, line 125: Is it necessary to retrieve the `clientId` and return it later (line 131) as well?

packages/core/src/util/token.spec.ts
```diff
@@ -211,15 +211,59 @@ describe("getWebidFromTokenPayload", () => {
...
+    );
+    const { webId, clientId } = await getWebidFromTokenPayload(
+      jwt,
+      "https://some.jwks",
+      "https://some.issuer",
+      "https://some.clientId",
+    );
+    expect(webId).toBe("https://some.webid#me");
+    expect(clientId).toBe("some client");
+  });
```
NSeydoux, line 256: Instead of updating the existing tests, I think it would be fine to add a separate test just for the clientId, regardless of the `webId`.

packages/core/src/util/token.spec.ts
```diff
@@ -230,18 +274,18 @@ describe("getWebidFromTokenPayload", () => {
         "https://some.issuer",
         "https://some.clientId",
       ),
-    ).resolves.toBe("https://some.webid#me");
+    ).rejects.toThrow();
   });
 
-  it("returns the WebID it the 'sub' claim exists and it is IRI-like", async () => {
+  it("throws if the 'webid' claim exists and but 'azp' does not exist", async () => {
```
NSeydoux, line 280: According to https://openid.net/specs/openid-connect-core-1_0.html#IDToken, the `azp` claim is optional, so we don't want to throw even if it is absent.

packages/core/src/util/token.ts
```diff
@@ -26,20 +26,20 @@ import { jwtVerify, createRemoteJWKSet } from "jose";
...
+ * @returns an object with entries webId and clientId extracted from the ID token.
  * @internal
  */
 export async function getWebidFromTokenPayload(
   idToken: string,
   jwksIri: string,
   issuerIri: string,
   clientId: string,
-): Promise<string> {
+): Promise<{ webId: string; clientId: string }> {
```
NSeydoux, line 42:
```suggestion
): Promise<{ webId: string; clientId?: string }> {
```
As noted above, `azp` is optional.

packages/core/src/util/token.ts
```diff
@@ -26,20 +26,20 @@ import { jwtVerify, createRemoteJWKSet } from "jose";
...
   let payload: JWTPayload;
```
NSeydoux, line 43:
```suggestion
  let payload: JWTPayload;
  let webId: string;
  let clientId: string; 
```

packages/core/src/util/token.ts
```diff
@@ -56,7 +56,12 @@ export async function getWebidFromTokenPayload(
   }
 
   if (typeof payload.webid === "string") {
-    return payload.webid;
+    if (typeof payload.azp === "string") {
+      return { webId: payload.webid, clientId: payload.azp };
+    }
+    throw new Error(
+      `The token ${JSON.stringify(payload)} is invalid: it has no 'azp' claim.`,
+    );
```
NSeydoux, line 64: Let's not nest this conditional here. Instead, let's define a `webid` and `clientId` variables, each initialized in their own blocks. This means `return payload.webid;` just above would become `webId = payload.webid`, and there would be a separate branch as follows: 
```javascript
if (typeof payload.azp === "string") {
  clientId = payload.azp;
}
```

packages/core/src/util/token.ts
```diff
@@ -71,10 +76,16 @@ export async function getWebidFromTokenPayload(
...
       `The token has no 'webid' claim, and its 'sub' claim of [${payload.sub}] is invalid as a URL - error [${e}].`,
     );
   }
+  // The 'sub' claim is a well-formed IRI, hence a valid webId
+  if (typeof payload.azp === "string") {
+    return { webId: payload.sub, clientId: payload.azp };
+  }
+  throw new Error(
+    `The token ${JSON.stringify(payload)} is invalid: it has no 'azp' claim.`,
+  );
```
NSeydoux, line 90: Based on my previous comment, this whole block could be removed and replaced with 
`return { webId, clientId };`. I think type inference should be correct.

NSeydoux (changes requested):

NSeydoux:
And don't forget to add a `CHANGELOG.md` entry describing this fix.

packages/core/src/util/token.ts
garciafdezpatricia: I may need to initialise the `clientId` to any value, like `""`, for it not to complain about returning a variable before it being assigned. Then, return undefined or its value depending if it is `""` or not.
Does it sound reasonable?

NSeydoux: let's do it the other way, and initialize it as undefined, and set it to a `string` if appropriate (adjusting the declaration to `let clientId: string | undefined = undefined`. If I remember correctly, this should be equivalent to the current `let clientId: string;`, because variables are initialized to `undefined` by default, but I'm not 100% sure of that.

packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
```diff
@@ -110,20 +110,20 @@ export default class ClientCredentialsOidcHandler implements IOidcHandler {
...
         tokens.access_token,
         oidcLoginOptions.issuerConfiguration.jwksUri,
         oidcLoginOptions.issuer,
         // When validating the Access Token, the audience should always be 'solid'
         "solid",
-      );
+      ));
     } else {
-      webId = await getWebidFromTokenPayload(
+      ({ webId } = await getWebidFromTokenPayload(
```
garciafdezpatricia, line 121: And retrieve clientId here as well?

packages/core/src/util/token.spec.ts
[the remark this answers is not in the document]
garciafdezpatricia: With the early return approach I believe you're completely right, thank you for checking it out.

Just as context, I had added the test to check if the `clientId` was indeed `undefined` even if the `sub` or `webid` claims were not defined to avoid bugs on the new code (I didn't trust myself completely), more specifically all the conditionals included from  [this line](https://github.com/inrupt/solid-client-authn-js/blob/fix/SDK-3337/parse-clientId/packages/core/src/util/token.ts#L60). 

This test was actually the one that made me introduce `&& typeof payload.webid !== "string"` inside the try clause for the `sub` claim, as, instead of getting an `undefined clientId`, it was returning the error of the catch clause.

NSeydoux (approved):

NSeydoux:
I'll let you write the commit message. The current PR title looks good for a commit title.

PR #3699, opened 2024-09-26
Add rollup to dev dependencies

NSeydoux:
Supersedes https://github.com/inrupt/solid-client-authn-js/pull/3695 to make rollup a direct dev dependency.


edwardsph (approved):
LGTM

e2e/node/server/express.ts
NSeydoux, line 1: Turns out we were not using express quite correctly, and that had never upset TS until now.

PR #3750, opened 2024-10-30
Improve SessionEventListener types

NoelDeMartin:
Whilst debugging some issues, I realized that I was getting an `Error` at runtime, instead of a `string` like Typescript indicates. Digging into the code, I found [a FIXME indicating that this is not the intended behaviour](https://github.com/inrupt/solid-client-authn-js/blob/main/packages/browser/src/ClientAuthentication.ts#L130..L135).

However, since that's been there for 2 years, and I'm not sure if there's any plans to fix it, I think it'd be better that the Typescript annotations actually reflect reality.


NSeydoux:
Superseded by https://github.com/inrupt/solid-client-authn-js/pull/3766

PR #3759, opened 2024-10-31
Expose prompt option

NoelDeMartin:
This PR is mostly a feature request, but I figured I'd open a PR with the changes to propose a solution.

The feature request is basically to expose the `prompt` option in the login. I always thought that working with various Solid PODs was annoying because they asked for confirmation in every page reload, but recently I realized it's actually the expected behaviour because this library sends the "consent" prompt parameter during the [authentication request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). This is done internally, so it would be nice to expose this to clients and let them decide how to behave.

Currently, this can be bypassed using the `prompt` parameter anyways, and telling Typescript to ignore the problem. But there is another issue. Eventually, [the prompt parameter is resolved using a null coalescing operator that defaults to "consent"](https://github.com/inrupt/solid-client-authn-js/blob/main/packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts#L65). This is a problem because in some implementations (for example, [in Laravel](https://laravel.com/docs/11.x/passport#requesting-tokens-redirecting-for-authorization)), there is a distinction between using "none" or omitting the parameter altogether. There should be a way to explicitly avoid sending the parameter, even if the default behaviour is to use "consent".

### DIFF
--- a/packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
+++ b/packages/browser/src/login/oidc/oidcHandlers/AuthorizationCodeWithPkceOidcHandler.ts
@@ -62,7 +62,7 @@ export default class AuthorizationCodeWithPkceOidcHandler
       // profile referenced by the WebId.
       loadUserInfo: false,
       code_verifier: true,
-      prompt: oidcLoginOptions.prompt ?? "consent",
+      prompt: 'prompt' in oidcLoginOptions ? oidcLoginOptions.prompt : "consent",
     };
     /* eslint-enable camelcase */
 
--- a/packages/core/src/ILoginInputOptions.ts
+++ b/packages/core/src/ILoginInputOptions.ts
@@ -55,4 +55,11 @@ export default interface ILoginInputOptions {
    * secret to authenticate.
    */
   refreshToken?: string;
+  /**
+   * Specify whether the Solid Identity Provider may, or may not, interact with the user (for example,
+   * the normal login process **_requires_** human interaction for them to enter their credentials,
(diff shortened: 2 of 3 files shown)


NSeydoux:
Hi @NoelDeMartin, thanks for the suggestion.

The reason for this `prompt` parameter to be hard-coded is because it is mandatory when requesting the `offline_access` scope, which results in a refresh token being issued. Without this refresh token, the session is only good for the duration of the issued Access Token, whish is (purposefully) short.

My concern with the proposed change is that the consequences of opting out of the `prompt` parameter are not quite clear to a developer, and it may result in confusion when the users starts being logged out after 5 minutes (lifespan of the token).

For me to make sure I'm understanding the problem properly, when exactly are you getting prompted for consent? That should happen when logging in the first time, and then again when you do a "full" logout (as in, terminating the session on the OpenID Provider, not only on the client), but in between as long as you're using the same client, if you are "only" logging out of the app you should not get prompted again. I may only be applicable if you're using a Client Identifier though (because the consent is given to one specific client, so when using dynamic client registration, prompt is asked on each login).

NoelDeMartin:
You can reproduce the behaviour I'm talking about in this app: https://ramen.noeldemartin.com.

Every time I reload the app, I am redirected to ESS and I see this:

image

Ideally, once I have already consented, I wouldn't be shown that screen again. The log in is implemented here, in case you want to see how I'm using the library: https://github.com/NoelDeMartin/aerogel/blob/main/packages/plugin-solid/src/auth/authenticators/InruptAuthenticator.ts

PR #3766, opened 2024-11-05
Improve SessionEventListener types

NSeydoux:
Reapplies #3750 for CI

There are some scenarios where an Error is thrown rather than a string, for example within ClientAuthentication.


edwardsph (approved):
LGTM

PR #3823, opened 2025-01-07
Handle dynamic client expiration

NSeydoux:
Dynamically registered clients eventually expire, after a duration provided by the OpenID Provider at registration time. This change keeps track of the expiration date in the browser code, and performs a now dynamic client registration if the client in use expired.

If a legacy client is present in storage, it will be rotated as if it were expired, to avoid the issue where an expired client is used and results in an error on the OP.


packages/oidc-browser/src/dcr/clientRegistrar.ts
NSeydoux, line 1: No functional change, just some cleanup

packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -22,19 +22,43 @@
...
+    typeof OpenIdClient
+  >;
+  const mockedIssuerConfig = mockIssuerMetadata(issuerMetadata);
+  const mockedIssuer: OpenIdClient.Issuer<OpenIdClient.BaseClient> = {
+    metadata: mockedIssuerConfig,
+    Client: {
+      register: (jest.fn() as any).mockResolvedValueOnce({
+        metadata: clientMetadata,
+      }),
+      // The assertions are required because we only mock we is strictly necessary for our tests.
```
edwardsph, line 51:
```suggestion
      // The assertions are required because we only mock what is strictly necessary for our tests.
```
Is this what you mean?

packages/node/src/login/oidc/ClientRegistrar.spec.ts
```diff
@@ -50,15 +74,12 @@ describe("ClientRegistrar", () => {
 
   describe("getClient", () => {
     it("fails if there is not registration endpoint", async () => {
```
edwardsph, line 76:
```suggestion
    it("fails if there is no registration endpoint", async () => {
```

edwardsph (approved):
LGTM - 2 trivial comments

PR #3842, opened 2025-01-23
Fix browser demo error management

### LINKED ISSUE #3841: Type error: Object literal may only specify known properties, and 'onError' does not exist in type 'IHandleIncomingRedirectOptions'.
I'm getting this error when trying to compile https://github.com/inrupt/solid-client-authn-js/blob/main/packages/browser/examples/single/bundle/src/App.js#L51 as TypeScript.
Is it missing from 'IHandleIncomingRedirectOptions'?

NSeydoux:
Fixes #3841.

The browser demo no longer passes an incorrect argument (from a TS perspective) to the `handleIncomingRedirect` function. This also fixes how the React app root is created with respect to the latest upgrade.

### DIFF
--- a/packages/browser/examples/single/bundle/src/App.js
+++ b/packages/browser/examples/single/bundle/src/App.js
@@ -90,36 +90,34 @@ export default function App() {
   };
 
   return (
-    <div>
-      <main>
-        <h1>Sandbox app</h1>
-        <p>{webId ? `Logged in as ${webId}` : "Not logged in yet"}</p>
-        <div>
-          <form>
-            <input
-              type="text"
-              value={issuer}
-              onChange={(e) => {
-                setIssuer(e.target.value);
-              }}
-            />
-            <button onClick={(e) => handleLogin(e)}>Log In</button>
-            <button onClick={(e) => handleLogout(e)}>Log Out</button>
-          </form>
-        </div>
-        <hr />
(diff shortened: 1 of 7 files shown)


chelseapinka (approved):

packages/browser/examples/single/bundle/src/App.js
```diff
@@ -59,10 +60,9 @@ export default function App() {
     console.log(`${error} has occured: `, errorDescription);
   };
 
-  const handleLogin = (e) => {
-    // The default behaviour of the button is to resubmit.
-    // This prevents the page from reloading.
-    e.preventDefault();
+  events().on(EVENTS.ERROR, errorHandle);
```
NSeydoux, line 63: This is how error events should be handled on the session.

packages/browser/examples/single/bundle/src/index.js
```diff
@@ -20,7 +20,7 @@
 //
 
 import React from "react";
-import ReactDOM from "react-dom";
+import { createRoot } from 'react-dom/client';
 import App from "./App";
 
-ReactDOM.render(<App />, document.getElementById("container"));
```
NSeydoux, line 26: This approach was deprecated from React in a previous version.

chelseapinka (approved):

PR #3874, opened 2025-03-06
deno.js support

unit9a:
# New feature description
Support for Deno.js with a simple mod.ts file linked to a esm formatted rollup config via the browser package.

no new functions or logic added, just change to browser rollup config. 

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "rollup": "^4.22.4",
     "ts-jest": "^29.3.0",
     "ts-node": "^10.9.2",
-    "typedoc": "^0.28.0",
+    "typedoc": "^0.28.1",
     "typedoc-plugin-markdown": "^3.17.1",
     "typescript": "^5.8.2"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,7 +44,7 @@
         "rollup": "^4.22.4",
         "ts-jest": "^29.3.0",
         "ts-node": "^10.9.2",
-        "typedoc": "^0.28.0",
+        "typedoc": "^0.28.1",
         "typedoc-plugin-markdown": "^3.17.1",
         "typescript": "^5.8.2"
       },
@@ -22301,9 +22301,9 @@
       "license": "MIT"
     },
     "node_modules/typedoc": {
-      "version": "0.28.0",
-      "resolved": "https://registry.npmjs.org/typedoc/-/typedoc-0.28.0.tgz",
-      "integrity": "sha512-UU+xxZXrpnUhEulBYRwY2afoYFC24J2fTFovOs3llj2foGShCoKVQL6cQCfQ+sBAOdiFn2dETpZ9xhah+CL3RQ==",
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/typedoc/-/typedoc-0.28.1.tgz",
+      "integrity": "sha512-Mn2VPNMaxoe/hlBiLriG4U55oyAa3Xo+8HbtEwV7F5WEOPXqtxzGuMZhJYHaqFJpajeQ6ZDUC2c990NAtTbdgw==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {


package.json
```diff
@@ -40,8 +40,10 @@
     "build-docs-preview-site": "lerna run build-docs-preview-site; rm -r dist/website || true; lerna exec -- mkdir -p ../../dist/website/\\${PWD##*/}; lerna exec -- \"cp -r docs/api/build/html/. ../../dist/website/\\${PWD##*/}/ || true\"; echo 'API documentation: <a href=\"./browser/\">solid-client-authn-browser</a>, <a href=\"./node/\">solid-client-authn-node</a>, <a href=\"./core/\">solid-client-authn-core</a>, <a href=\"./demo/\">demo app</a>.' >> dist/website/index.html",
     "build-demo-app": "npm run build; cd packages/browser/examples/demoClientApp/; npm ci; npm run build; mkdir -p ../../../../dist/website/demo/; cp -r dist/. ../../../../dist/website/demo/"
   },
+  "dependencies": {
+    "@inrupt/base-rollup-config": "^3.2.4"
+  },
   "devDependencies": {
-    "@inrupt/base-rollup-config": "^3.2.4",
```
NSeydoux, line 44: Why is this change necessary? The rollup configuration should not be in the runtime dependencies.

packages/browser/package.json
```diff
@@ -30,8 +31,12 @@
     "@types/uuid": "^10.0.0"
   },
   "dependencies": {
+    "@inrupt/base-rollup-config": "^3.2.4",
     "@inrupt/oidc-client-ext": "^2.3.0",
     "@inrupt/solid-client-authn-core": "^2.3.0",
+    "@rollup/plugin-commonjs": "^28.0.2",
+    "@rollup/plugin-node-resolve": "^16.0.0",
+    "@rollup/plugin-terser": "^0.4.4",
```
NSeydoux, line 39: Ditto on the rollup dependency being a runtime dependency.

NSeydoux (commented):
I have an overall question about this PR, which comes from my complete lack of familiarity with Deno. I would have expected the Deno bundle to come from the NodeJS library, rather than the browser one. Does the Deno runtime provide all the browser APIs, including things like local storage? My understanding is that Deno is a server-side runtime (as opposed to a browser-side one), which would not be compatible with assumptions made in the browser library.

Can you tell me more about what deno support entails, and how this library would be used in a Deno environment, so that I can have a better understanding of the changes you're proposing?

unit9a:
> I have an overall question about this PR, which comes from my complete lack of familiarity with Deno. I would have expected the Deno bundle to come from the NodeJS library, rather than the browser one. Does the Deno runtime provide all the browser APIs, including things like local storage? My understanding is that Deno is a server-side runtime (as opposed to a browser-side one), which would not be compatible with assumptions made in the browser library.
> 
@NSeydoux  
You are correct about Deno.js being server side like Node.js. However it complies with browser web standards to much higher degree.  I am not sure about the idea of using localstorage server side.... I will ponder this.

If this was a ment be a true Deno port, the "events": "^3.3.0" depancy would not be needed.... i think.   

https://docs.deno.com/runtime/reference/web_platform_apis/

> Can you tell me more about what deno support entails, 

I should have been more specific., i ams using https://fresh.deno.dev/ .  it, inturn, uses Preact.  so all that was need was a browser/Preact compatible esm version the current iife distribution.  That I did via the rollup config. 

> and how this library would be used in a Deno environment, so that I can have a better understanding of the changes you're proposing?

this is how i import it from my local fork

`import {  login, getDefaultSession } from '../../../../../solid-client-authn-js/mod.ts';
`

I am currently experimenting  with this import method in my projects deno.json config file
`    "@inrupt/solid-client-authn-browser": "https://raw.githubusercontent.com/unit9a/solid-client-authn-js/refs/heads/main/mod.ts",
`

a "mod.ts" file is the entry point in deno convention for creating libraries.

packages/browser/package.json
unit9a: @NSeydoux 
 what is odd to me when I run `npm ci` is that when I run `npm ci`,  two builds are failing for me.
 
``` 
      npm error command failed
      npm error command sh -c webpack --mode production
```
``` 
      npm error command failed
      npm error command sh -c next build
```


```
 ✖  2/6 targets failed, including the following:

      - browser-in-bundle:build
      - test-app:build
```      

``` 
      npm error command failed
      npm error command sh -c npm run build
```

yet the rollup output seem to work fine.

unit9a:
@NSeydoux 
 what is odd to me when I run `npm ci` is that when I run `npm ci`,  two builds are failing for me.
 
``` 
      npm error command failed
      npm error command sh -c webpack --mode production
```
``` 
      npm error command failed
      npm error command sh -c next build
```


```
 ✖  2/6 targets failed, including the following:

      - browser-in-bundle:build
      - test-app:build
```      

``` 
      npm error command failed
      npm error command sh -c npm run build
```

yet the rollup output seem to work fine.

unit9a:
@NSeydoux any thoughts on the build failures?

unit9a:
@NSeydoux  is this a bug? 
i am getting this `400 (Bad Request)`  error 
```
... [IdentityProviderFactory] {Primary} warn: OIDC request failed: invalid_redirect_uri - redirect_uris must only contain strings
```
throw by: 

https://github.com/inrupt/solid-client-authn-js/blob/74cbaeaa37977b130e6980ab64f9747e9f2fc568/packages/browser/src/login/oidc/ClientRegistrar.ts#L102

when i call
```
 await handleIncomingRedirect({
    restorePreviousSession: true,
    url: location.href
  })
``` 
with location.href  being this test string:  
```
http://localhost:8000/pwa/settings/solidAuth?code=-mqaQQqEBuO7LJv9BWL7o5vrk6FINyQfkeStBZZo7W7&state=00bcf50e4b8c4f1fb52fabe1cf79ef32&iss=http%3A%2F%2Flocalhost%3A3000%2F
 ```

NSeydoux:
I'll look into the build failures.

Regarding the `400 Bad Request` response, let me make sure I get the context right: you are going through the authorization code flow (browser-based redirections), and getting the error response when after logging in the OpenID provider, and going back to your app.

If that is the case, `registerClient` should not be called, as the client is registered earlier in the interaction, before redirecting the user to the OpenID Provider authorization endpoint. The client being registered where you are seeing it means the app cannot find the previously registered client, which should have been stored in `localStorage`.

I suspect this is a manifestation of the discrepancy between the Deno APIs and the browser's.

unit9a:
@NSeydoux 
did not mean to close this pr. still debugging.

> I suspect this is a manifestation of the discrepancy between the Deno APIs and the browser's.

i am getting this error in chrome, NOT  deno.

PR #3904, opened 2025-04-04
SDK-3405: Emit event when new tokens are received

edwardsph:
# Emit event when new tokens are received

# Checklist

- [ ] Relevant documentation, if any, has been written/updated. - Separate ticket

### DIFF
--- a/e2e/browser/solid-client-authn-browser/test/fixtures.ts
+++ b/e2e/browser/solid-client-authn-browser/test/fixtures.ts
@@ -130,6 +130,7 @@ const createClientIdDoc = async (
     clientName: string;
     redirectUrl: string;
     scope?: string;
+    grantTypes?: string[];
   },
   container: string,
   session: Session,
@@ -153,7 +154,7 @@ const createClientIdDoc = async (
     // should be updated so that it has the offline_access scope and supports the
     // refresh_token grant type.
     scope: clientInfo.scope ?? "openid webid",
-    grant_types: ["authorization_code"],
+    grant_types: clientInfo.grantTypes ?? ["authorization_code"],
     response_types: ["code"],
     post_logout_redirect_uris: [
       `http://localhost:${CONSTANTS.CLIENT_AUTHN_TEST_PORT}/postLogoutUrl`,
@@ -275,6 +276,7 @@ export interface ISeedPodResponse {
 
 export async function seedPod(
   setupEnvironment: TestingEnvironmentNode,
+  enableRefresh?: boolean,
 ): Promise<ISeedPodResponse> {
   if (
     setupEnvironment.clientCredentials.owner.type !== "ESS Client Credentials"
@@ -315,6 +317,10 @@ export async function seedPod(
       redirectUrl: new URL(
         `http://localhost:${CONSTANTS.CLIENT_AUTHN_TEST_PORT}`,
       ).href,
+      scope: enableRefresh ? "openid webid offline_access" : undefined,
+      grantTypes: enableRefresh
+        ? ["authorization_code", "refresh_token"]
+        : undefined,
     },
     podRoot,
     session,
--- a/e2e/node/script/e2e-test.spec.ts
+++ b/e2e/node/script/e2e-test.spec.ts
@@ -39,6 +39,8 @@ import { getSolidDataset } from "@inrupt/solid-client";
 // The relative import is used to run on latest library, rather than latest published.
 // eslint-disable-next-line import/no-relative-packages
 import { EVENTS, Session } from "../../../packages/node/src/index";
+// eslint-disable-next-line import/no-relative-packages
+import type { SessionTokenSet } from "../../../packages/core/src/index";
 
 custom.setHttpOptionsDefaults({
   timeout: 15000,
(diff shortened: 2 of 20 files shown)


packages/core/src/SessionEventListener.ts
```diff
@@ -20,11 +20,42 @@
...
 }
 
+/**
+ * A set of tokens to be passed to the client
+ */
+export type SessionTokenSet = {
+  /**
+   * JWT-serialized access token
+   */
+  access_token: string;
```
NSeydoux, line 36:
```suggestion
  accessToken: string;
```
The snake casing is aligned with the specification, but we'll want to use camel case for our type. I'm surprised the linter didn't complain, we have to add linting exceptions for such names in other places. Can you check the linter runs appropriately on the core package?

packages/core/src/SessionEventListener.ts
```diff
@@ -51,10 +82,17 @@ type TIMEOUT_SET_ARGS = {
   eventName: typeof EVENTS.TIMEOUT_SET;
   listener: (timeoutId: number) => void;
 };
+/**
+ * @deprecated
```
NSeydoux, line 86:
```suggestion
 * @deprecated Use NEW_TOKENS_ARGS instead
```

e2e/node/server/express.ts
NSeydoux, line 1: This will be more relevant for the next PR, but we'll want to refactor this server to retrieve the sessions from storage.

e2e/browser/solid-client-authn-browser/test/fixtures.ts
```diff
@@ -275,6 +276,7 @@ export interface ISeedPodResponse {
 
 export async function seedPod(
   setupEnvironment: TestingEnvironmentNode,
+  enableRefresh?: boolean,
```
edwardsph, line 279: The changes here are all about allowing a test to use the refresh flow but by default it is still not enabled

e2e/node/script/e2e-test.spec.ts
```diff
@@ -33,12 +33,9 @@ import {
...
 } from "@jest/globals";
 import { custom } from "openid-client";
-// Here we want to test how the local code behaves, not the already published one.
-// eslint-disable-next-line import/no-relative-packages
 import { getSolidDataset } from "@inrupt/solid-client";
-// The relative import is used to run on latest library, rather than latest published.
-// eslint-disable-next-line import/no-relative-packages
-import { EVENTS, Session } from "../../../packages/node/src/index";
+import { EVENTS, Session } from "@inrupt/solid-client-authn-node";
+import type { SessionTokenSet } from "@inrupt/solid-client-authn-node";
```
edwardsph, line 38: Lerna uses symlinks so the imports can be simplified and still use local versions of the library

e2e/node/server/express.ts
```diff
@@ -37,8 +42,19 @@ export function createApp(
 ) {
   const app = express();
 
-  // Initialised when the server comes up and is running...
-  let session: Session;
+  const sessionTokenSets = new Map<string, SessionTokenSet>();
+
+  app.use(
```
edwardsph, line 47: The express app now gets the session from storage which will support additional tests when the API changes are made.

packages/core/src/SessionEventListener.ts
```diff
@@ -20,11 +20,42 @@
 //
 import type { EventEmitter } from "events";
 import type { EVENTS } from "./constant";
+import type { KeyPair } from "./authenticatedFetch/dpopUtils";
 
 export interface IHasSessionEventListener {
   events: ISessionEventListener;
 }
 
+/**
```
edwardsph, line 29: This is the object pass in the NEW_TOKENS event

packages/core/src/SessionEventListener.ts
```diff
@@ -51,10 +82,17 @@ type TIMEOUT_SET_ARGS = {
   eventName: typeof EVENTS.TIMEOUT_SET;
   listener: (timeoutId: number) => void;
 };
+/**
+ * @deprecated Use NEW_TOKENS_ARGS instead
+ */
```
NSeydoux, line 87: Can you add the deprecation warning on the signatures using this event name? E.g. `events.on(eventName: NEW_REFRESH_TOKEN_ARGS.eventName, listener: NEW_REFRESH_TOKEN_ARGS.listener`. Some people will not be using the types directly, and get the typing from IDE hints directly, so the deprecation should apply even if the literals are used and not the exported type.

packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
```diff
@@ -59,7 +60,6 @@ export default class ClientCredentialsOidcHandler implements IOidcHandler {
       // If a redirect URL is present, the static client should use the
       // authorization code flow.
       typeof oidcLoginOptions.redirectUrl !== "string" &&
-      typeof oidcLoginOptions.client.clientId === "string" &&
```
NSeydoux, line 62: Why has this line been removed?

NSeydoux (approved):

packages/node/src/login/oidc/oidcHandlers/ClientCredentialsOidcHandler.ts
edwardsph: I should have added a comment in GitHub. It is redundant as all variants of IClient have clientId as a mandatory string. My IDE pointed it out but I checked myself by looking at the types and agreed.

PR #3924, opened 2025-04-17
Remove debug code from server example

NSeydoux:


davidboweninrupt (approved):
Not really sure why `time` was removed from the URL

NSeydoux:
It had been added just for the purpose of the sprint review demo, and shouldn't have been committed in the first place.

PR #3935, opened 2025-04-28
Internal refactoring of multisession functions

NSeydoux:
Split the Session manipulation functions between the legacy ones using the embedded storage and the new ones integrated with external storage.

This is just some code moving around, it has no effect on the public API.


edwardsph (approved):
LGTM

PR #3938, opened 2025-04-30
Fix race condition in browser-based test

NSeydoux:
In some cases, a stale value was read from the application UI wile it was being re-rendered. This change makes it so that the test-id being rendered is removed from the DOM while the data is being fetched so that the test script waits for the latest value to be displayed.


NSeydoux:
Due to an auto-merge mistake, this includes the changes from https://github.com/inrupt/solid-client-authn-js/pull/3935. The relevant changes to review here are located in `e2e/browser/solid-client-authn-browser/test-app/components/authenticatedFetch/index.tsx`

e2e/browser/solid-client-authn-browser/test-app/components/authenticatedFetch/index.tsx
NSeydoux, line 1: This is the relevant changes for this PR, other changes have been reviewed in a different PR that got merged into this one.

jholleran (approved):

PR #3976, opened 2025-06-05
Reintroduce custom fetch to buildAuthenticatedFetch as an option

### LINKED ISSUE #3810: buildAuthenticatedFetch no longer supports custom fetch as a first argument
This is either bug or (regressed) feature request.

### Search terms you've used

`buildAuthenticatedFetch`

### Impacted package

- [x] solid-client-authn-core

### Description

It looks like since v2, the library no longer supports passing custom fetch to `buildAuthenticatedFetch` (used to be first argument).  That can be considered a breaking change and feature regression. Please note that others depend directly on `buildAuthenticatedFetch`, and passing custom fetch can be very useful. In my case building authenticated request in Cypress.

### Expected result

It would be awesome if we can continue passing custom fetch to buildAuthenticatedFetch (e.g. as an option). I don't mind if API changes (e.g. passing fetch as optional property in second parameter).

NSeydoux:
Reapplies https://github.com/inrupt/solid-client-authn-js/pull/3952
Fixes #3810

### DIFF
--- a/packages/core/src/authenticatedFetch/fetchFactory.ts
+++ b/packages/core/src/authenticatedFetch/fetchFactory.ts
@@ -95,8 +95,9 @@ async function makeAuthenticatedRequest(
   url: RequestInfo | URL,
   defaultRequestInit?: RequestInit,
   dpopKey?: KeyPair,
+  unauthFetch = fetch,
 ) {
-  return fetch(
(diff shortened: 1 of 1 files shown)


edwardsph (approved):
LGTM

PR #4267, opened 2026-05-06
Upgrade jose to v6

NSeydoux:


packages/node/src/util/dpopInput.ts
NSeydoux, line 1: Note for reviewers: I'm not thrilled about this workaround but hopefully we get the openid-client upgrade in shortly after, and this can be removed. That upgrade is a pretty significant change of the openid-client API, so it is too large to be done in the same PR as the jose upgrade and keep things reviewable.

NSeydoux:
The sonar issues are false positives or will be addressed in the follow-up openid-client upgrade.

emmettownsend (approved):
We need to make it clear in our documentation that this is a breaking change, or developers using it may end up with compile errors.

PR #4356, opened 2026-08-06
fix(browser): re-register when the OIDC provider no longer knows the stored client

### LINKED ISSUE #4355: Silent authentication loops forever when the OIDC provider no longer knows the stored dynamic client
When a Solid OIDC provider drops a browser's dynamically-registered client (for example, the provider kept client registrations in memory and restarted), the browser's stored `clientId` is no longer known to the provider. On session restore, `silentlyAuthenticate` redirects to the authorization endpoint with the stale `clientId` and `prompt=none`.

The provider cannot redirect the error back to the app (it cannot validate the `redirect_uri` of an unknown client), so it returns a non-redirectable "unknown client" error page (e.g. Community Solid Server returns HTTP 400 `E0003 Unknown client`). Because `KEY_CURRENT_URL` is only cleared after a *successful* silent auth, the next page load re-attempts silent auth with the same rejected `clientId` — looping indefinitely.

### Impact

- Affected users see permanently broken sessions/pods. Clearing localStorage on the app origin does not help, because the stale registration lives on the pod subdomain origin.
- The retry storms add load to the provider. Observed in production against Community Solid Server (solidcommunity.net), where it surfaced as rate-limiting (429s).

### Environment

- `@inrupt/solid-client-authn-browser` v5.0.0
- Any Solid-OIDC provider using dynamic client registration that can lose registrations (e.g. Community Solid Server with in-memory OIDC adapter storage).

### Steps to reproduce

1. Log in to an app that uses dynamic client registration against a CSS instance; the session is restored across reloads via silent auth.
2. Cause the provider to forget the dynamic client (e.g. restart a CSS instance whose OIDC adapter storage is in-memory).
3. Reload the app: silent auth redirects to the provider, gets `400 Unknown client`, and on returning to the app re-attempts with the same `clientId`. The session never recovers.

### Expected

On an unknown/invalid client during silent authentication, the library should discard the stored client registration and re-register instead of looping.

### Proposed fix

Detect an incomplete previous silent-auth attempt (a `KEY_CURRENT_URL` still set on re-entry) and discard the stored client registration so the next login re-registers. PR to follow.

jeswr:
Fixes #4355.

## Description

When a Solid OIDC provider drops a browser's dynamically-registered client (e.g. it kept client registrations in memory and restarted), the browser's stored `clientId` is no longer known to the provider. On session restore, `silentlyAuthenticate` redirects to the authorization endpoint with the stale `clientId` and `prompt=none`. The provider cannot redirect the error back (it can't validate the `redirect_uri` of an unknown client), so it returns a non-redirectable "unknown client" error (e.g. Community Solid Server returns `400 E0003 Unknown client`). Because `KEY_CURRENT_URL` is only cleared after a *successful* silent auth, the next load re-attempts with the same rejected `clientId` — looping indefinitely. The user's session appears permanently broken, and the retry storms add load to the provider. Observed in production against Community Solid Server (solidcommunity.net), where it surfaced as rate-limiting (429s).

## Change

`silentlyAuthenticate` now detects an incomplete previous attempt (a `KEY_CURRENT_URL` still set on re-entry) and, instead of retrying, discards the stored dynamic client registration and stops. The next login registers a fresh client and recovers automatically.

- `SessionInfoManager.clearClientRegistrationInfo(sessionId)` — targeted removal of just the dynamic client registration keys (`clientId`, `clientSecret`, `clientType`, `clientName`, `expiresAt`, `idTokenSignedResponseAlg`); the rest of the session is left intact.
- `ClientAuthentication.clearClientRegistrationInfo(sessionId)` — public passthrough used by `silentlyAuthenticate`.
- Regression test in `Session.spec.ts`; CHANGELOG "Unreleased → Bugfixes" entry.

## Notes / trade-offs

- `KEY_CURRENT_URL` is shared across tabs, so a concurrent silent auth in another tab could be interpreted as an incomplete previous attempt and trigger a re-registration. This is recoverable (the next login re-registers) and strictly better than the current infinite loop; happy to refine (e.g. a dedicated per-attempt marker) if you'd prefer.
- The change is contained to the browser package with a typed cast to the concrete `SessionInfoManager`. I'm happy to promote `clearClientRegistrationInfo` to `ISessionInfoManager` (and add a node implementation) if you'd rather have it on the interface.

## Checklist

- [x] Unit test for the regression (`packages/browser/src/Session.spec.ts`)
- [x] CHANGELOG updated (Unreleased → Bugfixes)
- [x] All affected browser specs pass; lint clean


packages/browser/src/ClientAuthentication.ts
```diff
@@ -109,6 +110,16 @@ export default class ClientAuthentication extends ClientAuthenticationBase {
     return sessionInfo;
   };
 
+  // Discards the stored dynamic client registration for a session so that the next login
+  // registers a fresh client. Used to recover when the OIDC provider no longer recognises the
+  // stored client (e.g. it dropped the registration on a restart), which would otherwise cause
+  // silent authentication to loop with a client ID the provider rejects.
+  clearClientRegistrationInfo = async (sessionId: string): Promise<void> => {
+    await (
+      this.sessionInfoManager as SessionInfoManager
```
NSeydoux, line 119: Could the type assertion here be replaced with a proper type guard?

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 
+### Bugfixes
+
+- Browser: recover from a stored dynamic client registration that the OIDC provider no longer
+  recognises. Previously, if the provider had dropped a client's registration (for example on a
+  restart), silent authentication would redirect to the provider, get a non-redirectable
+  "unknown client" error, and — because the `KEY_CURRENT_URL` marker was never cleared — retry
+  indefinitely with the same rejected client ID, leaving the user's session permanently broken.
+  On detecting an incomplete previous silent-authentication attempt, the stored client
+  registration is now discarded so the next login re-registers a fresh client.
```
NSeydoux:
```suggestion
#### browser

- Recover from a stored dynamic client registration that the OIDC provider no longer
  recognises. Previously, if the provider had dropped a client's registration (for example on a
  restart), silent authentication would redirect to the provider, get a non-redirectable
  "unknown client" error, and — because the `KEY_CURRENT_URL` marker was never cleared — retry
  indefinitely with the same rejected client ID, leaving the user's session permanently broken.
  On detecting an incomplete previous silent-authentication attempt, the stored client
  registration is now discarded so the next login re-registers a fresh client.
```
Just for consistency with other entries.

NSeydoux (approved):

NSeydoux:
This looks good, once the only feedback has been addressed (the type assertion) I'll open a new PR reapplying your commits so that CI runs, and then I'll merge and release it.

PR #4388, opened 2026-09-09
 Support registered clients in static Session ctor 

NSeydoux:
`Session.fromAuthorizationRequestState` can now be used with a statically registered client. A NodeJS client that has been registered can authenticate to the OpenID Provider when going through the Authorization Code flow. Previously, only Solid-OIDC URL-based client identifiers were supported.
## Checklist

- [ ] Relevant documentation, if any, has been written/updated. N/A
- [ ] New functions/types have been exported in `index.ts`, if applicable. N/A


tsconfig.build.json
```diff
@@ -10,6 +10,7 @@
     "module": "preserve",
     "strict": true,
     "declaration": true,
+    "declarationMap": true,
```
NSeydoux, line 13: This is not directly related to the changes in this PR, but it makes navigating the code easier (it was long overdue).

packages/core/src/sessionInfo/ISessionInfoManager.ts
```diff
@@ -34,6 +38,18 @@ export interface ISessionInfoManagerOptions {
   webId?: string;
 }
 
+export type SessionManagerAuthorizationState = AuthorizationRequestState & {
```
NSeydoux, line 41: Enforce the supported states in the type to prevent invalid state being representable.

packages/node/examples/server/.env.example
```diff
@@ -11,3 +11,5 @@ POST_LOGOUT_REDIRECT_URL="http://localhost:3001"
 
 # Solid-OIDC Client identifier for the application.
 CLIENT_ID="https://storage.inrupt.com/d70fd154-de71-4627-89f3-9f9515950f7a/client_ids/ca999ecd-7951-4982-b7eb-d4376985f978"
+# For a statically registered client, set the CLIENT_ID and CLIENT_SECRET to the credentials issued by the Identity Provider
```
jholleran, line 14: Should we have to set a CLIENT_TYPE? It would be a bit more explicit if we had to.

packages/node/src/Session.ts
```diff
@@ -177,21 +179,40 @@ export class Session implements IHasSessionEventListener {
...
+        keepAlive: false,
+        clientType: "solid-oidc",
+      };
+    } else if (hasSecret) {
+      state = {
+        ...authorizationRequestState,
+        keepAlive: false,
+        clientType: "static",
+        clientSecret,
+      };
```
jholleran, line 212: A follow on from: https://github.com/inrupt/solid-client-authn-js/pull/4388/changes#r3970848217 - it might be more extensible if we avoid this logic and have a proper `type` setting.

NSeydoux: That's a good point, we track this internally anyway. However, I'd prefer to do this change separately, so that it is applied consistently across all the relevant APIs (which goes beyond the scope of this PR): the client type should be set before the authorization request (so in the `login` function), and the same change should be applied to the browser package. I'll create the follow-up ticket.

jholleran (approved):

